PR #2, opened 2018-11-07
Allow openAPI 3 to have correct requestBody schema

danieltheodosius:
Changes:
- add `HttpContentType` that currently only supports `application/json` and `text/html`
- add optional `requestContentType` into endpoint that will get defaulted to `application/json`
- openAPI 3 now have generated `content` for the `requestBody`

Before the change:
```yaml
      requestBody:
        $ref: '#/components/schemas/CreateUserRequest'
```

After the change:
```yaml
      requestBody:
        content:
          application/json: 
            schema:
              $ref: '#/components/schemas/CreateUserRequest'
```

Thing to discuss:
- is it a good idea to default the content type as `application/json`?

### DIFF
--- a/lib/src/generators/contract/openapi3-schema.ts
+++ b/lib/src/generators/contract/openapi3-schema.ts
@@ -28,7 +28,9 @@ export function openApiV3ContentTypeSchema(contentType: HttpContentType, type: T
       return {
         content: {
           "text/html": {
-            schema: openApi3TypeSchema(type)
+            schema: {
+              type: "string"
+            }
           }
         }
       };
@@ -157,7 +159,7 @@ export interface OpenAPI3SchemaApplicationJsonContentType extends OpenAPI3BaseSc
 export interface OpenAPI3SchemaTextHtmlContentType extends OpenAPI3BaseSchemaType {
   content: {
     'text/html': {
-      schema: OpenAPI3SchemaType | null
+      schema: OpenAPI3SchemaTypeString
     }
   }
 }
--- a/lib/src/generators/contract/openapi3-schema.ts
+++ b/lib/src/generators/contract/openapi3-schema.ts
@@ -1,6 +1,7 @@
 import assertNever from "../../assert-never";
 import { Type } from "../../models";
 import compact = require("lodash/compact");
+import {HttpContentType} from "@zenclabs/api";
 
 export function rejectVoidOpenApi3SchemaType(
   type: Type,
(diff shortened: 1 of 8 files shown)


lib/src/generators/contract/openapi3-schema.ts
```diff
@@ -118,6 +142,26 @@ export interface OpenAPI3BaseSchemaType {
...
+    'application/json': {
+      schema: OpenAPI3SchemaType | null
+    }
+  }
+}
+
+export interface OpenAPI3SchemaTextHtmlContentType extends OpenAPI3BaseSchemaType {
+  content: {
+    'text/html': {
+      schema: OpenAPI3SchemaType | null
```
fwouts: When the content is `text/html`, isn't the schema always `string`?

lib/src/lib.ts
```diff
@@ -56,6 +57,20 @@ export function isHttpMethod(method: string): method is HttpMethod {
   }
 }
 
+export function isHttpContentType(contentType: String): contentType is HttpContentType {
```
fwouts: Use `string`, not `String`. See https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

lib/src/parser.ts
```diff
@@ -188,6 +190,23 @@ function extractEndpoint(
       )}`
     );
   }
+
+  let requestContentType = "application/json";
+  if (requestContentTypeLiteral) {
+    if (!isStringLiteral(requestContentTypeLiteral)) {
+      throw panic(
+        `Invalid method in endpoint description: ${endpointDescriptionExpression.getText(
```
fwouts: `Invalid request content type`?

lib/src/parser.ts
```diff
@@ -188,6 +190,23 @@ function extractEndpoint(
...
+          sourceFile
+        )}`
+      );
+    }
+    requestContentType = requestContentTypeLiteral.text;
+
```
fwouts: nit: remove line break here.

fwouts (approved):
Thanks, changes look good. Just a few minor comments.

lib/src/generators/contract/openapi3-schema.ts
danieltheodosius: yes, that is correct. changing it

lib/src/lib.ts
danieltheodosius: sorry it is a typo

lib/src/parser.ts
danieltheodosius: changing it

lib/src/parser.ts
danieltheodosius: changing it

fwouts (approved):

PR #3, opened 2018-11-08
Fix OpenAPI 2 generator to produces vallid OpenAPI 2

danieltheodosius:
Changes:
- rename `#/components/schemas/...` into `#/definitions/...`
- remove `requestBody` and replace it with body parameters
- append `requestContentType` as `consumes` attribute

Before change:
```yaml
      parameters: []
      requestBody:
        $ref: '#/components/schemas/CreateUserRequest'
      responses:
        '200':
          schema:
            $ref: '#/components/schemas/CreateUserResponse'
          description: ''
        default:
          description: ''
```

After the change:
```yaml
      consumes:
        - application/json
      parameters:
        - in: body
          name: body
          description: TODO
          required: true
          schema:
            $ref: '#/definitions/CreateUserRequest'
      responses:
        '200':
          schema:
            $ref: '#/definitions/CreateUserResponse'
          description: ''
        default:
          description: ''
```

### DIFF
--- a/lib/src/generators/contract/openapi2-schema.ts
+++ b/lib/src/generators/contract/openapi2-schema.ts
@@ -78,14 +78,8 @@ export function openApi2TypeSchema(type: Type): OpenAPI2SchemaType | null {
     case "optional":
       throw new Error(`Unsupported top-level optional type`);
     case "union":
-      const types = type.types.map(t => openApi2TypeSchema(t));
-      const withoutNullTypes = compact(types);
-      if (withoutNullTypes.length !== types.length) {
-        throw new Error(`Unsupported void type in union`);
-      }
-      return {
-        allOf: withoutNullTypes
-      };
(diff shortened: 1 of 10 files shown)


lib/src/generators/contract/openapi2-schema.ts
```diff
@@ -86,11 +84,11 @@ export function openApi2TypeSchema(type: Type): OpenAPI2SchemaType | null {
         throw new Error(`Unsupported void type in union`);
       }
       return {
-        oneOf: withoutNullTypes
+        allOf: withoutNullTypes
```
fwouts: Please just throw an error since Swagger doesn't support unions: `throw new Error('Unions are not supported in OpenAPI 2')`. Add a comment linking to https://github.com/OAI/OpenAPI-Specification/issues/333.

fwouts (changes requested):

fwouts (approved):

PR #4, opened 2018-11-19
Add support for query parameters to describe API

danieltheodosius:


fwouts:
This will fix https://github.com/zenclabs/spot/issues/14 :)

examples/src/multi-file/api.ts
```diff
@@ -1,3 +1,4 @@
 import "./endpoints/create-user";
 import "./endpoints/delete-user";
+import "./endpoints/get-users";
```
fwouts: How about `find-users` to avoid confusion with `get-user`?

examples/src/multi-file/endpoints/get-users.ts
```diff
@@ -0,0 +1,17 @@
+import {api, endpoint, Optional, queryParam} from "@zenclabs/spot";
+
+@api()
+export class Api {
+  @endpoint({
+    method: "GET",
+    path: "/users"
+  })
+  getUsers(
+    @queryParam({required: true}) limit: number
```
fwouts: Instead of having a `required` boolean, why not rely on TypeScript optional semantics?

In this case, you would use:
```@queryParam() limit: number```

If you wanted it to be optional, you could use either:
```
@queryParam() limit?: number;
@queryParam() limit: Optional<number>;
```

**Review by fwouts — CHANGES_REQUESTED** [review]
Nice work. Please run Prettier—I'll add a linting rule to CircleCI to enforce this soon.

File: examples/src/multi-file/api.ts (hunk 1 above)
Comment (danieltheodosius) [reply examples/src/multi-file/api.ts]: changed

File: examples/src/multi-file/endpoints/get-users.ts (hunk 2 above)
Comment (danieltheodosius) [reply examples/src/multi-file/endpoints/get-users.ts]: changed to `@queryParam limit:Optional<number>`

File: examples/src/multi-file/endpoints/find-users.ts [hunk 3]
```diff
@@ -0,0 +1,19 @@
+import { api, endpoint, Optional, queryParam } from "@zenclabs/spot";
+
+@api()
+export class Api {
+  @endpoint({
+    method: "GET",
+    path: "/users"
+  })
+  getUsers(
+    @queryParam limit: number
```
fwouts: Is there a specific reason for making this different from `single-file`? (it's not optional here)

To clarify: the two folders were meant to be equivalent.

examples/src/multi-file/endpoints/find-users.ts
```diff
@@ -0,0 +1,19 @@
+import { api, endpoint, Optional, queryParam } from "@zenclabs/spot";
+
+@api()
+export class Api {
+  @endpoint({
+    method: "GET",
+    path: "/users"
+  })
+  getUsers(
```
fwouts: Do you mind renaming to `findUsers`? Same in the other example.

lib/src/generators/typescript/__snapshots__/axios-client.spec.ts.snap
```diff
@@ -201,6 +240,45 @@ export async function createUser(request: CreateUserRequest, authToken: string |
     }
 }
 
+export async function getUsers(): Promise<{
```
fwouts: It looks like the parameter is missing in the generated TypeScript library.

lib/src/lib.ts
```diff
@@ -94,6 +94,16 @@ export function pathParam(
...
 ) {}
 
+export function queryParam(
+  target: any,
+  propertyKey: string,
+  parameterIndex: number
+) {}
+
+export interface QueryParamDescription {
+  required: boolean;
```
fwouts: This could be removed now?

lib/src/models.ts
```diff
@@ -58,6 +59,12 @@ export interface Header {
   type: Type;
 }
 
+export interface QueryParamComponent {
+  name: string;
+  type: Type;
+  required: boolean;
```
fwouts: Instead of having a `required` field here, I would check whether `type.kind === "optional"`.

fwouts (changes requested):

examples/src/multi-file/endpoints/find-users.ts
danieltheodosius: there is no good reason, just want to have example, I will update it

examples/src/multi-file/endpoints/find-users.ts
danieltheodosius: changing it

lib/src/lib.ts
danieltheodosius: yes, it can be removed

lib/src/models.ts
danieltheodosius: changed

lib/src/generators/typescript/__snapshots__/axios-client.spec.ts.snap
danieltheodosius: true, I forgot to add it to the axios client

fwouts (approved):

PR #17, opened 2018-11-19
Add Prettier checks on CircleCI

fwouts:


.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
danieltheodosius, line 1: do we want to specify some configurations rather than leaving this empty?

danieltheodosius (approved):

.prettierrc
fwouts: I'd rather stay with the default :)

PR #22, opened 2018-11-21
Path param descriptions

lfportal:
Currently SPOT generates OpenAPI documentation with placeholder text for path param descriptions. This will change this to custom descriptions.

### DIFF
--- a/examples/src/multi-file/endpoints/get-user.ts
+++ b/examples/src/multi-file/endpoints/get-user.ts
@@ -7,7 +7,7 @@ export class Api {
     path: "/users/:userId"
   })
   getUser(
-    @pathParam userId: string
+    @pathParam() userId: string
   ): Promise<{
     name: string;
     age?: number;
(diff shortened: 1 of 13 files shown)


lib/src/parser.ts
```diff
@@ -238,7 +238,8 @@ function extractEndpoint(
               ? undefined
               : nextNonNamePositionRelative
           ),
-          type: VOID
+          type: VOID,
+          description: 'TODO'
```
fwouts: Instead of `TODO`, could we make this an empty string?

fwouts (changes requested):
Sweet!

lib/src/parser.ts
lfportal: Yeap

fwouts (approved):

examples/src/multi-file/endpoints/get-user.ts
```diff
@@ -7,7 +7,7 @@ export class Api {
     path: "/users/:userId"
   })
   getUser(
-    @pathParam userId: string
+    @pathParam() userId: string
```
danieltheodosius: should we make the example `multi-file` and `single-file` to be consistent? this one does not have description, while the one in the single file has description

danieltheodosius (approved):

PR #30, opened 2018-11-26
[TOOL-114] Split up parser.ts into multiple files

fwouts:
This is the very first step of cleaning up the parser.


danieltheodosius (commented):
changes look good, I wonder if we can break down the method `parseEndpointMethod` inside  `api-class-parser.ts`

danieltheodosius (approved):

PR #36, opened 2018-11-26
[TOOL-114] Parse types and literals when extracting decorators

fwouts:
This removes parsing logic from the upper layers and should make code a bit easier to follow.

It does mean that we lose the ability to output only the part of the code that failed to parse, and instead have to output the whole endpoint code. That's probably fine, as they should still be rather small. Arguably, it will also give useful context.

### DIFF
--- a/lib/src/parsing/components/endpoint-method.ts
+++ b/lib/src/parsing/components/endpoint-method.ts
@@ -1,7 +1,7 @@
 import * as ts from "typescript";
 import { Api, Endpoint } from "../../models";
 import { extractSingleDecorator } from "../decorators";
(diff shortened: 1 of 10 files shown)


lfportal (approved):
@fwouts this is looks really good. Not strictly related to this PR - at first glance I assumed `endpoint-method.ts` was supposed extract the endpoint's method (e.g. POST), but this is actually what `method.ts` does, purely from just looking at the file names. Within `components/` each file exports a function that extracts a very specific piece of information. The exceptions are `api-class.ts` and `endpoint-method.ts` which seem more like orchestration components. Could we perhaps make this distinction apparent in the folder/file structure?

fwouts:
Good point. I'll send out another PR for that.

PR #44, opened 2018-11-28
[TOOL-130] Add endpoint description

danieltheodosius:


package.json
```diff
@@ -59,6 +59,8 @@
     "postpack": "rm -f oclif.manifest.json",
     "prepack": "rm -rf build; tsc -b --force && rm -rf build/examples && oclif-dev manifest && oclif-dev readme",
     "test": "jest",
+    "prettier:list": "prettier --list-different \"**/*.js\" \"**/*.jsx\" \"**/*.ts\" \"**/*.tsx\"",
```
fwouts, line 62: Nit: these kind of changes generally deserve their own, separate PR. I'll let this one pass if you prefer not to :P

fwouts (approved):

package.json
danieltheodosius: I did remember before not to check it in for the previous PR, however I forgot to exclude it in this PR

lfportal (approved):

PR #45, opened 2018-11-28
[TOOL-131] Enable tagging endpoint

danieltheodosius:


lib/src/generators/contract/openapi2.ts
```diff
@@ -26,11 +27,22 @@ export function generateOpenApiV2(api: Api, format: "json" | "yaml") {
 export function openApiV2(api: Api): OpenApiV2 {
   return {
     swagger: "2.0",
-    tags: [
-      {
-        name: "TODO"
-      }
-    ],
+    tags: uniqBy(
```
lfportal: Can we extract this into a function?

lfportal (changes requested):

lib/src/generators/contract/openapi2.ts
danieltheodosius: yup, we can

lfportal (approved):

PR #49, opened 2018-11-30
[TOOL-135] Support Date and DateTime

danieltheodosius:


lfportal:
How will/should date/time formats be handled?

lfportal (approved):

PR #50, opened 2018-12-03
[TOOL-138] Fail on TypeScript compilation errors #tech

lfportal:
TypeScript errors are silently ignored by the file parser. It should fail instead if there are errors.

### DIFF
--- a/lib/src/parsing/file-parser.ts
+++ b/lib/src/parsing/file-parser.ts
@@ -18,6 +18,9 @@ const merge = require("lodash/merge");
  * This should be smart enough not to get stuck into an infinite loop for circular dependencies.
  */
 export async function parsePath(sourcePath: string): Promise<Api> {
+  if (!isValidTypeScript(sourcePath)) {
+    throw panic("TypeScript compilation error");
+  }
   const api: Api = parseRootFile(sourcePath);
   const errors = validate(api);
   if (errors.length > 0) {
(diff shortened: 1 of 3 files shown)


lib/src/parsing/file-parser.ts
```diff
@@ -198,3 +201,38 @@ function containsApiDeclaration(sourcePath: string): boolean {
...
+      console.error(
+        `${diagnostic.file.fileName} (${line + 1},${character + 1}): ${message}`
+      );
+    } else {
+      console.error(
+        `${ts.flattenDiagnosticMessageText(diagnostic.messageText, "\n")}`
+      );
+    }
+  });
+  return diagnostics.length > 0 ? false : true;
```
danieltheodosius: shoud just be `return diagnostics.length === 0`

lib/src/parsing/file-parser.ts
```diff
@@ -198,3 +201,38 @@ function containsApiDeclaration(sourcePath: string): boolean {
...
+  const program = ts.createProgram([sourcePath], {
+    target: ts.ScriptTarget.ESNext,
+    experimentalDecorators: true,
+    moduleResolution: ts.ModuleResolutionKind.NodeJs,
+    baseUrl: "./",
+    paths: {
+      "@airtasker/spot": ["./lib/src/lib"]
+    }
+  });
+  let diagnostics = program.getSemanticDiagnostics();
```
danieltheodosius: is this `getSemanticDiagnostics()` enough? Do we need to do `getSyntheticDiagnostics()`?

danieltheodosius (changes requested):

danieltheodosius (approved):

PR #51, opened 2018-12-03
Usage documentation updates for the readme

lfportal:


README.md
```diff
@@ -126,18 +334,24 @@ Generates the boilerplate for an API.
...
+- api.ts
+- tsconfig.json
+- package.json
 ```
 
 _See code: [cli/src/commands/init.js](https://github.com/airtasker/spot/blob/master/cli/src/commands/init.ts)_
+
 <!-- commandsstop -->
+
+```
```
danieltheodosius: there is an extra code line here

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+@api({
+  name: "My API",
+  description: "My really cool API"
+})
+class MyAPI {}
+```
+
+| Field         | Description                           |
+| ------------- | ------------------------------------- |
+| `name`        | (**required**) Name of the API        |
```
fwouts, line 98: Why are these two required? Just thinking in terms of reducing friction for users who don't need to generate documentation. Description isn't required by OpenAPI3 either: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#infoObject.

This is fine for now, but I'd like to address this in the future. It may be more of a philosophical question, but I think Spot should only require the absolute minimum, which is the structure of the API without any human-readable names or descriptions.

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
 ```
 
+## `@api`
+
+Define an API. This is required and must only be defined once:
+
+```TypeScript
+import { api } from "@airtasker/spot";
+
+@api({
```
fwouts, line 89: Now that `@api()` is used separately from `@endpoint()`, the syntax looks a bit unnecessarily complicated.

Something to address separately: maybe we actually just want to expose a conventional function `declareApi()`, so you do the following:
```typescript
declareApi({
  name: "My API",
  description: "My really cool API"
});
```

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+## `@endpoint`
+
+Define a HTTP endpoint for the API. An endpoint describes a particular HTTP action on a URL path:
+
+```TypeScript
+import { endpoint, genericError, header, pathParam, queryParam, request, response, specificError } from "@airtasker/spot";
+
+class MyUserEndpoints {
+  @endpoint({
```
fwouts, line 110: It may be useful to add a comment before each of these endpoint definitions, paraphrasing what the declaration does.

```suggestion
  // GET /users expects a mandatory `search_term` query parameter and returns a list of users.
  @endpoint({
```

Note #1: I wrote "mandatory", but I wonder if `search_term` should be `Optional<string>` to reflect real endpoints more accurately? It's rare for a search query parameter to be mandatory.

Note #2: Should we instead return a `UserListResponse`? Returning a list of responses doesn't feel very idiomatic.

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+interface CreateUserRequest {
+  firstName: string;
+  lastName: string;
+}
+
+interface CreateUserResponse {
+  success: boolean;
+}
+
+interface ApiError {
```
fwouts: Should this be renamed to `ApiErrorResponse`?

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+```
+
+| Field                | Default            | Description                                               |
+| -------------------- | ------------------ | --------------------------------------------------------- |
+| `method`             |                    | (**required**) [HTTP method](#suppported-http-methods)    |
+| `path`               |                    | (**required**) URL path                                   |
+| `description`        | `""`               | Description of the endpoint                               |
+| `requestContentType` | `application/json` | Content type of the request body                          |
+| `successStatusCode`  | `200`              | HTTP status code for a successful response                |
+| `tags`               |                    | Array of tags used for endpoint grouping in documentation |
```
fwouts: Nit: I personally find the term `List` less implementation-specific than `Array` (to me, "array" implies that indexes are important). Thoughts?

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+### `@request`
+
+Define a request body for requests that require one. This is commonly used for `POST` and `PUT` requests:
```
fwouts: Maybe also add a note that it isn't valid for `GET` requests.

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+  firstName: string;
+  lastName: string;
+}
+
+interface UpdateUserRequest {
+  lastName: string;
+}
+```
+
+### `@header`
```
fwouts, line 225: Would it be possible to add examples for each of these too, or mention that the examples above cover them?

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+Define a request header. `@header` can be used multiple times to define multiple headers.
+
+| Field         | Description                       |
+| ------------- | --------------------------------- |
+| `name`        | (**required**) Name of the header |
+| `description` | Description of the header         |
+
+### `@pathParam`
+
+Define path parameters that appear in the `@endpoint.path`. The parameter must be defined in the `@endpoint.path`. A `@pathParam` should be defined for each parameter defined in `@endpoint.path`.
```
fwouts: It may not be super obvious that `@endpoint.path` refers to `@endpoint({ path: ... })` so I'd suggest the following:
```suggestion
Define path parameters that appear in the `path` provided in `@endpoint()`. For example if the path is `/users/:userId`, the endpoint method must define a matching argument with `@pathParam() userId: string`.

Note: the name of the argument must match the name of the path parameter.
```

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+| Field         | Description                       |
+| ------------- | --------------------------------- |
+| `description` | Description of the path parameter |
+
+### `@queryParam`
+
+Define query parameters. `@queryParam` can be used multiple times to define multiple query parameters.
```
fwouts:
```suggestion
Define query parameters. `@queryParam` can be used multiple times to define multiple query parameters.

Note: the name of the argument must match the name of the query parameter.
```

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+| Field         | Description                        |
+| ------------- | ---------------------------------- |
+| `description` | Description of the query parameter |
+
+### `@specificError<T>`
+
+Define a known error for the endpoint. `@specificError` can be used multiple times to define multiple errors.
```
fwouts:
```suggestion
Define a known error for the endpoint. `@specificError` can be used multiple times to define multiple errors.

`T` must be replaced with the response type when the error occurs, for example `@specificError<UnauthenticatedErrorResponse>`.
```

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+| Field        | Description                                   |
+| ------------ | --------------------------------------------- |
+| `name`       | (**required**) Name of the error              |
+| `statusCode` | (**required**) HTTP status code for the error |
+
+### `@genericError<T>`
+
+Define a default error for the endpoint. This can only be used once for an `@endpoint`.
```
fwouts:
```suggestion
Define a default error for the endpoint. This can only be used once for an `@endpoint`.

`T` must be replaced with the response type when the error occurs, for example `@specificError<UnauthenticatedErrorResponse>`.
```

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+
+## Matcher Types
+
+| Type          | Description                                                                                             | Example                                                |
+| ------------- | ------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| `string`      | A string value                                                                                          |
+| `number`      | A number value                                                                                          |
+| `Int32`       | A 32-bit integer                                                                                        |
```
fwouts: The inconsistency between lower and uppercase types makes me wonder if we could rename our types to be lowercase instead... Alternatively, we could support the uppercase aliases, which already exist: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

README.md
```diff
@@ -57,48 +60,253 @@ This is work in progress as of 14 Nov 2018:
...
+| `Int64`       | A 64-bit integer                                                                                        |
+| `Float`       | A 32-bit floating point number                                                                          |
+| `Double`      | A 64-bit floating point number                                                                          |
+| `boolean`     | A boolean value                                                                                         |
+| `Date`        | [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html) string representation of a date      |
```
fwouts: These could use examples (as well as each of the above).

fwouts (approved):
This is really, really good. Thank you. A few more comments and I think we'll be ready to go.

Feel free to create tickets for the various issues I've raised, if you agree with them.

README.md
lfportal: Description makes sense to be optional, but I think name should be required. Without a name for the service/api, we can't realistically convert to other forms of contracts e.g. OpenAPI and Pact both require a service name/title (unless of course we give it a default name but could potentially be unsafe when working with a contract broker holding contracts for multiple services)

README.md
lfportal: Adding test-interactions will potentially introduce a fair amount of new syntax - which may or may not work well with the current syntax. It makes sense at that point that we evaluate the syntax overall to ensure both consistency and intuitiveness.

README.md
lfportal: Makes perfect sense

README.md
fwouts: Makes sense.

README.md
```diff
@@ -57,48 +60,344 @@ This is work in progress as of 14 Nov 2018:
...
+```
+
+| Field         | Description                       |
+| ------------- | --------------------------------- |
+| `name`        | (**required**) Name of the header |
+| `description` | Description of the header         |
+
+### `@pathParam`
+
+Define path parameters that appear in the `path` provided in `@endpoint()`. For example if the path is `/users/:userId`, the endpoint method must define a matching argument with `@pathParam() userId: string`:
```
fwouts:
```suggestion
Define path parameters that appear in the `path` provided in `@endpoint()`. For example if the path is `/users/:id`, the endpoint method must define a matching argument with `@pathParam() id: string`:
```

(to be consistent with the example below)

README.md
```diff
@@ -57,48 +60,344 @@ This is work in progress as of 14 Nov 2018:
...
+
+**Note**: the name of the argument must match the name of the query parameter.
+
+| Field         | Description                        |
+| ------------- | ---------------------------------- |
+| `description` | Description of the query parameter |
+
+### `@specificError<T>`
+
+Define a known error for the endpoint. `@specificError` can be used multiple times to define multiple errors. `T` must be replaced with the response type when the error occurs, for example `@specificError<UnauthenticatedErrorResponse>`:
```
fwouts:
```suggestion
Define a known error for the endpoint. `@specificError` can be used multiple times to define multiple errors. `T` must be replaced with the response type when the error occurs, for example `@specificError<UnauthorizedErrorResponse>`:
```

README.md
```diff
@@ -57,48 +60,344 @@ This is work in progress as of 14 Nov 2018:
...
+
+```TypeScript
+  //...
+  @endpoint({
+    method: "POST",
+    path: "/users",
+    description: "Create a user"
+  })
+  @specificError<UnauthenticatedErrorResponse>({
```
fwouts:
```suggestion
  @specificError<UnauthorizedErrorResponse>({
```

fwouts (approved):

README.md
```diff
@@ -57,48 +60,344 @@ This is work in progress as of 14 Nov 2018:
...
+    path: "/users",
+    description: "Retrieve all users"
+  })
+  getUsers(@queryParam({ description: "Search term" }) search_term: Optional<string>): UserResponse[] {
+    return response();
+  }
+  //...
+```
+
+**Note**: the name of the argument must match the name of the query parameter.
```
danieltheodosius, line 294: Is this true? I don't think there is any argument that needs to match the name of the query param. Maybe it is for path param

danieltheodosius (approved):

PR #59, opened 2018-12-12
Refactored axios client generation to expose a configurable class

lfportal:


lib/src/generators/typescript/axios-client.ts
```diff
@@ -63,28 +69,74 @@ export function generateAxiosClientSource(api: Api): string {
       ),
       ts.createStringLiteral("./validators")
     ),
-    ...Object.entries(api.endpoints).map(([endpointName, endpoint]) =>
-      generateEndpointFunction(api, endpointName, endpoint)
+    generateSpotApiOptionsInterface(),
+    ts.createClassDeclaration(
```
danieltheodosius: just need to extract this to a function

danieltheodosius (approved):

PR #61, opened 2018-12-14
Allow usage of interface inheritance

fwouts:
This adds support for extending interfaces.

For example:
```typescript
export interface BaseError {
  message: string;
}

export interface ForbiddenError extends BaseError {
  signedInAs: string;
}
```


MitchWeiss:
I don't understand most of this code but I think this is awesome! Can't wait to use it

lfportal (approved):
looking good

PR #65, opened 2018-12-20
Implement a mock server based on the API contract

fwouts:


danieltheodosius (commented):
You might want to include a base url for the server. Current axios client does have a constructor `SpotApi` that takes a config object with `baseUrl` option. Because most contracts do not specify `/api/v2`.

fwouts:
Thanks @danieltheodosius, done. I called this `pathPrefix` instead, as I think `baseUrl` more often refers to the whole base including the domain.

lfportal (commented):
Can the mock server be extended to return headers defined in the contract?

fwouts:
@lfportal we don't define response headers at this time :)

lfportal (approved):

PR #70, opened 2019-01-10
New syntax parsing

lfportal:
Introduces a new syntax for Spot. This PR only includes parsing of the raw data. Complex validation logic is out of scope.


libnext/src/models/nodes.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export interface RequestNode {
+  headers: HeaderNode[];
+  pathParams: PathParamNode[];
+  queryParams: QueryParamNode[];
+  body?: BodyNode;
+}
+
+export interface ResponseNode extends DefaultResponseNode {
+  status: number;
```
fwouts, line 40: Why not `status?` in `DefaultResponseNode` and just have one type?

libnext/src/models/types/kinds.ts
```diff
@@ -0,0 +1,16 @@
+export enum Kind {
+  Null = "null",
```
fwouts: Styling note: personally, I prefer enum cases to be `UPPERCASE` as it makes it easier to separate them from types. Up to you though.

libnext/src/models/types/primitive-types.ts
```diff
@@ -0,0 +1,36 @@
+import { Kind } from "./kinds";
+
+/** Primitive types native to the typescript language */
```
fwouts:
```suggestion
/** Primitive types native to the TypeScript language. */
```

libnext/src/models/types/special-types.ts
```diff
@@ -0,0 +1,34 @@
...
+    kind: Kind.Union,
+    types: unionTypes
+  };
+}
+
+export interface ReferenceType {
+  kind: Kind.TypeReference;
+  referenceKind: Kind;
+  name: string;
+  location: string;
```
fwouts, line 21: Could you document this? What does a location look like?

libnext/src/models/types/kinds.ts
```diff
@@ -0,0 +1,16 @@
+export enum Kind {
```
fwouts: How about calling this `TypeKind`?

libnext/src/parsers/nodes/api-parser.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+import { parseApi } from "./api-parser";
+
+describe("@api parser", () => {
+  const apiClassName = "MyApi";
+  const apiName = "My API";
+
+  test("parses all information", () => {
+    const apiDescription = "Some description";
+    const content = `
+      /** ${apiDescription} */
```
fwouts: I'd feel a bit more confident in these tests if we didn't use string interpolation (just hardcoding the constants, duplicating them here). I know that's less code reuse, but these are tests and we should try to minimise unnecessary logic as much as we can :)

libnext/src/parsers/nodes/default-response-parser.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export function parseDefaultResponse(
+  method: MethodDeclaration
+): DefaultResponseNode {
+  method.getDecoratorOrThrow("defaultResponse");
+  const headersParameter = methodParamWithDecorator(method, "headers");
+  const bodyParameter = methodParamWithDecorator(method, "body");
+  const description = extractJsDocComment(method);
+  const headers =
+    headersParameter === undefined ? [] : parseHeaders(headersParameter);
```
fwouts:
```suggestion
    headersParameter ? parseHeaders(headersParameter) : [];
```

libnext/src/models/nodes.ts
lfportal: Whilst it may be a bit trivial, it does make the internal representation a bit more typesafe

fwouts: I think I understand why I was confused. `ResponseNode` is not semantically extending `DefaultResponseNode`: they're separate concepts. You're using `extends` as a convenient way to reuse code. I wonder if instead the code should be duplicated?

Either way, I think it's worth adding a bit of JSDoc to explain what each of these two types means.

fwouts (approved):
Nice work 🎉

PR #73, opened 2019-01-15
Use assert-never NPM package instead of our own copy

fwouts:
This was previously necessary because our `tsconfig.json` was messy. The fix in https://github.com/airtasker/spot/commit/f63abb4c073fc0451116da7ac80b8bba0726b28c made it simpler to remove this hack.


lfportal (approved):
looks good

PR #74, opened 2019-01-15
Parse request path into a structured list and make RequestNode required

fwouts:
This will make it a lot easier to write generators based on the internal data models.


fwouts:
Abandoning as it makes more sense to extract information from the path in a helper function used by generators. There's no need for the parser to do this.

PR #85, opened 2019-01-21
Replace confusing usage of require() with import statements

fwouts:


lib/src/generators/contract/openapi2.ts
```diff
@@ -1,17 +1,17 @@
 import { HttpContentType } from "@airtasker/spot";
 import assertNever from "assert-never";
 import * as YAML from "js-yaml";
+import compact from "lodash/compact";
+import defaultTo from "lodash/defaultTo";
+import pickBy from "lodash/pickBy";
+import uniqBy from "lodash/uniqBy";
```
lfportal, line 7: Can we import from the root module instead? `import { compact, defaultTo, pickBy ...... } from "lodash";`

lfportal (changes requested):

lib/src/generators/contract/openapi2.ts
fwouts: We could, but this should help keep the bundle size smaller (because we don't need to rely on the compiler being smart enough to tree-shake out the unused functions).

lfportal (approved):

PR #96, opened 2019-01-30
Infer discriminator in OpenAPI 3 generator

fwouts:


lib/src/generators/contract/openapi3.ts
```diff
@@ -45,25 +46,36 @@ export function openApiV3(contractDefinition: ContractDefinition): OpenApiV3 {
...
+                    contractDefinition.types,
+                    endpoint.defaultResponse
+                  )
+                }
               : {}),
             ...endpoint.responses.reduce<{
               [statusCode: string]: OpenAPIV3Body;
             }>((acc, responseNode) => {
-              acc[responseNode.status.toString(10)] = response(responseNode);
+              acc[responseNode.status.toString(10)] = response(
```
committedteadrinker, line 75: Where does the '10' come from?

committedteadrinker (changes requested):

lib/src/generators/contract/openapi3.ts
fwouts: It's the base in which to represent the number. It actually defaults to base 10, so I could remove it if you think that's more readable.

committedteadrinker: ohh never knew this, TIL. Thanks!

committedteadrinker (approved):

PR #98, opened 2019-02-01
fix README

Kastet:


README.md
```diff
@@ -43,6 +42,13 @@ interface CreateUserResponse {
 
 For available methods, please check [here](https://github.com/airtasker/spot/wiki/Spot-available-methods)
 
+
+With [yarn](https://yarnpkg.com/en/docs/usage) installed and initialised add `@airtasker/spot` to your project:
+
+```sh
+yarn add @airtasker/spot
```
fwouts, line 49: `npx` below doesn't require you to manually install the package. Are you sure this is necessary?

lfportal: when running `init` the CLI executes `yarn add ...` - we should try and decouple the project from the package manager.

fwouts: True. I think we could also just default to NPM (the official option) and allow using `yarn` via an additional command line argument.

fwouts (approved):

PR #99, opened 2019-02-04
Add @securityHeader decorator used to indicate the global authentication header

fwouts:
This adds:
- a @securityHeader decorator to be used in a top-level API declaration
- logic for parsing @securityHeader
- OpenAPI 2 & 3 output

For reference, please check:
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

### DIFF
--- a/lib/src/cleansers/nodes/security-header-cleanser.ts
+++ b/lib/src/cleansers/nodes/security-header-cleanser.ts
@@ -1,0 +1,12 @@
+import { SecurityHeaderDefinition } from "../../models/definitions";
+import { SecurityHeaderNode } from "../../models/nodes";
+
+export function cleanseSecurityHeader(
+  securityHeaderNode: SecurityHeaderNode
+): SecurityHeaderDefinition {
+  const name = securityHeaderNode.name.value;
+  const description =
+    securityHeaderNode.description && securityHeaderNode.description.value;
+  const type = securityHeaderNode.type;
+  return { name, description, type };
+}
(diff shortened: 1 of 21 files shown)


lib/src/parsers/nodes/security-header-parser.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param a field declaration
+ */
+export function parseSecurityHeader(
+  property: PropertyDeclaration
+): Locatable<SecurityHeaderNode> {
+  const decorator = property.getDecoratorOrThrow("securityHeader");
+
+  const name = extractPropertyNameLocatable(property);
+  const description = extractJsDocCommentLocatable(property);
+  const type = parseType(property.getTypeNodeOrThrow());
```
lfportal, line 22: Is there a verifier for the security header type? Allowed values should be numbers, strings or number/string literals

lfportal (changes requested):

lib/src/parsers/nodes/security-header-parser.ts
fwouts: Good point. Shouldn't it be only `string`?

lfportal (approved):

lib/src/cleansers/nodes/security-header-cleanser.ts
```diff
@@ -0,0 +1,12 @@
+import { SecurityHeaderDefinition } from "../../models/definitions";
+import { SecurityHeaderNode } from "../../models/nodes";
+
+export function cleanseSecurityHeader(
+  securityHeaderNode: SecurityHeaderNode
+): SecurityHeaderDefinition {
+  const name = securityHeaderNode.name.value;
+  const description =
+    securityHeaderNode.description && securityHeaderNode.description.value;
```
Kastet, line 9: this short-circuit evaluation looks like magic :)

PR #101, opened 2019-02-08
Test annotation for endpoints

lfportal:
Added the `@test` annotation for specifying an interaction for testing purposes. This includes a parser and cleanser components. Verification rules for the tests are not enforced in this PR.


lib/src/models/definitions.ts
```diff
@@ -72,3 +73,34 @@ export interface BodyDefinition {
   description?: string;
   type: DataType;
 }
+
+export interface TestDefinition {
+  description?: string;
+  states?: TestStateDefinition[];
```
fwouts: Shouldn't this be required, but with an empty list when absent?

lib/src/models/definitions.ts
```diff
@@ -72,3 +73,34 @@ export interface BodyDefinition {
...
+  request?: TestRequestDefinition;
+  response: TestResponseDefinition;
+}
+
+export interface TestStateDefinition {
+  name: string;
+  params?: { name: string; expression: DataExpression }[];
```
fwouts: ditto (also each one below)

lib/src/models/nodes.ts
```diff
@@ -73,3 +74,34 @@ export interface BodyNode {
   description?: Locatable<string>;
   type: DataType;
 }
+
+export interface TestNode {
+  description?: Locatable<string>;
+  states?: TestStateNode[];
```
fwouts, line 87: ditto

lib/src/parsers/utilities/expression-parser.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+            }
+          ]
+        },
+        { kind: "null" }
+      ],
+      kind: "array"
+    });
+  });
+
+  test("parses a object literal expression", () => {
```
fwouts:
```suggestion
  test("parses an object literal expression", () => {
```

lib/src/parsers/utilities/parser-utility.ts
```diff
@@ -173,6 +186,83 @@ export function extractNumberProperty(
...
+    }
+    const value = property.getInitializerIfKindOrThrow(
+      ts.SyntaxKind.ArrayLiteralExpression
+    );
+    const location = property.getSourceFile().getFilePath();
+    const line = value.getStartLineNumber();
+
+    return { value, location, line };
+  } else {
+    return;
```
fwouts: Maybe `return undefined` for clarity? You could also just remove this entirely (`return undefined` is the default behaviour).

fwouts (changes requested):
Nice work. Would be good to tweak it to add type safety later :)

lib/src/models/nodes.ts
lfportal: Planning to introduce `Locatable` to some of the `@test` related nodes in the future

lib/src/models/nodes.ts
```diff
@@ -80,3 +81,34 @@ export interface BodyNode {
...
+  request?: Locatable<TestRequestNode>;
+  response: Locatable<TestResponseNode>;
+}
+
+export interface TestStateNode {
+  name: string;
+  params?: { name: string; expression: DataExpression }[];
```
fwouts, line 94: Should these (and below in TestRequestNode and TestResponseNode) also not be optional?

fwouts (approved):

PR #102, opened 2019-02-11
Verification component to verify DataExpressions with JSON schemas.

lfportal:
The `json-schema-verifier` compares `DataExpression`s  with JSON schemas (converted from `DataTypes` using the existing JSON schema generator logic.

### DIFF
--- a/lib/src/verifiers/utilities/json-schema-verifier.spec.ts
+++ b/lib/src/verifiers/utilities/json-schema-verifier.spec.ts
@@ -18,118 +18,12 @@ import {
   TypeKind,
   UnionType
 } from "../../models/types";
-import { dataExpressionToJson, verifyJsonSchema } from "./json-schema-verifier";
+import { verifyJsonSchema } from "./json-schema-verifier";
 
 describe("json schema verifier", () => {
-  describe("dataExpressionToJson", () => {
-    it("converts null expressions to null", () => {
-      const data: NullExpression = {
-        kind: TypeKind.NULL
-      };
-      expect(dataExpressionToJson(data)).toBeNull();
-    });
-
-    it("converts boolean expressions to boolean values", () => {
-      const data: BooleanExpression = {
-        kind: TypeKind.BOOLEAN_LITERAL,
-        value: true
-      };
-      expect(dataExpressionToJson(data)).toBe(true);
-    });
-
-    it("converts string expressions to string values", () => {
-      const data: StringExpression = {
-        kind: TypeKind.STRING_LITERAL,
-        value: "hello there"
-      };
-      expect(dataExpressionToJson(data)).toBe("hello there");
-    });
-
-    it("converts number expressions to number values", () => {
-      const data: NumberExpression = {
-        kind: TypeKind.NUMBER_LITERAL,
-        value: 456
-      };
-      expect(dataExpressionToJson(data)).toBe(456);
-    });
-
-    it("converts array expressions to array objects", () => {
-      const data: ArrayExpression = {
-        kind: TypeKind.ARRAY,
-        elements: [
-          {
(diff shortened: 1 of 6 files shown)


lib/src/verifiers/utilities/json-schema-verifier.ts
```diff
@@ -0,0 +1,54 @@
...
+    )
+  };
+  const validateFn = ajv.compile(schema);
+  const valid = validateFn(dataExpressionToJson(data));
+  if (!valid) {
+    throw new Error(`Invalid: ${ajv.errorsText(validateFn.errors)}`);
+  }
+}
+
+export function dataExpressionToJson(data: DataExpression): any {
```
fwouts: This isn't really JSON Schema specific. What do you think about extracting it out into its own file?

Also, this isn't really generating JSON (the N stands for "Notation"). It's generating JavaScript values. How about calling it `valueFromDataExpression()` or something of the sort?

lib/src/verifiers/utilities/json-schema-verifier.ts
```diff
@@ -0,0 +1,54 @@
...
+import { TypeNode } from "../../models/nodes";
+import { DataExpression, DataType, TypeKind } from "../../models/types";
+import Ajv = require("ajv");
+
+export function verifyJsonSchema(
+  dataType: DataType,
+  data: DataExpression,
+  typeStore: TypeNode[]
+) {
+  const ajv = new Ajv();
```
fwouts: That's a fairly cryptic library name... How about naming the variable `jsonSchemaValidator`?

lib/src/verifiers/utilities/json-schema-verifier.spec.ts
```diff
@@ -0,0 +1,603 @@
...
+        const data: NumberExpression = {
+          kind: TypeKind.NUMBER_LITERAL,
+          value: 53
+        };
+        expect(() => verifyJsonSchema(dataType, data, [])).toThrow();
+      });
+    });
+
+    describe("array schema", () => {
+      it("validates", () => {
```
fwouts: I'd also like to see that it doesn't validate `undefined` and it does validate `[]`.

lib/src/verifiers/utilities/json-schema-verifier.spec.ts
```diff
@@ -0,0 +1,603 @@
...
+              kind: TypeKind.NUMBER_LITERAL,
+              value: 54
+            }
+          ]
+        };
+        expect(() => verifyJsonSchema(dataType, data, [])).toThrow();
+      });
+    });
+
+    describe("object schema", () => {
```
fwouts, line 254: worth checking with an empty object as well

lib/src/verifiers/utilities/json-schema-verifier.spec.ts
```diff
@@ -0,0 +1,603 @@
...
+            {
+              kind: TypeKind.STRING_LITERAL,
+              value: "bye"
+            }
+          ]
+        };
+        expect(() => verifyJsonSchema(dataType, data, [])).not.toThrow();
+      });
+
+      it("invalidates", () => {
```
fwouts: I think it would be worth having slightly more specific names for each test case, e.g. `invalidates incorrect item type in list` or `validates empty list`.

fwouts (approved):
A few minor changes, overall approach looks good.

lib/src/verifiers/utilities/json-schema-verifier.spec.ts
lfportal: We can test for `[]` but there's no way to test `undefined`. The Array itself is being validated - `undefined` is not an array, so this situation should never happen.

PR #103, opened 2019-02-12
Dependency upgrades

lfportal:
Upgraded dependencies containing no obvious breaking changes.


lib/src/generators/contract/openapi2-schema.ts
```diff
@@ -91,8 +91,8 @@ export function openApi2TypeSchema(type: DataType): OpenAPI2SchemaType {
         return {
           type: "string",
           enum: compact(
-            type.types.map(
-              t => (t.kind === TypeKind.STRING_LITERAL ? t.value : null)
+            type.types.map(t =>
```
fwouts, line 94: How come this had to change?

fwouts (approved):

lib/src/generators/contract/openapi2-schema.ts
lfportal: `prettier` upgrade

PR #106, opened 2019-02-19
Add a `spot lint` command with two example rules

fwouts:
The `lint` command will help ensure that an API follows a set of reasonable requirements.

We're starting by defining linting rules based on Airtasker's API design guidelines, but the goal is to make it an extendable set over time.

The command will be invoked with:
```
spot lint contract.ts
```

The plan is to later introduce a linting configuration file, e.g. `.spotlintrc` which would be automatically found based on the contract's path. Or we could make that provided via a flag. This will be defined later.

## Implementation details

As part of implementing the `lint` command, we need a way to access the `ContractNode`, which has useful debugging information such as the source location. The `safeParse()` function is therefore updated to return both the source node and the cleansed node.


fwouts:
I should probably add some tests, actually 😄

driesg:
really cool and great idea. Did you consider leveraging ESLint / TSLint?

fwouts:
@driesg Good point, usually people think of "linting" as "syntax linting" whereas this is more "contract linting". I'm not sure the two concepts should be merged into the same thing. Maybe we should have a different name for this command? Or maybe this should be merged with the `validate` command?

driesg:
ESLint is also used for "bad patterns" so I think it would be fine to use that. We could also leverage their error/warning/off functionality for rules and the ability to run inside your IDE (without a separate command or having to save)

It seems easier to maintain a set of rules than rules and a linter.

But to answer your question: I think calling it "linting" is fine. 

~Another alternative would be Speccy with custom rules http://speccy.io/rules/2-custom-rulesets.~ That's at the openapi level. Not Spot, so not an option

fwouts:
@driesg I suspect integrating with ESLint isn't as simple as it may seem. Would you be OK with sticking to this approach in the medium term?

driesg:
yes of course! Sorry should've been clearer about that

lib/src/linting/rules/has-request-payload.ts
```diff
@@ -0,0 +1,57 @@
...
+        endpoint.value.name.value
+      } should not have a request payload as its method is ${
+        endpoint.value.method.value
+      }`,
+      source: endpoint
+    }));
+};
+
+function isMutationEndpoint(endpoint: Locatable<EndpointNode>) {
+  switch (endpoint.value.method.value) {
```
lfportal, line 45: can be simplified to something like `MUTATION_METHODS = ["POST", "PUT", "PATCH"]` `MUTATION_METHODS.includes(endpoint.value.method.value)`

lfportal (approved):

lib/src/linting/rules/has-request-payload.ts
fwouts: I like it! But it's not as efficient and I'd argue a bit easier to break (you'd need `MUTATION_METHODS: HttpMethod[] = ["POST", "PUT", "PATCH"]` to make it safer, and that's less readable). I'll keep the switch syntax for now.

PR #111, opened 2019-02-22
Introduce Float and String type aliases

fwouts:
This will make it easier to use consistent types (all uppercase) in Spot definitions, and more explicit when a field is a meant to be a floating point number vs a general "number".

Just food for thought, not sure if this is a good idea 🙂


danieltheodosius:
This is interesting that you use Float instead of Number here. Isn't it going to add confusion in regards of using `Float` and `number`?

lfportal:
Does this affect the parser?

fwouts:
@danieltheodosius You're right, it means there are two ways of saying the same thing (`Float` is the same as `number`). I'm not sure whether that's better or worse. Maybe we could have a lint rule that discourages using `number`?

@lfportal This adds new primitive types for the parser, but otherwise shouldn't break any existing contracts. Not completely sure what's your question?

lfportal:
> @danieltheodosius You're right, it means there are two ways of saying the same thing (`Float` is the same as `number`). I'm not sure whether that's better or worse. Maybe we could have a lint rule that discourages using `number`?
> 
> @lfportal This adds new primitive types for the parser, but otherwise shouldn't break any existing contracts. Not completely sure what's your question?

The parser currently provides special treatment for aliased types. If the parser encounters a type alias, it will first check if the name is `Integer`, or `Date` etc and treat those as Spot internal types - otherwise it will treat it as a user defined type. I think the parser may need to be updated to add `String` and any newly introduced internal types to its list of special cases.

fwouts:
@lfportal I think that's what I did in https://github.com/airtasker/spot/pull/111/files#diff-a45bf354cc2d0a0ec0c4c44248c535f7R89, did you have something else in mind?

lfportal:
> @lfportal I think that's what I did in https://github.com/airtasker/spot/pull/111/files#diff-a45bf354cc2d0a0ec0c4c44248c535f7R89, did you have something else in mind?

Yup sorry, you're right, missed that.

lfportal (approved):

danieltheodosius:
@fwouts we need to standardise the type, we have `Date`, `Integer`, and `DateTime`, and we should still support but discourage the use of `number` when we introduce `Float.

danieltheodosius (approved):

PR #112, opened 2019-03-11
Test annotation verifier

lfportal:


lib/src/parsers/utilities/parser-utility.ts
```diff
@@ -281,11 +306,31 @@ export function extractDecoratorFactoryConfiguration(
...
+): ObjectLiteralExpression | undefined {
+  if (decorator.getArguments().length >= 2) {
+    const argument = decorator.getArguments()[1];
+    if (TypeGuards.isObjectLiteralExpression(argument)) {
+      return argument;
+    } else {
+      throw new Error(`expected object literal`);
+    }
+  } else {
+    return undefined;
```
fwouts: How about returning `null` instead? I generally think of `undefined` as not returning anything, whereas here you're returning the absence of a value.

lib/src/verifiers/nodes/test-verifier.spec.ts
```diff
@@ -0,0 +1,397 @@
...
+        pathParams: []
+      }),
+      response: fakeLocatable({
+        status: fakeLocatable(204)
+      }),
+      options: {
+        allowInvalidRequest: true
+      }
+    });
+    expect(verifyTestNode(testNode, endpointNode, [])).toHaveLength(1);
```
fwouts: Why not expect a specific error message? (same below)

My reasoning is that it's good to check that an error is returned, but even better to check that the error is what we expect it to be.

fwouts (approved):
🎉

PR #139, opened 2019-03-15
Test runner

lfportal:
The test runner runs contract tests for Spot contracts.


cli/src/commands/test.ts
```diff
@@ -0,0 +1,69 @@
...
+  static description = "Test a Spot contract";
+
+  static examples = ["$ spot test api.ts -u http://localhost:3000"];
+
+  static args = [
+    {
+      name: ARG_API,
+      required: true,
+      description: "path to Spot contract",
+      hidden: false
```
fwouts, line 23: What is this for? Is the default behaviour `hidden: true`?

cli/src/commands/test.ts
```diff
@@ -0,0 +1,69 @@
...
+      char: "u",
+      description: "Base URL"
+    }),
+    stateUrl: flags.string({
+      char: "s",
+      description: "State change URL"
+    }),
+    testFilter: flags.string({
+      char: "t",
+      description: "Filter by endpoint and test"
```
fwouts, line 40: Worth adding examples?

cli/src/commands/test.ts
```diff
@@ -0,0 +1,69 @@
...
+  static flags = {
+    help: flags.help({ char: "h" }),
+    url: flags.string({
+      required: true,
+      char: "u",
+      description: "Base URL"
+    }),
+    stateUrl: flags.string({
+      char: "s",
+      description: "State change URL"
```
fwouts, line 36: It'd be good to document somewhere what this endpoint should look like. I wonder if we could find a good way to describe that API endpoint... 😛

cli/src/test-utils/test-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+  mute,
+  success,
+  warn
+};
+
+export default TestLogger;
+
+function error(message: string) {
+  // tslint:disable-next-line: no-console
+  console.log(chalk.red(message));
```
fwouts, line 13: Should this use `console.error` so it goes onto `process.stderr` instead of `process.stdout`? same thing with `warn()`. Or does it make more sense to use a single stream?

cli/src/test-utils/test-runner.spec.ts
```diff
@@ -0,0 +1,209 @@
+import nock from "nock";
```
fwouts, line 1: This library is fun!

cli/src/test-utils/test-runner.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        201,
+        { name: "My Company" },
+        { Location: `${baseUrl}/companies/abc` }
+      )
+      .post("/state")
+      .query({ action: "teardown" })
+      .reply(200);
+
+    const result = await runTest(contract, stateUrl, baseUrl);
+    expect(scope.isDone()).toEqual(true);
```
fwouts: FYI I think the more common approach is to use `toBe()` for primitive values: https://jestjs.io/docs/en/using-matchers

cli/src/test-utils/test-runner.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      .post("/state", { name: "a company exists", params: { id: "abc" } })
+      .query({ action: "setup" })
+      .reply(200)
+      .post("/state")
+      .query({ action: "teardown" })
+      .reply(200);
+
+    const result = await runTest(contract, stateUrl, baseUrl);
+    expect(scope.isDone()).toEqual(true);
+    expect(result).toEqual(false);
```
fwouts: I think it would be valuable to also test for specific error messages. Could be done in a separate PR though :)

cli/src/test-utils/test-runner.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  test("response body mismatch", async () => {
```
fwouts: Do you already have a test that adds an extra field on top of the expected response? Should it pass? (I know we said it should with Pact, but now that we're rolling out our own testing framework, I wonder if we should question this statement)

cli/src/common/utils.ts
```diff
@@ -0,0 +1,8 @@
+export async function asyncForEach<T>(
```
fwouts: Worth adding some JSDoc here, as one could expect that it runs in parallel.

Note before you address this, I have another suggestion to delete this function altogether below.

cli/src/test-utils/test-runner.ts
```diff
@@ -0,0 +1,300 @@
...
+
+export async function runTest(
+  definition: ContractDefinition,
+  stateUrl: string,
+  baseUrl: string,
+  testFilter?: TestFilter
+): Promise<boolean> {
+  let allPassed = true;
+
+  await asyncForEach(definition.endpoints, async endpoint => {
```
fwouts: Why not use a simple for loop?

cli/src/test-utils/test-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const TestLogger = {
+  error,
+  log,
+  mute,
+  success,
+  warn
+};
+
+export default TestLogger;
```
fwouts: I would personally recommend not using default exports (using a named import instead with `export const TestLogger`), to ensure that it's called the same in the file that defines it and it files that import it.

cli/src/test-utils/test-runner.ts
```diff
@@ -0,0 +1,300 @@
...
+    throw new Error(
+      `a response with status ${
+        test.response.status
+      } was not found and a default response has not been defined`
+    );
+  }
+  return correlatedResponse;
+}
+
+function generateAxiosConfig(
```
fwouts, line 135: I'd like to see some JSDoc on each of these functions (not necessarily super extensive)

fwouts (changes requested):
Looks really good 🎉

cli/src/commands/test.ts
lfportal: `hidden: true` is the default and will hides the argument from the `help` menu

cli/src/commands/test.ts
lfportal: I'll add a section about it in the Readme for now

cli/src/test-utils/test-logger.ts
lfportal: I'm impartial to this. Is there a tangible benefit?

fwouts: It can be useful if you need to distinguish between expected output of a CLI tool (to pipe it to another tool) vs error messages, but I don't think that's the case here. Leave it as is.

cli/src/test-utils/test-runner.spec.ts
lfportal: Good point, I'll add a test for this. When deciding whether or not to pass something for clients - in general I think we should simply ask the question: Will this break my client integration? If the answer is no, then we allow it

cli/src/test-utils/test-runner.ts
lfportal: `forEach` syntax offers improved readability. This is even especially important in this case where we have a nested loop - cognitive overhead of keeping track of multiple indexes

cli/src/commands/test.ts
lfportal: in a seperate PR

fwouts: Or maybe in the wiki? I think the time has come to spend effort on the documentation for other teams (and curious outsiders). We should probably have a sync about this next week.

fwouts (approved):

committedteadrinker (approved):

PR #176, opened 2019-04-04
Provider state initialisation phase

lfportal:
Providers may require common data to be setup prior to running any test. E.g. super user, country data etc. The provider state initialisation phase allows any non specific pre-test setup to be executed.

### DIFF
--- a/cli/src/test-utils/test-runner.ts
+++ b/cli/src/test-utils/test-runner.ts
@@ -82,7 +82,7 @@ async function executeTest(
   typeStore: TypeNode[]
 ): Promise<boolean> {
   if (
-    (await executeStateInitialisation(stateUrl)) &&
+    (await executeStateInitialization(stateUrl)) &&
     (await executeStateSetup(test, stateUrl))
   ) {
     const testResult = await executeRequestUnderTest(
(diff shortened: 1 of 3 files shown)


cli/src/test-utils/test-runner.ts
```diff
@@ -222,6 +225,35 @@ async function executeRequestUnderTest(
...
 }
 
+/**
+ * Execute the state initialisation request.
+ *
+ * @param stateUrl state change URL
+ */
+async function executeStateInitialisation(stateUrl: string): Promise<boolean> {
+  try {
+    TestLogger.mute("\tPerforming state initialisation request");
```
committedteadrinker: what does `mute` do?

committedteadrinker (approved):
Long live 'initialisation' 🇦🇺

cli/src/test-utils/test-runner.ts
lfportal: prints in a neutral grey colour

fwouts (approved):
I actually think we should follow international conventions and use 🇺🇸 English in our code... Just like French developers don't write their open source libraries in 🇫🇷 French.

Kastet (approved):

PR #183, opened 2019-04-11
Reference types resolver bug

lfportal:
Reference types are not being resolved correctly when parsing root level complex types (e.g. arrays, unions) on request/response bodies. Root level reference types will now simply run through the same recursive reference type resolver as non-root level reference types. This removes the root level unique reference optimisation parsing which provided a minor performance impact at best. 

The OpenAPI generator is now configured to alphabetise definitions/components to ensure snapshots (testing) are more immune to implementation changes.

### DIFF
--- a/lib/src/generators/contract/openapi2.ts
+++ b/lib/src/generators/contract/openapi2.ts
@@ -70,12 +70,22 @@ export function openApiV2(contractDefinition: ContractDefinition): OpenApiV2 {
       };
       return acc;
     }, {}),
-    definitions: contractDefinition.types.reduce<{
-      [typeName: string]: OpenAPI2SchemaType;
-    }>((acc, typeNode) => {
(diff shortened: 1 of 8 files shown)


lib/src/generators/contract/openapi2.ts
```diff
@@ -70,12 +70,22 @@ export function openApiV2(contractDefinition: ContractDefinition): OpenApiV2 {
...
       return acc;
     }, {}),
-    definitions: contractDefinition.types.reduce<{
-      [typeName: string]: OpenAPI2SchemaType;
-    }>((acc, typeNode) => {
-      acc[typeNode.name] = openApi2TypeSchema(typeNode.type);
-      return acc;
-    }, {}),
+    definitions: contractDefinition.types
+      .sort((a, b) => {
```
committedteadrinker: Please use better description on what these variables are.

committedteadrinker (approved):

PR #184, opened 2019-04-11
Test runner query string serialization

lfportal:
Spot expects query string params to be serialized using the `deepObject` strategy from the OpenAPI spec. The test runner is now configured to do so.


committedteadrinker (approved):

medric (approved):
LGTM 💯

PR #209, opened 2019-05-03
Test runner execution timer

lfportal:
- Added test execution timing to each phase of testing
- Modified `mute` logging to be more visible

### DIFF
--- a/cli/src/test-utils/test-logger.ts
+++ b/cli/src/test-utils/test-logger.ts
@@ -3,7 +3,6 @@ import chalk from "chalk";
 export const TestLogger = {
   error,
   log,
-  mute,
   success,
   warn
 };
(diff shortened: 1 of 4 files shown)


cli/src/test-utils/test-runner.spec.ts
```diff
@@ -35,7 +35,7 @@ describe("test runner", () => {
     expect(result).toBe(true);
   });
 
-  test("single provider state", async () => {
+  test.only("single provider state", async () => {
```
fwouts: Revert this?

cli/src/test-utils/test-logger.ts
```diff
@@ -18,7 +18,7 @@ function log(message: string) {
 }
```
committedteadrinker: I'm gonna comment on this `chalk.white` why `white`? Why can't it be the default colour on the console's theme itself?

fwouts (approved):

committedteadrinker (approved):

PR #216, opened 2019-05-08
Introduce a type unicity verifier and fix endpoint unicity verifier

fwouts:
We didn't have a way to ensure that the same type name was used twice.

It turns out, our endpoint name verifier was also incorrect because it compared Locatables and not strings.

Adding unit tests to ensure we never break this again :)


lfportal (approved):
Nice pickup on the incorrect endpoint names.

PR #219, opened 2019-05-09
Introduce Int32 and Int64 type aliases

fwouts:
This reshuffles how we handle number types in Spot:
- NUMBER is renamed to FLOAT internally for clarity (still mapped to both `number` and `Float` in TypeScript)
- INTEGER is renamed to INT32 internally (mapped to both `Integer` and the new `Int32` TypeScript alias)
- INT64 is introduced (and mapped to the new `Int64` TypeScript alias)

There are no breaking changes, as `number`, `Float` and `Integer` are handled in the same way as before.

See relevant discussion internally for context: https://github.com/airtasker/api-guidelines/pull/21

Happy to discuss, there are probably better approaches.


lib/src/models/types/index.ts
```diff
@@ -32,9 +32,9 @@ export type DataType =
   | UnionType;
 
 export type NumberLikeType =
-  | NumberType
+  | FloatType
   | NumberLiteral
-  | IntegerType
+  | Int32Type
```
lfportal, line 38: What of `Int64`?

lib/src/models/types/kinds.ts
```diff
@@ -25,9 +26,9 @@ export const StringLikeKind = [
 ];
 
 export const NumberLikeKind = [
-  TypeKind.NUMBER,
+  TypeKind.FLOAT,
   TypeKind.NUMBER_LITERAL,
-  TypeKind.INTEGER
+  TypeKind.INT32
```
lfportal: What of `Int64`?

lib/src/parsers/utilities/type-reference-resolver.ts
```diff
@@ -165,8 +165,8 @@ function retrieveTypeReferencesFromType(
       case TypeKind.NULL:
       case TypeKind.BOOLEAN:
       case TypeKind.STRING:
-      case TypeKind.NUMBER:
-      case TypeKind.INTEGER:
+      case TypeKind.FLOAT:
+      case TypeKind.INT32:
```
lfportal, line 169: `INT64`?

lfportal (approved):

lfportal (changes requested):
changes in previous review

lfportal:
https://github.com/airtasker/spot/wiki/Guide#4-types will require an update as well

fwouts:
Thanks @lfportal, all addressed.

lfportal (approved):

PR #226, opened 2019-05-14
Set required property only when needed

alephao:
Having a `"required": []` property is not a valid openapi3-schema, this pull request prevents spot from adding the `required` field if there are no required properties.

**To illustrate, if I had this on spot:**
```typescript
interface MyObject {
  my_property?: string;
}
```

**it would generate:**
```json
"MyObject": {
  "type": "object",
  "properties": {
    "my_property": {
      "type": "string"
    }
  },
  "required": [] // <--- wrong
}
```

**After fixing:**
```json
"MyObject": {
  "type": "object",
  "properties": {
    "my_property": {
      "type": "string"
    }
  }
}
```

### DIFF
--- a/lib/src/generators/contract/openapi3-schema.ts
+++ b/lib/src/generators/contract/openapi3-schema.ts
@@ -72,10 +72,11 @@ export function openApi3TypeSchema(
       };
     case TypeKind.OBJECT:
       return type.properties.reduce<
-        OpenAPI3SchemaTypeObject & { required: string[] }
+        OpenAPI3SchemaTypeObject
       >(
         (acc, property) => {
           if (!property.optional) {
+            if (acc.required === undefined) { acc.required = []; }
             acc.required.push(property.name);
           }
           acc.properties[property.name] = openApi3TypeSchema(
@@ -86,8 +87,7 @@ export function openApi3TypeSchema(
         },
         {
           type: "object",
-          properties: {},
-          required: []
+          properties: {}
         }
       );
     case TypeKind.ARRAY:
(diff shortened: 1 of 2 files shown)


lib/src/generators/contract/openapi3-schema.ts
```diff
@@ -72,10 +72,11 @@ export function openApi3TypeSchema(
       };
     case TypeKind.OBJECT:
       return type.properties.reduce<
-        OpenAPI3SchemaTypeObject & { required: string[] }
+        OpenAPI3SchemaTypeObject
       >(
         (acc, property) => {
           if (!property.optional) {
+            if (acc.required === undefined) { acc.required = []; }
```
fwouts: Run `yarn prettier:write` to fix the formatting :)

fwouts (commented):
Thanks for the PR!

https://json-schema.org/understanding-json-schema/reference/object.html#required-properties This specifically says "The required keyword takes an array of zero or more strings".

OpenAPI 3 is based on JSON Schema, although it does deviate on some topics. Can you point to the documentation that says this isn't allowed?

alephao:
@fwouts  Sorry, should have linked some stuff. OpenAPI is not using the latest Draft of json schema

[OpenAPI-Specification#data-types](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types)
> Primitive data types in the OAS are based on the types supported by the JSON Schema Specification Wright Draft 00. 

[draft-wright-json-schema-validation-00#section-5.15](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.15)
> The value of this keyword MUST be an array.  This array MUST have at
   least one element.  Elements of this array MUST be strings, and MUST
   be unique.

fwouts (approved):
OK, that makes sense! FYI @committedteadrinker @medric in case this affects the other codegen.

lfportal (approved):
Thanks @alaphao !

PR #228, opened 2019-05-14
Run tests in parallel on CircleCI by increasing resource size

fwouts:
It costs a bit more money, but it's also twice as fast. Given the very little time we spend running Spot builds, the improvement seems worth it.

I suspect we could remove the need for such a powerful machine by using the `ts` library directly instead of `ts-morph`, but that would likely be a big effort.


.circleci/config.yml
```diff
@@ -25,7 +26,7 @@ jobs:
 
       - run:
           name: Run tests
-          command: yarn test --runInBand
+          command: yarn test
```
lfportal, line 29: Do you remember the reason for `--runInBand`?

fwouts: It was running out of memory :)

lfportal (approved):

fwouts:
@dependabot squash and merge

PR #234, opened 2019-05-15
Install speccy and swagger2openapi in devDependencies

fwouts:
This means that:
- we will always know which version of speccy and swagger2openapi is used on CircleCI (as opposed to `yarn install` pulling in the latest)
-  they will get cached, speeding up builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,10 +37,6 @@ jobs:
       - run: ./bin/run --help
 
       - run:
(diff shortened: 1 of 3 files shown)


driesg (approved):
looks good to me.

It seems that Speccy is deprecated in favor of Spectral so we might want to have a look at that at some point

fwouts:
Sounds good. It's only used on CircleCI anyway :)

PR #235, opened 2019-05-15
Unicity error messages

lfportal:
Implemented more useful error logging for unicity verifiers.

### DIFF
--- a/lib/src/verifiers/unicity/endpoints.spec.ts
+++ b/lib/src/verifiers/unicity/endpoints.spec.ts
@@ -64,7 +64,7 @@ describe("unique endpoint names verifier", () => {
     };
     expect(verifyUniqueEndpointNames(contractNode)).toMatchObject([
       {
-        message: "endpoints must have unique names"
+        message: "endpoints must have unique names: EndpointOne"
       }
     ]);
   });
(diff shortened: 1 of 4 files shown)


lib/src/verifiers/unicity/endpoints.spec.ts
```diff
@@ -64,7 +64,7 @@ describe("unique endpoint names verifier", () => {
     };
     expect(verifyUniqueEndpointNames(contractNode)).toMatchObject([
       {
-        message: "endpoints must have unique names"
+        message: "endpoints must have unique names: EndpointOne"
```
fwouts, line 67: I think a better wording now would be `Found multiple endpoints with the name "EndpointOne"`. Thoughts? (same for types)

fwouts (approved):

lib/src/verifiers/unicity/endpoints.spec.ts
lfportal: Would be worth considering for a complete rethink of error messages, but currently error messages state the rule that the contract must follow. For now I think consistency is best

fwouts: SGTM!

PR #247, opened 2019-05-27
Helpful error message for missing String declaration

lfportal:
When using the `String` native type, the TypeScript compiler does not complain as there exists native `String` interfaces in TypeScript. However, Spot will complain with an unhelpful message. A more helpful error message has been added to help with understanding the error.

### DIFF
--- a/lib/src/parsers/utilities/parser-utility.ts
+++ b/lib/src/parsers/utilities/parser-utility.ts
@@ -449,7 +449,17 @@ export function getTargetDeclarationFromTypeReference(
     : symbol;
   const declarations = targetSymbol.getDeclarations();
   if (declarations.length !== 1) {
-    throw new Error("expected exactly one declaration");
+    const location = typeReference.getSourceFile().getFilePath();
+    const line = typeReference.getStartLineNumber();
+    // String interface must not be redefined and must be imported from the Spot native types
+    const errorMsg = `${location}#${line}: expected exactly one declaration for ${symbol.getName()}`;
+    if (symbol.getName() === "String") {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


fwouts (approved):
Have you considered the alternative of simply behaving as if the import was there?

fwouts (changes requested):
Oh also, a test would be good 🙂

lfportal:
> Have you considered the alternative of simply behaving as if the import was there?

Do you mean to accept all forms of `String`?
- typescript native
- imported from airtasker syntax
- user redefined

fwouts:
Yes, exactly. Thoughts?

lfportal:
> Yes, exactly. Thoughts?

From a purest POV, I find it odd that we let users redefine `String`. I think allowing the use of the typescript native `String` interface is fine, though it may mean more work in the parser somewhere - will need to try a few things to see what the effects are.

lfportal:
@fwouts There doesn't seem to be an elegant way to determine whether `String` is using the native TypeScript interface. We can only look at declaration file paths. There are 9 declarations of `String` in TypeScript - none of which are type-aliasing `string`. Even if we simply pick the first declaration to be used, it will error downstream when checking for whether the Spot internal `String` type has been redefined, which require creating another exception for. I think this may lead to more discrete behaviour required to handle `String` elsewhere in the code which will be harder to maintain as Spot evolves. I'm thinking that we should hold off on allowing `String` without imports until it's determined to be too much a nuisance to require importing. Thoughts?

fwouts:
You're right, in that case let's just clearly reject any missing String imports.

fwouts (approved):

PR #268, opened 2019-06-06
Test runner debug mode

lfportal:
The Spot test runner does not print out responses when validations succeed. There can be cases where it is useful to see what data is returned from the provider to assist debugging even if the data is compliant with the contract. For example an expected body "{}" definition is compliant with any response body returned. However the status code may be incorrect. The user has no simple way to see the actual response body to debug any issues. A debug mode will be added to enable this.

### DIFF
--- a/cli/src/test-utils/test-logger.ts
+++ b/cli/src/test-utils/test-logger.ts
@@ -1,38 +1,82 @@
 import chalk from "chalk";
 
-export const TestLogger = {
-  error,
-  log,
-  success,
-  warn
-};
+export class TestLogger {
+  private readonly debugMode: boolean;
+
+  constructor(opts?: LoggerOpts) {
+    this.debugMode = opts ? !!opts.debugMode : false;
+  }
+
+  // tslint:disable:no-console
+  debug(message: string, opts?: LogOpts): void {
+    if (this.debugMode) {
+      console.log(chalk.gray(this.transformMessage(message, opts)));
+    }
+  }
+
+  log(message: string, opts?: LogOpts): void {
+    console.log(chalk.dim.white(this.transformMessage(message, opts)));
+  }
+
+  success(message: string, opts?: LogOpts): void {
+    console.log(chalk.green(this.transformMessage(message, opts)));
+  }
+
+  warn(message: string, opts?: LogOpts): void {
+    console.log(chalk.yellow(this.transformMessage(message, opts)));
+  }
+
+  error(message: string, opts?: LogOpts): void {
+    console.log(chalk.red(this.transformMessage(message, opts)));
+  }
+  // tslint:enable:no-console
+
+  private transformMessage(message: string, customOpts?: LogOpts): string {
+    const opts = {
+      indent: customOpts ? customOpts.indent || 0 : 0
+    };
+    const indents = "\t".repeat(opts.indent);
+    return indents + message.replace(/\n/g, `\n${indents}`);
+  }
+}
+
+// export const TestLogger = {
+//   error,
+//   log,
+//   success,
+//   warn
+// };
 
 // tslint:disable:no-console
-function error(message: string, opts?: LoggerOpts) {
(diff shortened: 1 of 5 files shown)


cli/src/test-utils/test-logger.ts
```diff
@@ -1,38 +1,54 @@
...
-};
-
-// tslint:disable:no-console
-function error(message: string, opts?: LoggerOpts) {
-  console.log(chalk.red(transformMessage(message, opts)));
-}
+export class TestLogger {
+  /** Prepares an object for printing */
+  static formatObject(obj: any): string {
+    return JSON.stringify(obj, undefined, 2);
```
committedteadrinker, line 6: what's 2?

Kastet:
@lfportal do we need to add this to README? Also, what will the test report look like when status code didn't match but response body was compliant?

cli/src/test-utils/test-runner.ts
```diff
@@ -13,472 +13,472 @@ import {
...
+
+        if (result) {
+          this.logger.success(
+            `Test ${endpoint.name}:${
+              test.name
+            } passed (${TestTimer.formattedDiff(testStartTime)})`,
+            { indent: 1 }
+          );
+        } else {
+          this.logger.error(
```
committedteadrinker, line 73: This does not seem to log the error cause from `executeTest`? Is that the intention for now?

cli/src/test-utils/test-logger.ts
lfportal: the number of spaces to use for json indentation
```
2 space:
{
  "blah": 123
}

10 space:
{
          "blah": 123
}
```

committedteadrinker: TIL

cli/src/test-utils/test-runner.ts
lfportal: The reason for failure is reported during the test run, `result` is simply the boolean to let the test runner know that this test failed - to print out "this test failed" message

lfportal:
@Kastet 
Screen Shot 2019-06-06 at 11 22 33 am

cli/src/test-utils/test-runner.ts
```diff
@@ -13,472 +13,472 @@ import {
...
+   */
+  async test(
+    definition: ContractDefinition,
+    config?: TestConfig
+  ): Promise<boolean> {
+    const testSuiteStartTime = TestTimer.startTime();
+
+    let allPassed = true;
+
+    for (const endpoint of definition.endpoints) {
```
committedteadrinker, line 43: just reminding to revisit this for loooops later on

committedteadrinker (approved):

cli/src/test-utils/test-runner.ts
lfportal: 👍coverage reports and junit formatted test summary reports will require a bit of a refactor to how this is done. It will be a good time to look at it then.

lfportal:
> @lfportal do we need to add this to README?

documentation of CLI flags will auto generate on the next release

Kastet (approved):

cli/src/test-utils/test-runner.ts
```diff
@@ -13,472 +13,472 @@ import {
...
+        },
+        {}
+      )
+    };
+    const validateFn = jsv.compile(schema);
+    const valid = validateFn(response.data);
+    if (valid) {
+      this.logger.success("Body compliant", { indent: 2 });
+      return true;
+    } else {
```
medric: You can omit the `else`. You'd have already returned. 
```javascript
if (valid) {
  this.logger.success("Body compliant", { indent: 2 });
  return true;
}

this.logger.error(
  `Body is not compliant: ${jsv.errorsText(
      validateFn.errors
    )}\nReceived:\n${TestLogger.formatObject(response.data)}`,
    { indent: 2 }
  );
return false;
```

medric (approved):

PR #275, opened 2019-06-11
Add helpful error messages when using enum, indexed types/interfaces and maps

fwouts:
At the moment, we don't support the following types:
```typescript
// Enums are not supported at all.
enum SomeEnum {
  A,
  B,
  C
}

// Interfaces and types where keys are not fixed.

// 1) Interface syntax.
interface DictionaryInterface {
  [key: string]: string;
}

// 2) Type alias syntax.
type DictionaryType = {
  [key: string]: string;
}

// ES6 Maps.
type MapType = Map<string, string>;
```

Unfortunately, someone who uses Spot for the first time may assume that these may work. Therefore we introduce explicit errors so that it fails in a helpful way.

We may separately consider adding support for these syntaxes in the future.

### DIFF
--- a/lib/src/parsers/utilities/parser-utility.ts
+++ b/lib/src/parsers/utilities/parser-utility.ts
@@ -17,6 +17,7 @@ import {
 } from "ts-morph";
 import { HttpMethod } from "../../models/http";
 import { Locatable } from "../../models/locatable";
+import { inspect } from "util";
 
 /**
  * Extracts the JS Doc comment from a node.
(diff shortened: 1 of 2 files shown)


lib/src/parsers/utilities/parser-utility.ts
```diff
@@ -462,12 +469,43 @@ export function getTargetDeclarationFromTypeReference(
     }
   }
   const targetDeclaration = declarations[0];
+  if (
+    TypeGuards.isInterfaceDeclaration(targetDeclaration) &&
+    targetDeclaration.getIndexSignatures().length > 0
```
Kastet, line 475: What is an index signature? And why checking for them here?

lib/src/parsers/utilities/parser-utility.ts
```diff
@@ -462,12 +469,43 @@ export function getTargetDeclarationFromTypeReference(
...
   }
+  if (TypeGuards.isEnumDeclaration(targetDeclaration)) {
+    throw new Error(
+      `enums are not supported (offending type: ${targetDeclaration.getName()})`
+    );
+  }
+  if (TypeGuards.isEnumMember(targetDeclaration)) {
+    throw new Error(
+      `enums are not supported (offending type: ${targetDeclaration
+        .getParent()
```
Kastet, line 509: What's `parent` for and why only in enums?

lfportal (approved):

lib/src/parsers/utilities/parser-utility.ts
fwouts: Added clarifying comments :)

lib/src/parsers/utilities/parser-utility.ts
fwouts: The parent of an enum member is the enum type itself. For example when the type is `SomeEnum.A`, we want to show `"offending type: SomeEnum"`.

Kastet (approved):

PR #277, opened 2019-06-11
Add type safety with optional type parameters with @test<RequestBody, ResponseBody>

fwouts:


lib/src/test/examples/contract.ts
```diff
@@ -76,7 +76,7 @@ class CreateUser {
     @body body: ErrorBody
   ) {}
 
-  @test({
+  @test<CreateUserRequestBody, UserBody>({
```
Kastet, line 79: What effect do these parameter have? Request and response body will be checked against these types? Will `allowInvalidRequest` option still work in this case?

lfportal: They will enforce type-safety at compile time - and also provide type hints for the IDE. `allowInvalidRequest` will not work as expected when the Request parameter is used.

Kastet: Ok, then we also need to update the doc https://github.com/airtasker/spot/wiki/Contract-Testing

fwouts: Before we go and update the docs @Kastet does the approach look OK to you?

Kastet: yeah it's a nice improvement!

Kastet (approved):

lfportal (approved):

lib/src/test/examples/contract.ts
fwouts: PR is now merged. @Kastet what did you want to change in the wiki? I just noticed the examples don't actually have a request body at the moment.

PR #281, opened 2019-06-13
[TT-278] Generate spot contract hash 

medric:
**WHAT**

This PR introduces a way to generate spot contracts hashes. Generated hashes will be used to detect changes in such contracts and will be reflected in client SDKs version numbers i.e. `{major}.{minor}.{patch}-{hash}`.

**HOW**

The approach use to generate such hashes relies on object serialiazation and a call to a cryptographic hash function. The result is outputted to the standard console

Screen Shot 2019-06-13 at 1 14 32 pm


cli/src/commands/tag.ts
```diff
@@ -0,0 +1,44 @@
+import { Command, flags } from "@oclif/command";
+import { hashContractDefinition } from "../../../lib/src/tagging/hash";
+import { safeParse } from "../common/safe-parse";
+
+const ARG_API = "spot_contract";
+
+/**
+ * oclif command to generate a tag based on a Spot contract
+ */
+export default class Tag extends Command {
```
lfportal: Tag doesn't sound like it describes this feature. How about something similar to `Checksum`

lfportal (approved):

cli/src/commands/tag.ts
medric: No worries Leslie. I can find a more appropriate name. Will address.

medric: No worries Leslie. I can find a more appropriate name. Will address.

cli/src/commands/tag.ts
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  async run() {
+    const {
+      args,
+      flags: { algorithm }
+    } = this.parse(Tag);
+    try {
+      const contract = safeParse.call(this, args[ARG_API]).definition;
+      const hash = hashContractDefinition(contract, { algorithm });
```
alephao: Hello, I have a question

1. There are cases where the Spot contract will change, but the generated output like `openapi` won't change. In this case, Spot will generate a different hash but the output (`openapi`) will be the same
E.g.: 
```
// You have the interface Foo
interface Foo {
  a: String;
  b: String;
}

// But decide to move one of the properties to another interface
interface Bar {
  b: String;
}

interface Foo extends Bar {
  a: String;
}
```

2. There are cases where the Spot implementation changes, but my Spot contract have no changes. In this case, it will generate a new output, but the hash won't change. 
E.g.: from v0.2.8 to v0.2.9

Should we do the checksum against generated output instead of the spot contract?

README.md
```diff
@@ -263,5 +263,26 @@ EXAMPLE
...
+```
+USAGE
+  $ spot tag SPOT_CONTRACT
+
+ARGUMENTS
+  SPOT_CONTRACT  path to Spot contract
+
+OPTIONS
+  -h, --help             show CLI help
+  --algorithm=algorithm  digest algorithm used to generate the hash
```
alephao: What are the options?

fwouts: I'd prefer not to be able to specify an algorithm. It's important that two different people running `spot tag` on the same contract (with the same version of Spot) end up with the same result.

Allowing you to pick a specific algorithm introduces uncertainty: if you see a different result, is it because the other person had used a different algorithm?

Based on this, I think we should hardcode the choice of algorithm.

lib/src/tagging/hash.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { cleanse } from "../cleansers/cleanser";
+import { parse } from "../parsers/parser";
+import { hashContractDefinition } from "./hash";
+
+describe("Hash", () => {
+  describe("hashContractDefinition", () => {
+    it("returns a new hash when the contract is updated", () => {
```
fwouts: I think we need more extensive test coverage here.

For example, you could implement ` hashContractDefinition` as follows:
```
function hashContractDefinition(contract) {
  return hash(contract.api.name);
}
```

This test would pass, but the underlying hash wouldn't actually be good enough.

I'd recommend adding tests for:
- adding a new endpoint
- changing a property in request body
- changing a property in response body
- possibly also header and request parameters

fwouts (changes requested):
Nice :)

README.md
medric: hey  -  will remove this option. See comment below.

cli/src/commands/tag.ts
medric: hey @alephao. 

1. We're not aiming to do a checksum based on an openapi output here, but rather on an internal representation of the contract itself (in memory representation) as in contains data that would not be reflected in the openapi output - e.g. `tests` :

Screen Shot 2019-06-14 at 10 03 18 am

2. We acknowledge this but we don't expect the spot openapi codegen (templates) to change as it's pretty stable. The contract definition is our source of truth for now but we're keeping in mind that this will need to be improved. The point here is to ensure we generate ~accurate hashes for versionning our client sdks.

README.md
medric: yea i actually hesitated. Makes sense to hardcode this.

lib/src/tagging/hash.spec.ts
medric: Thank you for the feedback. Will review this.

README.md
fwouts: Reminder to remove it from the README now that you've removed the code. Feel free to resolve this comment then.

lib/src/tagging/hash.spec.ts
```diff
@@ -0,0 +1,102 @@
+import { cleanse } from "../cleansers/cleanser";
+import { EndpointDefinition } from "../models/definitions";
+import { TypeKind } from "../models/types/kinds"; //
+import { parse } from "../parsers/parser";
+import { hashContractDefinition } from "./hash";
+
+describe("Hash", () => {
+  describe("hashContractDefinition", () => {
+    it("returns a new hash when a new endpoint is added", () => {
```
fwouts, line 17: Should we also add a test that checks the hash is consistent? E.g.

```
it("returns a consistent hash", () => {
  const result = parse("./lib/src/test/examples/contract.ts");
  const contractDefinition = cleanse(result);
  const hash0 = hashContractDefinition(contractDefinition);
  const hash1 = hashContractDefinition(contractDefinition);
  expect(hash0).toEqual(hash1);
});
```

lib/src/tagging/hash.ts
```diff
@@ -0,0 +1,17 @@
+import { createHash } from "crypto";
+import { ContractDefinition } from "../models/definitions";
+
+export function hashContractDefinition(
+  contractDefinition: ContractDefinition
+): string {
+  const contractDefinitionString = JSON.stringify(contractDefinition).replace(
+    /\s/g,
```
fwouts, line 9: Why is this necessary?

medric: Don't know if this is really necessary, but read that `JSON.stringify` could add unwanted whitespace - so that would make it formatting independent.

README.md
```diff
@@ -263,5 +263,25 @@ EXAMPLE
   $ spot validate api.ts
 ```
 
+_See code: [build/cli/src/commands/validate.js](https://github.com/airtasker/spot/blob/v0.2.13/build/cli/src/commands/validate.js)_
+
+## `spot tag SPOT_CONTRACT`
```
fwouts, line 268: This still refers to `tag`, not checksum

cli/src/commands/checksum.ts
```diff
@@ -0,0 +1,38 @@
...
+import { hashContractDefinition } from "../../../lib/src/tagging/hash";
+import { safeParse } from "../common/safe-parse";
+
+const ARG_API = "spot_contract";
+
+/**
+ * oclif command to generate a tag based on a Spot contract
+ */
+export default class Checksum extends Command {
+  static description = "Generate a version tag based on a Spot contract";
```
fwouts, line 11:
```suggestion
  static description = "Generate a checksum based on a Spot contract";
```

cli/src/commands/checksum.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  static examples = ["$ spot mock api.ts"];
```
fwouts, line 13:
```suggestion
  static examples = ["$ spot checksum api.ts"];
```

lib/src/tagging/hash.ts
fwouts: I'd prefer to remove it.

fwouts (approved):

PR #283, opened 2019-06-14
Move testing logic under lib/src

fwouts:
Until now, all logic for contract testing lived under cli/. This directory was meant to only contain code directly related to the CLI (e.g. interacting with shell and printing to stdout).

This PR moves the code to lib/src/testing. In order to avoid confusion with the existing lib/src/test directory, which contains test helpers (for Jest), that directory was renamed to lib/src/spec-helpers. Additionally, directories containing examples were renamed to __examples__ and moved to a more meaningful spot.

This also adds `TestPrinter` to `TestRunnerConfig` so that the Spot test runner doesn't directly depend on `console.log`. This also reduces spammy output when running tests :)


lfportal (approved):
👍

PR #285, opened 2019-06-17
Remove usage of ramda

fwouts:
We have both lodash and ramda. They do the same job, but lodash is more widely used in the codebase. Therefore, this removes all usage of ramda.


lfportal (approved):
👍

PR #286, opened 2019-06-17
[TT-278] Spot contract hash

medric:
Doing some clean up 
- Update readme accordingly
- Don't remove whitespaces in JSON serialization


fwouts (changes requested):
Thanks! I also noticed that the directory is currently called `tagging`, should it be renamed to `checksum`?

fwouts (approved):

PR #288, opened 2019-06-17
Add a lint rule to enforce union types have a discriminator

fwouts:


lib/src/linting/rules/has-discriminator.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Checks that all union types have a discriminator.
+ */
+export const hasDiscriminator: LintingRule = contract => {
+  const topLevelTypes = flatten([
+    ...contract.types,
+    ...contract.endpoints.map(extractEndpointTypes)
+  ]);
+  const unionTypes = flatten(
```
committedteadrinker, line 17: is the data that is `flatten`-ed guaranteed to be only 1 depth? Or do we want to only flatten 1 depth?

fwouts: https://lodash.com/docs/4.17.11#flatten yep

lib/src/linting/rule.ts
```diff
@@ -9,5 +9,5 @@ export type LintingRule = (contract: ContractNode) => LintingRuleViolation[];
 
 export interface LintingRuleViolation {
   message: string;
-  source: Locatable<unknown>;
+  source?: Locatable<unknown>;
```
committedteadrinker, line 12: What is `Locatable`? I think I asked this before and all I got was it is a table to enable access to things from different places.

committedteadrinker (approved):
Just a couple of questions

lib/src/linting/rule.ts
fwouts: It's a wrapper around a value that also contains a location (source file and line) so that we can show helpful error messages including the location where the bad code was found. Does that make sense?

PR #293, opened 2019-06-19
 Extract helper methods out and remove NamedType

fwouts:
This will be used for other linting rules later on.

### DIFF
--- a/lib/src/models/nodes.ts
+++ b/lib/src/models/nodes.ts
@@ -8,10 +8,10 @@ export interface ContractNode {
   types: TypeNode[];
 }
 
-export interface TypeNode {
+export interface TypeNode<T extends DataType = DataType> {
   name: string;
   description?: string;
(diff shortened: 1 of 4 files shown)


lib/src/models/nodes.ts
```diff
@@ -8,10 +8,10 @@ export interface ContractNode {
   types: TypeNode[];
 }
 
-export interface TypeNode {
+export interface TypeNode<T extends DataType = DataType> {
```
lfportal, line 11: what does this do?

fwouts: It means that `TypeNode` can be restricted to a specific subtype:
- `TypeNode` is the same as `TypeNode<DataType>`
- `TypeNode<UnionType>` is specifically containing a `UnionType`

lfportal (approved):

PR #294, opened 2019-06-19
Simplify linting commands

fwouts:
It's much easier to run `yarn lint:fix` rather than remembering two different comments for Prettier and TSLint.


medric (approved):
👍

PR #296, opened 2019-06-19
Do not fail the has-discriminator rule for nullable types

fwouts:
The following types should not trigger the `has-discriminator` lint rule:

```
type A = B | null;
```

```
type A = string | null
```

### DIFF
--- a/lib/src/linting/rules/has-discriminator.spec.ts
+++ b/lib/src/linting/rules/has-discriminator.spec.ts
@@ -12,7 +12,8 @@ import {
   objectType,
   STRING,
   stringLiteral,
-  unionType
+  unionType,
+  NULL
 } from "../../models/types";
 import { fakeLocatable } from "../../spec-helpers/fake-locatable";
 import { hasDiscriminator } from "./has-discriminator";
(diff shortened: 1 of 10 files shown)


lib/src/linting/rules/has-discriminator.spec.ts
```diff
@@ -79,6 +80,63 @@ describe("rule: has-discriminator", () => {
...
+          name: "NullableObject",
+          type: unionType([
+            objectType([
+              {
+                name: "name",
+                optional: false,
+                type: STRING
+              }
+            ]),
+            NULL
```
committedteadrinker, line 116: ah interesting, does this mean that the generator will only generate 1 one-of child?

committedteadrinker (approved):
Looks straightforward, just a question.

lib/src/linting/rules/has-discriminator.spec.ts
fwouts: No, the generator will directly generate the type with the `nullable` property set to true. So it's not generating a one-of at all :)

PR #299, opened 2019-06-21
Fix a bug where `string | null` triggered a false positive in the no-nested-types-within-unions linting rule

fwouts:


lib/src/linting/rules/no-nested-types-within-unions.ts
```diff
@@ -16,6 +17,7 @@ export const noNestedTypesWithinUnions: LintingRule = contract => {
     topLevelTypes.map(t => extractNestedUnionTypes(t.type, t.name))
   );
   return unionTypes
+    .filter(negate(isUnionOfSingleTypeWithNull))
```
committedteadrinker, line 20: Does this mean that we want to 'catch' types that do not have a union of single | null? Or am I understanding this the other way around

fwouts: This is saying "exclude types that are of the form SingleType | null` :)

committedteadrinker (approved):

lib/src/linting/rules/no-nested-types-within-unions.ts
committedteadrinker: oooooooohh, yeah the other way around.

PR #306, opened 2019-06-26
Draft decorator

Kastet:
This adds `@draft` decorator to endpoint class. `@draft` decorator will later be used by generators to augment endpoint path to indicate that the endpoint is under construction.


lib/src/parsers/nodes/endpoint-parser.spec.ts
```diff
@@ -59,6 +59,7 @@ describe("@endpoint parser", () => {
 
     expect(parseEndpoint(klass)).toStrictEqual({
       value: {
+        isDraft: false,
```
lfportal, line 59: I think we should add a `isDraft: true` test to ensure it works correctly

lfportal (approved):

PR #314, opened 2019-07-03
Negative expression parsing

lfportal:
The Spot parser fails to parse negative numbers in `@test` methods.

### DIFF
--- a/lib/src/parsers/utilities/expression-parser.ts
+++ b/lib/src/parsers/utilities/expression-parser.ts
@@ -23,17 +23,17 @@ export function parseExpression(expression: Expression): DataExpression {
   } else if (TypeGuards.isStringLiteral(expression)) {
     return stringExpression(expression.getLiteralValue());
   } else if (TypeGuards.isPrefixUnaryExpression(expression)) {
-    if (expression.getOperatorToken() === SyntaxKind.MinusToken) {
-      const operand = expression.getOperand();
-      if (TypeGuards.isNumericLiteral(operand)) {
+    const operand = expression.getOperand();
+    if (!TypeGuards.isNumericLiteral(operand)) {
+      throw new Error("unary operators may only be used with numeric literals");
+    }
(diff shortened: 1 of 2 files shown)


committedteadrinker (approved):

lib/src/parsers/utilities/expression-parser.ts
```diff
@@ -22,6 +22,19 @@ export function parseExpression(expression: Expression): DataExpression {
     return booleanExpression(expression.getLiteralValue());
   } else if (TypeGuards.isStringLiteral(expression)) {
     return stringExpression(expression.getLiteralValue());
+  } else if (TypeGuards.isPrefixUnaryExpression(expression)) {
+    if (expression.getOperatorToken() === SyntaxKind.MinusToken) {
```
fwouts: I'd suggest a `switch` statement instead, and handling the plus token as well:

```
const operand = expression.getOperand();
if (!TypeGuards.isNumericLiteral(operand)) {
  throw new Error("unary operators may only be used with numeric literals");
}
switch (expression.getOperatorToken()) {
  case SyntaxKind.MinusToken:
    return numberExpression(-operand.getLiteralValue());
  case SyntaxKind.PlusToken:
    return numberExpression(+operand.getLiteralValue());
  default:
    throw new Error("unknown prefix operator token");
}
```

fwouts (approved):

lemattma (approved):

lib/src/parsers/utilities/expression-parser.ts
lfportal: is `age: +30` syntax we want to support?

fwouts: I think so, at least I see no reason to support -30 but not support +30. It can be useful for example when you have two tests with opposite signs, and you want to make it clear that the difference is the sign. See what I mean?

PR #316, opened 2019-07-03
Include draft flag

Kastet:
Adds `--includeDraft` flag to the test runner CLI that allows to include draft endpoint tests during development. Documentation not included.

### DIFF
--- a/lib/src/testing/test-runner.spec.ts
+++ b/lib/src/testing/test-runner.spec.ts
@@ -147,7 +147,7 @@ describe("test runner", () => {
     expect(result).toBe(true);
   });
 
-  test("draft endpoint excluded", async () => {
+  test("non draft endpoints are still tested when draft not included", async () => {
     const contract = parseAndCleanse(
       `${testExamplesBasePath}/draft-and-nondraft-endpoint.ts`
     );
@@ -176,7 +176,7 @@ describe("test runner", () => {
     expect(result).toBe(true);
   });
 
-  test("draft endpoint included", async () => {
+  test("passing draft endpoint passes the test when included", async () => {
     const contract = parseAndCleanse(
       `${testExamplesBasePath}/draft-endpoint.ts`
     );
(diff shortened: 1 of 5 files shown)


lib/src/testing/test-runner.spec.ts
```diff
@@ -176,6 +176,26 @@ describe("test runner", () => {
...
+      .post("/companies", { name: "My Company", private: true })
+      .reply(201, { name: "My Company" })
+      .post("/state/teardown")
+      .reply(200);
+
+    const result = await testRunner.test(contract, {
+      includeDraft: true
+    });
+    expect(scopeDraft.isDone()).toBe(true);
+    expect(result).toBe(true);
```
fwouts, line 196: Would it be worth having another test where it fails, and confirming that it's true when includeDraft is false, and false when includeDraft is true?

Kastet: The check here is line 195 - `expect(scopeDraft.isDone()).toBe(true);`.
There is an [existing test](https://github.com/airtasker/spot/blob/test-runner-include-draft-flag/lib/src/testing/test-runner.spec.ts#L150) that checks that without `includeDraft` draft tests are excluded `expect(scopeDraft.isDone()).toBe(false);`

Kastet: Added

lib/src/testing/test-runner.spec.ts
```diff
@@ -176,7 +176,7 @@ describe("test runner", () => {
     expect(result).toBe(true);
   });
 
-  test("include draft", async () => {
+  test("draft endpoint included", async () => {
     const contract = parseAndCleanse(
       `${testExamplesBasePath}/draft-endpoint.ts`
```
fwouts, line 181: Is there a reason why the contract used isn't the same as for `draft endpoint excluded`?

lib/src/testing/test-runner.spec.ts
```diff
@@ -196,6 +196,46 @@ describe("test runner", () => {
     expect(result).toBe(true);
   });
 
+  test("draft endpoint included - response body mismatch", async () => {
```
fwouts: To be clear, what I'd like the test to be responsible for is `failing draft endpoint fails the test when draft endpoints are included`

(I'm not entirely sure the test below is necessary)

fwouts (approved):

PR #328, opened 2019-07-11
double support

lfportal:
Double data type support. A Spot native Number type has also been added to bring it in line with how Integer data types are supported.


committedteadrinker:
what happens when it's just `number` no format? Does it default to `float`?

committedteadrinker (approved):

lfportal:
> what happens when it's just `number` no format? Does it default to `float`?

`number` turns into `float`

PR #335, opened 2019-07-15
[TT-347] Proxy non-draft endpoints to get real response

medric:
This introduces a new module used within the main server middleware to proxy requests that expect real response as opposed to mock ones. The proxy will be leveraged when an endpoint is not in a draft state. 

To specify the url of the server the proxy needs to hit, users can set a `--proxyBaseUrl` flag when starting the mock server.


lib/src/mockserver/proxy.ts
```diff
@@ -0,0 +1,36 @@
...
+  response,
+  pathPrefix,
+  proxyBaseUrl
+}: {
+  incomingRequest: Request;
+  response: Response;
+  pathPrefix: string;
+  proxyBaseUrl: string;
+}) {
+  const [protocol] = proxyBaseUrl.split("://");
```
fwouts: Make sure to fail with a helpful message if we forgot to specify the protocol? Or fallback to http:// maybe?

lib/src/mockserver/server.ts
```diff
@@ -24,6 +27,20 @@ export function runMockServer(
   app.use((req, resp) => {
     for (const endpoint of api.endpoints) {
       if (isRequestForEndpoint(req, pathPrefix, endpoint)) {
+        const shouldProxy = !endpoint.isDraft;
+
+        // non-draft end points get real response
+        if (shouldProxy && proxyBaseUrl) {
+          proxyRequest({
```
fwouts: For conciseness, consider:
```suggestion
          return proxyRequest({
```

cli/src/commands/mock.ts
```diff
@@ -23,6 +23,10 @@ export default class Mock extends Command {
 
   static flags = {
     help: flags.help({ char: "h" }),
+    proxyBaseUrl: flags.string({
+      description:
+        "If set, the server would act as a proxy and fetch data from the given remote server instead of mocking it"
```
fwouts:
```suggestion
        "If set, the server will act as a proxy and fetch data from the given remote server for all non-draft endpoints instead of mocking them"
```

README.md
```diff
@@ -230,9 +230,11 @@ ARGUMENTS
   SPOT_CONTRACT  path to Spot contract
 
 OPTIONS
-  -h, --help               show CLI help
-  -p, --port=port          (required) [default: 3010] Port on which to run the mock server
-  --pathPrefix=pathPrefix  Prefix to prepend to each endpoint path
+  -h, --help                   show CLI help
+  -p, --port=port              (required) [default: 3010] Port on which to run the mock server
+  --pathPrefix=pathPrefix      Prefix to prepend to each endpoint path
+  --proxyBaseUrl=proxyBaseUrl  If set, the server would act as a proxy and fetch data from the given remote server
```
fwouts, line 258: I wonder, should we only update the README.md once we release a new version?

fwouts (changes requested):
Nice one. Unfortunately it looks like the mock server doesn't have test coverage yet (my bad). Would you mind adding some? We need to ensure that it correctly mocks draft endpoints and proxies the others unless there is no proxy URL.

lfportal (approved):

cli/src/common/infer-protocol.ts
```diff
@@ -0,0 +1,11 @@
+export default function inferProtocol(proxyBaseUrl: string): "http" | "https" {
+  const [protocol] = proxyBaseUrl && proxyBaseUrl.split("://");
+
+  if (protocol !== "http" && protocol !== "https") {
+    throw new Error(
+      'Err - could not infer protocol from proxy base url, should be either "http" or "https".'
```
fwouts: I don't think `Err - ` is necessary since it should be fairly clear in the stack trace that it's an error already?

```suggestion
      'Could not infer protocol from proxy base url, should be either "http" or "https".'
```

lib/src/mockserver/server.spec.ts
```diff
@@ -0,0 +1,122 @@
+import nock from "nock";
+import request from "supertest";
+import { ContractDefinition } from "../models/definitions";
+import { TypeKind } from "../models/types";
+import { runMockServer } from "./server";
+
+describe("Server", () => {
+  const proxyBaseUrl = "http://localhost:9988";
```
fwouts, line 8: Could you add another test where we don't set a proxy URL? Everything should be mocked in that case.

fwouts (changes requested):

cli/src/common/infer-proxy-config.ts
```diff
@@ -0,0 +1,20 @@
+import { ProxyConfig } from '../../../lib/src/mockserver/server';
+
+export default function inferProxyConfig(proxyBaseUrl: string): ProxyConfig | void {
```
fwouts:
```suggestion
export default function inferProxyConfig(proxyBaseUrl: string): ProxyConfig | null {
```

cli/src/common/infer-proxy-config.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (!proxyBaseUrl) {
+    return undefined;
```
fwouts: I think `null` better represents the concept of "absence of a value" rather than `undefined`. Thoughts?

```suggestion
    return null;
```

cli/src/common/infer-proxy-config.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  const [protocol] = proxyBaseUrl && proxyBaseUrl.split("://");
+
+  if (protocol !== "http" && protocol !== "https") {
+    throw new Error(
+      'Err - could not infer protocol from proxy base url, should be either "http" or "https".'
```
fwouts: Please address https://github.com/airtasker/spot/pull/335#discussion_r305178929

fwouts (approved):

PR #338, opened 2019-07-16
Experimental docs preview command

lfportal:
An experimental `docs` command to preview Spot contracts. This implementation uses [Redoc](https://github.com/Redocly/redoc) to preview an OpenAPI3 contract generated from a Spot contract. Spot completely supports the native OpenAPI3 spec, so is suitable to be used at this point in time.


cli/src/commands/docs.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  async run() {
+    const { args, flags } = this.parse(Docs);
+    const { port } = flags;
+
+    const server = express();
+    const docsDir = path.join(__dirname, "docs", "public");
+    server.use(express.static(docsDir));
+
+    server.get("/contract", (req, res) =>
```
fwouts: I think it'd be good to have a comment that explains how this endpoint is being used.

cli/src/commands/docs.ts
```diff
@@ -0,0 +1,62 @@
...
+      )
+    );
+
+    const start = async () => {
+      try {
+        this.log(`Documentation server started on port ${port}`);
+        this.log(`Open http://localhost:${port} to view documentation`);
+        await server.listen(port);
+      } catch (err) {
+        process.exit(1);
```
fwouts: I think there's a command `this.fatal()` or something of the like where you print the error and oclif exits. I think it would also give us enough logging to debug issues.

cli/src/commands/docs.ts
fwouts: Should this endpoint be renamed to `/contract-openapi3` since it's specifically that format?

fwouts (commented):
Neat!

Should we test this somehow?

README.md
```diff
@@ -139,6 +140,27 @@ EXAMPLE
 
 _See code: [build/cli/src/commands/checksum.js](https://github.com/airtasker/spot/blob/v0.2.18/build/cli/src/commands/checksum.js)_
 
+## `spot docs SPOT_CONTRACT`
+
+Preview Spot contract documentation
```
danieltheodosius: can this document be extended? I feel like it needs to explain which url to visit to access the docs.

cli/src/commands/docs.ts
```diff
@@ -0,0 +1,62 @@
...
+            "json"
+          )
+        )
+      )
+    );
+
+    const start = async () => {
+      try {
+        this.log(`Documentation server started on port ${port}`);
+        this.log(`Open http://localhost:${port} to view documentation`);
```
danieltheodosius, line 60: missing `/contract`

cli/src/commands/docs.ts
```diff
@@ -0,0 +1,62 @@
...
+          generateOpenApiV3(
+            safeParse.call(this, args[ARG_API]).definition,
+            "json"
+          )
+        )
+      )
+    );
+
+    const start = async () => {
+      try {
```
danieltheodosius, line 58: do you need to catch exception? or will `express` handle it if it cannot start?

cli/src/commands/docs.ts
danieltheodosius: in the case of exception, you might want to log the error to console

README.md
lfportal: The command will output:
```
Documentation server started on port 8080
Open http://localhost:8080 to view documentation
```
Is this sufficient? or do you think the Readme also needs more information?

cli/src/commands/docs.ts
lfportal: The `/contract` endpoint is used by the the React app. `/` just loads the `index.html` containing the React app.

cli/src/commands/docs.ts
lfportal: I think if express can't start it'll just throw an error like any other exception, I've changed the handling to ` this.error(err, { exit: 1 });` as per @fwouts suggestion so we should see the error in the console.

cli/src/commands/docs.ts
lfportal: as above

README.md
danieltheodosius: is it this or `http://localhost:8080/contract`?

danieltheodosius: Would be good to put it into readme

danieltheodosius (approved):

fwouts (approved):

PR #351, opened 2019-07-22
TT-348 - Update to strip draft string from endpoints.

committedteadrinker:
## WHY
Need to strip out the string `/_draft` from endpoints


lib/src/mockserver/server.ts
```diff
@@ -41,6 +41,8 @@ export function runMockServer(
             response: resp,
             ...proxyConfig
           });
+        } else if (!shouldProxy) {
```
medric: Don't need `else if`, line above would return and end the current call.

lib/src/mockserver/server.ts
```diff
@@ -66,6 +72,10 @@ export function runMockServer(
   };
 }
 
+function stripDraft(endpointName: string): string {
```
fwouts: I don't think `endpointName` is the correct name for this argument given its usage?

lib/src/mockserver/server.ts
```diff
@@ -66,6 +72,10 @@ export function runMockServer(
...
+  return endpointName.replace("/_draft", "");
```
fwouts: Nit to remove potential unwanted changes such as `/_draft_cats` being affected:
```suggestion
  return endpointName.replace("/_draft/", "/");
```

lib/src/mockserver/server.ts
committedteadrinker: Good point. Maybe `url` or `path` is more appropriate.

lib/src/mockserver/server.ts
committedteadrinker: good point. Thanks.

lib/src/mockserver/server.spec.ts
```diff
@@ -147,5 +148,25 @@ describe("Server", () => {
...
+          protocol,
+          proxyBaseUrl
+        }
+      });
+
+      request(app)
+        .post("/api/_draft/companies")
+        .expect(201)
+        .then(response => {
+          expect(response.body.name).not.toBe("This is the real response");
```
fwouts, line 169: maybe also check that it is indeed a string?

fwouts (approved):

lib/src/mockserver/server.ts
```diff
@@ -66,6 +72,10 @@ export function runMockServer(
   };
 }
 
+function stripDraft(url: string): string {
+  return url.replace("/_draft/", "/");
```
fwouts: Now that I look at this, it might be easier to read and less likely to get out of sync with the `if()` that comes above if it were inlined?

committedteadrinker: that's a good point, was taking this out in a separate method for the sake of 'only doing 1 thing'. But I guess it doesn't need it for now as it doesn't have any reusing, I'd rather take out the word `/_draft/` to a separate constant that can be reused anywhere as that's what's happening right now. There are `/_draft` used on the openapi3.ts and here.

lib/src/mockserver/server.spec.ts
committedteadrinker: good point, it's returning undefined, will need to figure out why it's not returning anything.

lib/src/mockserver/server.ts
```diff
@@ -29,6 +29,12 @@ export function runMockServer(
 ) {
   const app = express();
   app.use(cors());
+  app.use((req, resp, next) => {
+    if (req.path.includes("/_draft/")) {
+      req.url = stripDraft(req.url);
```
medric: As discussed offline - I was wondering what was the point of just stripping the `_draft` prefix without doing anything particular with it. @fwouts should we also add a context to the request, something along the lines of `req.isDraft = true`, and decide whether to return mock data based on this ?

committedteadrinker: a good discussion to start with @fwouts

committedteadrinker: it's mostly because the request coming in the server does not have a concept of `isDraft` we can only know that a request is a draft from it and without stripping that out, we cannot get the endpoint that we want to correctly mock.

committedteadrinker: just to clarify, the `/_draft/` prefix is also for developers to know when they're debugging which endpoints are returning mock responses.

PR #357, opened 2019-07-24
[TT-40] Expand output path with tilde

medric:
Node doesn't expand `~`to the current user's home directory - see https://github.com/nodejs/node/issues/684. 

Looks like this has to be implemented manually if we really want to leverage this feature, hence the following changes.


lfportal (approved):

README.md
```diff
@@ -255,7 +255,8 @@ OPTIONS
   -h, --help                   show CLI help
   -p, --port=port              (required) [default: 3010] Port on which to run the mock server
   --pathPrefix=pathPrefix      Prefix to prepend to each endpoint path
-  --proxyBaseUrl=proxyBaseUrl  If set, the server would act as a proxy and fetch data from the given remote server
+
+  --proxyBaseUrl=proxyBaseUrl  If set, the server will act as a proxy and fetch data from the given remote server
```
fwouts, line 259: This seems unrelated?

lib/src/utilities/expand-path-with-tilde.ts
```diff
@@ -0,0 +1,11 @@
+import os from "os";
+
+export default function expandPathWithTilde(path: string) {
```
fwouts: Consider using a named export to follow conventions in the Spot codebase.

fwouts (approved):

PR #360, opened 2019-07-30
[TT-415] Add linting rule for optional response fields 

medric:
This PR introduces a new linting rule to ensure responses don't define optional fields within any object it may encapsulate. Nullable should be used instead. 

I've also done a bit of refactor to leverage composition an re-use small individual functions, as in extracting response types.


lib/src/linting/rules/no-optional-fields-within-responses.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+                    ]),
+                    optional: false
+                  }
+                ])
+              })
+            })
+          ]
+        })
+      ],
+      types: []
```
fwouts: I'd like to see some tests where a nested response type is a reference type as well (both valid and invalid case), to ensure that reference types are correctly handled.

lib/src/linting/rules/no-optional-fields-within-responses.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  INT32,
+  NULL,
+  objectType,
+  STRING,
+  unionType
+} from "../../models/types";
+import { fakeLocatable } from "../../spec-helpers/fake-locatable";
+import { noOptionalFieldsWithinResponses } from "./no-optional-fields-within-responses";
+
+describe("rule: no optional fields within responses", () => {
```
fwouts: Both of the following can be interpreted as "optional":
```
{
  a?: string;
}

{
  a: string | null;
}
```

I prefer the term "omittable", which more clearly points to the former.

lib/src/linting/rules/no-optional-fields-within-responses.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+        })
+      ],
+      types: []
+    });
+    expect(errors).toEqual([
+      {
+        message:
+          "The object type `listUsers (response body for status 201)` defines an optional property. Use nullable instead."
+      },
+      {
```
fwouts: Why is it duplicated?

lib/src/utilities/extract-endpoint-types.ts
```diff
@@ -2,15 +2,10 @@ import { compact } from "lodash";
 import { Locatable } from "../models/locatable";
 import { EndpointNode, TypeNode } from "../models/nodes";
 
-export function extractEndpointTypes(
+export function extractResponseTypes(
```
fwouts, line 5: It looks like both `extractRequestTypes` and `extractResponseTypes` return an array with at most one element. How about changing their return type to `TypeNode | null` instead?

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,127 @@
+import assertNever from "assert-never";
+import { flatten } from "lodash";
+import { TypeNode } from "../models/nodes";
+import { ObjectType, TypeKind, UnionType } from "../models/types";
+import { resolveType } from "../verifiers/utilities/type-resolver";
+
+export function extractNestedUnionTypes({
```
fwouts: It would be worth writing some JSDoc that explains that nested reference types are not included. This is a change of behaviour compared to the previous implementation.

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,127 @@
...
+    };
+  } catch (e) {
+    // Bubble up exception
+    throw e;
+  }
+}
+
+export function extractNestedObjectTypes(
+  typeNode: TypeNode,
+  typeStore: TypeNode[] = []
```
fwouts: I'd prefer to not use a default value here. It's very easy to omit the second argument by mistake.

fwouts (changes requested):
Nice work.

lib/src/utilities/extract-endpoint-types.ts
medric: `extractResponseTypes` can return an array with more than 1 element -> `endpoint.value.responses.map` and default response.

lib/src/linting/rules/no-omittable-fields-within-responses.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+              optional: false
+            }
+          ])
+        }
+      ]
+    });
+    expect(errors).toEqual([]);
+  });
+
+  test("rejects a response object when a field is optional instead of nullable", () => {
```
fwouts: This still says "optional" instead of "omittable" :)

(other tests too)

lib/src/linting/rules/no-omittable-fields-within-responses.ts
```diff
@@ -0,0 +1,32 @@
...
+    extractObjectTypes,
+    flatten
+  );
+
+  const types = flatten(contract.endpoints.map(extractTypes)) as Array<
+    TypeNode<ObjectType>
+  >;
+
+  return types.filter(hasOptionalProperties).map(typeNode => ({
+    message: `The object type \`${typeNode.name}\` defines an optional property. Use nullable instead.`
```
fwouts:
```suggestion
    message: `The object type \`${typeNode.name}\` defines an omittable property. Use nullable instead.`
```

lib/src/linting/rules/no-omittable-fields-within-responses.ts
```diff
@@ -0,0 +1,32 @@
...
+  const extractObjectTypes = (t: TypeNode[]) =>
+    t.map((type: TypeNode) => extractNestedObjectTypes(type, contract.types));
+
+  const extractTypes = flow(
+    extractResponseTypes,
+    extractObjectTypes,
+    flatten
+  );
+
+  const types = flatten(contract.endpoints.map(extractTypes)) as Array<
```
fwouts: Why is a cast needed?

lib/src/utilities/extract-endpoint-types.ts
fwouts: Good point. What about `extractRequestTypes`?

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,139 @@
+import assertNever from "assert-never";
+import { flatten } from "lodash";
+import { TypeNode } from "../models/nodes";
+import { ObjectType, TypeKind, UnionType } from "../models/types";
+import { resolveType } from "../verifiers/utilities/type-resolver";
+
+/**
+ * Recursively traverse the Types tree, find and group union types definitions into an flat array
```
fwouts:
```suggestion
 * Recursively traverse the Types tree, find and group union types definitions into a flat array.
 *
 * Note: Type references are not traversed.
```

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  try {
+    const unreferencedType = resolveType(typeNode.type, typeStore);
+
+    return {
+      name: typeNode.name,
+      type: unreferencedType
+    };
+  } catch (e) {
+    // Bubble up exception
```
fwouts: As far as I know:
```
try {
  [code]
} catch (e) {
  throw e;
}
```

is exactly equivalent to:
```
[code]
```

What's the reasoning for this?

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,139 @@
...
+    throw e;
+  }
+}
+
+/**
+ * Recursively traverse the Types tree, find, resolve references, and group objects types definitions into an flat array
```
fwouts:
```suggestion
 * Recursively traverse the Types tree, find, resolve references, and group objects types definitions into a flat array.
```

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+export function extractNestedObjectTypes(
```
fwouts, line 71: This has a different behaviour from `extractNestedUnionTypes`: this one follows type refs, whereas the unions version doesn't. I think this is likely to trip up someone, no matter how well documented it is. Is there a good reason for the inconsistency?

fwouts (changes requested):

lib/src/utilities/extract-nested-types.ts
medric: Trying to find a way to avoid repeating logic, don't really now yet.

package.json
```diff
@@ -24,6 +24,7 @@
     "fs-extra": "^8.1.0",
     "inquirer": "^6.5.0",
     "js-yaml": "^3.13.1",
+    "jsdoc": "3.6.2",
```
fwouts: Why was this added?

lib/src/utilities/extract-nested-types.ts
fwouts: Let's chat in person.

lib/src/utilities/extract-endpoint-types.ts
```diff
@@ -25,3 +20,25 @@ export function extractEndpointTypes(
     )
   ]);
 }
+
+export function extractRequestTypes(
```
fwouts:
```suggestion
export function extractRequestType(
```

lib/src/utilities/extract-endpoint-types.ts
```diff
@@ -25,3 +20,25 @@ export function extractEndpointTypes(
...
+    null
+  );
+}
+
+export function extractEndpointTypes(
+  endpoint: Locatable<EndpointNode>
+): TypeNode[] {
+  return compact([
+    ...extractResponseTypes(endpoint),
+    extractRequestTypes(endpoint)
```
fwouts:
```suggestion
    extractRequestType(endpoint)
```

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,143 @@
...
+import { TypeNode } from "../models/nodes";
+import { ObjectType, TypeKind, UnionType } from "../models/types";
+import { resolveType } from "../verifiers/utilities/type-resolver";
+
+export function maybeResolveRef(
+  typeNode: TypeNode,
+  typeStore: TypeNode[]
+): TypeNode {
+  if (!typeStore.length) {
+    return typeNode;
```
fwouts: Shouldn't we crash when this occurs?

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,143 @@
...
+    case TypeKind.STRING:
+    case TypeKind.FLOAT:
+    case TypeKind.DOUBLE:
+    case TypeKind.INT32:
+    case TypeKind.INT64:
+    case TypeKind.DATE:
+    case TypeKind.DATE_TIME:
+    case TypeKind.BOOLEAN_LITERAL:
+    case TypeKind.STRING_LITERAL:
+    case TypeKind.NUMBER_LITERAL:
```
fwouts, line 26:
```suggestion
    case TypeKind.NUMBER_LITERAL:
      return [];
```

fwouts: (otherwise it will fall through!)

lib/src/utilities/extract-nested-types.ts
```diff
@@ -0,0 +1,143 @@
+import assertNever from "assert-never";
+import { flatten } from "lodash";
+import { TypeNode } from "../models/nodes";
+import { ObjectType, TypeKind, UnionType } from "../models/types";
+import { resolveType } from "../verifiers/utilities/type-resolver";
+
+export function maybeResolveRef(
```
fwouts: Could we delete this function and use https://github.com/airtasker/spot/blob/master/lib/src/verifiers/utilities/type-resolver.ts#L41 instead?

package.json
fwouts: Please address this comment too.

fwouts (changes requested):

fwouts (changes requested):
Please also add your linting rule to `rules.ts` :)

lib/src/utilities/extract-nested-types.ts
medric: why ?

fwouts (approved):

lfportal (approved):

PR #373, opened 2019-08-05
Add a linting rule to forbid nullable arrays

fwouts:


lib/src/linting/rules/no-nullable-arrays.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Checks that the contract doesn't define nullable arrays.
+ */
+export const noNullableArrays: LintingRule = contract => {
+  const topLevelTypes = flatten([
+    ...contract.types,
+    ...contract.endpoints.map(extractEndpointTypes)
+  ]);
+  const unionTypes = flatten(
+    topLevelTypes.map(t => extractNestedUnionTypes(t, contract.types))
```
medric, line 17: Why does this apply to union types only ?

fwouts: In the Spot world, just like in TypeScript, a nullable type is represented by a union of a type with `null`.

For example:
```typescript
type A = string | null;
```

This results in the type:
```
{
  kind: "union",
  types: [
    { kind: "string" },
    { kind: "null" }
  ]
}
```

medric (approved):

PR #374, opened 2019-08-05
[TT-416] Add a new linting rule to prevent null types in request body

medric:
This PR adds a linting rule to prevent null types from being used in request body objects. As per 
The API guidelines, omittable fields should be used instead.

### DIFF
--- a/lib/src/linting/rules.ts
+++ b/lib/src/linting/rules.ts
@@ -8,13 +8,13 @@ import { noOmittableFieldsWithinResponses } from "./rules/no-omittable-fields-wi
 import { oneSuccessResponsePerEndpoint } from "./rules/one-success-response-per-endpoint";
 
 export const availableRules = {
+  "has-discriminator": hasDiscriminator,
   "has-request-payload": hasRequestPayload,
   "has-response-payload": hasResponsePayload,
-  "has-discriminator": hasDiscriminator,
   "no-nested-types-within-unions": noNestedTypesWithinUnions,
   "no-nullable-arrays": noNullableArrays,
-  "no-omittable-fields-within-responses": noOmittableFieldsWithinResponses,
   "no-nullable-fields-within-requests": noNullableFieldsWithinRequests,
+  "no-omittable-fields-within-responses": noOmittableFieldsWithinResponses,
   "one-success-response-per-endpoint": oneSuccessResponsePerEndpoint
 };
 
--- a/lib/src/linting/rules.ts
+++ b/lib/src/linting/rules.ts
@@ -3,6 +3,7 @@ import { hasRequestPayload } from "./rules/has-request-payload";
 import { hasResponsePayload } from "./rules/has-response-payload";
 import { noNestedTypesWithinUnions } from "./rules/no-nested-types-within-unions";
 import { noNullableArrays } from "./rules/no-nullable-arrays";
+import { noNullableFieldsWithinRequests } from "./rules/no-nullable-fields-within-requests";
 import { noOmittableFieldsWithinResponses } from "./rules/no-omittable-fields-within-responses";
 import { oneSuccessResponsePerEndpoint } from "./rules/one-success-response-per-endpoint";
 
@@ -13,6 +14,7 @@ export const availableRules = {
   "no-nested-types-within-unions": noNestedTypesWithinUnions,
   "no-nullable-arrays": noNullableArrays,
   "no-omittable-fields-within-responses": noOmittableFieldsWithinResponses,
+  "no-nullable-fields-within-requests": noNullableFieldsWithinRequests,
   "one-success-response-per-endpoint": oneSuccessResponsePerEndpoint
 };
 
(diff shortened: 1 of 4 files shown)


lib/src/linting/rules.ts
```diff
@@ -13,6 +14,7 @@ export const availableRules = {
   "no-nested-types-within-unions": noNestedTypesWithinUnions,
   "no-nullable-arrays": noNullableArrays,
   "no-omittable-fields-within-responses": noOmittableFieldsWithinResponses,
+  "no-nullable-fields-within-requests": noNullableFieldsWithinRequests,
```
fwouts, line 16: would you mind re-ordering so they're in alphabetical order?

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+                }
+              ])
+            })
+          }),
+          responses: []
+        })
+      ],
+      types: [
+        {
+          name: "responseBody201",
```
fwouts: did you mean to make the response nullable instead of optional?

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+import { noNullableFieldsWithinRequests } from "./no-nullable-fields-within-requests";
+
+describe("rule: no nullable fields within request body", () => {
+  test("valid for correct usage", () => {
+    const errors = noNullableFieldsWithinRequests({
+      api: fakeLocatable<ApiNode>({
+        name: fakeLocatable("example-api")
+      }),
+      endpoints: [
+        // Endpoint with response payload
```
fwouts: it looks like comments need to be updated?

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+          request: fakeLocatable<RequestNode>({
+            body: fakeLocatable<BodyNode>({
+              type: {
+                kind: TypeKind.STRING
+              }
+            })
+          }),
+          responses: []
+        }),
+        // Endpoint with reference type in response payload
```
fwouts: ditto

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
fwouts: Should this type be referenced anywhere?

lib/src/linting/rules/no-nullable-fields-within-requests.ts
```diff
@@ -0,0 +1,27 @@
...
+    flatten
+  );
+
+  const types = flatten(contract.endpoints.map(extractTypes));
+
+  const filterNullable = (t: TypeNode<UnionType>) =>
+    t.type.types.find(possibleType => isNullType(possibleType));
+
+  return types.filter(filterNullable).map(typeNode => ({
+    message: `The object type \`${typeNode.name}\` defines an nullable property. Use omittable instead.`
```
fwouts:
```suggestion
    message: `The object type \`${typeNode.name}\` defines a nullable property. Make it omittable instead.`
```

lib/src/linting/rules/no-nullable-fields-within-requests.ts
```diff
@@ -0,0 +1,27 @@
...
+export const noNullableFieldsWithinRequests: LintingRule = contract => {
+  const extractTypes = flow(
+    extractRequestType,
+    (t: TypeNode) => extractNestedUnionTypes(t, contract.types),
+    flatten
+  );
+
+  const types = flatten(contract.endpoints.map(extractTypes));
+
+  const filterNullable = (t: TypeNode<UnionType>) =>
```
fwouts: Technically this doesn't do any filtering, it just returns a truthy value or not. The filtering occurs on the next line. Maybe call it `nullablePredicate`?

fwouts (changes requested):

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
medric: my bad - this is confusing. That's actually the request body i'm describing here.

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
fwouts:
```suggestion
        // Endpoint with inline request payload
```

lib/src/linting/rules/no-nullable-fields-within-requests.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+    });
+    expect(errors).toEqual([
+      {
+        message:
+          "The object type `listUsers (request body).id` defines an nullable property. Use omittable instead."
+      }
+    ]);
+  });
+
+  test("rejects a request body object when a field is nullable instead of omittable", () => {
```
fwouts: What's the difference between this test case and the previous one? (they have exactly the same description)

fwouts (changes requested):

fwouts (approved):

Issue #406, opened 2019-08-22
[Proposal] Add support to parameter serialization rules


alephao:
## Introduction
There is no way to specify parameter serialization rules defined by openapi:

https://swagger.io/docs/specification/serialization/#query
http://spec.openapis.org/oas/v3.0.2#parameterExplode

## Proposed solution
The proposed solution is to allow the user to specify a global configuration for parameter serialization.

### Global config
Currently, spot doesn't provide a way to specify global configurations or overrides for default openapi rules, this proposal includes adding the class decorator **@config** which would be used in conjunction with the **@api** decorator. I'll discuss it's usage in the next sections.

### Specifying Parameters Serialization
To specify the parameter serialization rules, we would use an object for the key ** paramSerializationStrategy**, this object can have configurations for the serialization of `path`,  `query`, `header`, or `cookie`. 

The `query` configuration, will be composed by the keys `array` and `object`, and each of them have a limited set of possible values:

* **array:** `ampersand`, `ampersandEscaped`, `comma`, `pipe`
* **object:** `ampersand`, `comma`, `deep`

### Example
Here is an example of the configuration, the input, and the output.

**Config**
```ts
@api({ name: "My API" })
@config({
  paramSerializationStrategy: { 
    query: {
      array: "comma";
      object: "deep";
    }
  }
})
class MyApi {}
```

**Input**
```ts
// ...
request(
  @queryParams
  queryParams: {
    some_object: MyObject;
    some_array?: string[];
  }
  //...
) {}
```

**Output**
```yml
- in: query
  name: some_object
  required: true
  schema:
    $ref: '#...MyObject'
  style: deepObject # This was added due to the global rule "object: deep"
  explode: true # This was added due to the global rule "object: deep"

- in: query
  name: some_array
  required: false
  schema:
    type: array
    items:
       type: string
  style: form # This was added due to the global rule "array: comma"
  explode: false # This was added due to the global rule "array: comma"
```

lfportal:
@alephao Spot will currently interpret that example as a `search` query parameter of type `object`, with properties `type` and `explode`.

It may be a reasonable (and good practice) to assume that APIs will be designed to follow a single convention for parameter serialisation. It would be convenient to be able to define the serialisation strategy globally.

alephao:
@lfportal Thanks for clarifying! I updated the issue as per our conversation

Please review and let me know your thoughts

lfportal:
[...]

PR #417, opened 2019-08-30
[TT-422] Support for query param strategy

alephao:
Implementation of #406

* Added a snapshot test for a contract using `@config` to configure query parameter serialization for arrays

### DIFF
--- a/lib/src/__examples__/contract-with-config.ts
+++ b/lib/src/__examples__/contract-with-config.ts
@@ -1,0 +1,131 @@
+import {
+  api,
+  body,
+  config,
+  defaultResponse,
+  endpoint,
+  Float,
+  headers,
+  pathParams,
+  queryParams,
+  request,
+  response,
+  securityHeader,
+  String,
+  test
+} from "@airtasker/spot";
+import "./contract-endpoint";
+import { Address, ErrorBody, UserBody } from "./models";
+
+/** This is the company API. It does cool things */
+@api({ name: "company-api" })
+@config({
+  paramSerializationStrategy: {
+    query: {
+      array: "comma"
+    }
+  }
+})
+class ExampleApi {
+  @securityHeader
+  "x-auth-token": String;
+}
+
+/** Creates a user in a company */
+@endpoint({
+  method: "POST",
+  path: "/company/:companyId/users",
+  tags: ["Company", "User"]
+})
+class CreateUser {
+  @request
+  request(
+    @pathParams
+    pathParams: {
+      /** company identifier */
+      companyId: String;
+    },
+    @headers
+    headers: {
+      /** Auth Header */
+      "x-auth-token": String;
+    },
+    @queryParams
+    queryParams: {
+      /** a demo query param */
+      "sample-query"?: String;
+      "sample-query-array": String[];
+    },
+    /** request body */
+    @body body: CreateUserRequestBody
+  ) {}
+
+  /** Successful creation of user */
+  @response({ status: 201 })
+  successResponse(
+    @headers
+    headers: {
+      /** Location header */
+      Location: String;
+    },
+    /** User response body */
+    @body body: UserBody
+  ) {}
+
+  /** Bad request response */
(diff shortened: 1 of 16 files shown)


lib/src/__examples__/contract-with-config.ts
```diff
@@ -0,0 +1,131 @@
...
+  securityHeader,
+  String,
+  test
+} from "@airtasker/spot";
+import "./contract-endpoint";
+import { Address, ErrorBody, UserBody } from "./models";
+
+/** This is the company API. It does cool things */
+@api({ name: "company-api" })
+@config({
```
fwouts, line 22: Have you considered making `@api` accept a `config` property, instead of creating a new `@config` decorator?

lib/src/generators/contract/openapi3.ts
```diff
@@ -64,7 +68,13 @@ export function openApiV3(contractDefinition: ContractDefinition): OpenApiV3 {
         operationId: endpoint.name,
         description: endpoint.description,
         tags: endpoint.tags,
-        parameters: getParameters(contractDefinition.types, endpoint),
+        parameters: getParameters(
+          contractDefinition.types,
+          endpoint,
+          contractDefinition.config
```
fwouts: In order to reduce the amount of conditional logic here, I'd suggest making `config` and `paramSerializationStrategy` required in the models. They could be set to a reasonable default value when parsing the TypeScript AST. Does that make sense?

lib/src/__examples__/contract-with-config.ts
alephao: Yes, when I chatted with @lfportal, we considered a couple of approaches, including the one you're describing. We ended up choosing `@config` because it looked nicer than the other approaches.

Also, we discarded `config` inside `@api` because, from the spot docs, `@api` is meant to have meta information only.

Does it make sense?

fwouts: You could argue that is is meta information? I'm happy either way, so feel free to resolve this thread :)

lib/src/generators/contract/openapi3.ts
alephao: It makes sense, updated

lib/src/models/definitions.ts
```diff
@@ -19,6 +20,11 @@ export interface ApiDefinition {
   securityHeader?: SecurityHeaderDefinition;
 }
 
+export const defaultConfigDefinition = { paramSerializationStrategy: {} };
```
alephao, line 23: I added this constant here instead of using `{ paramSerializationStrategy: {} }` everywhere.

Let me know if there is an issue with this or if you have a suggestion.

lib/src/syntax/config.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  }
+})
+class CompanyApi {}
+```
+ */
+export function config(config: ConfigConfig) {
+  return (target: any) => {};
+}
+interface ConfigConfig {
```
alephao, line 23: Following the current way we call the decorator config, ended up naming this "**ConfigConfig**", let me know if this is a problem

lib/src/models/types/parameter-serialization-strategies.ts
```diff
@@ -0,0 +1,53 @@
+export interface ParamSerializationStrategy {
+  query?: QueryParamSerializationStrategy;
+}
+
+export interface QueryParamSerializationStrategy {
+  array?: QueryParamArrayStrategy;
+}
+
+/** Supported serialization strategies for arrays in query parameters */
+export type QueryParamArrayStrategy =
```
lfportal, line 10: This may be better represented as an `enum` and referenced in the `switch/case` below

lfportal: As discussed offline, `enum`s would require a more awkward contract syntax.

lfportal (approved):

PR #422, opened 2019-09-04
Definition parser specs

lfportal:
Added specs for new parsers

### DIFF
--- a/lib/src/neu/parser.ts
+++ b/lib/src/neu/parser.ts
@@ -1,7 +1,7 @@
 import * as path from "path";
 import { CompilerOptions, Project, ts } from "ts-morph";
-import { Contract } from "../definitions";
-import { parseContract } from "./contract-parser";
+import { Contract } from "./definitions";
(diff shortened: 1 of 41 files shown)


lib/src/neu/parser.ts
```diff
@@ -0,0 +1,68 @@
...
+    moduleResolution: ts.ModuleResolutionKind.NodeJs,
+    experimentalDecorators: true,
+    baseUrl: "./",
+    paths: {
+      "@airtasker/spot": [path.join(__dirname, "../lib")]
+    }
+  };
+
+  // Creates a new typescript program in memory
+  return new Project({ compilerOptions });
```
committedteadrinker, line 46: Why does it need a new typescript program? I might need to be shown what this means.

committedteadrinker (approved):

lib/src/neu/parser.ts
lfportal: As discussed offline, this holds the typescript AST of the contract files in memory

PR #426, opened 2019-09-06
Result errors

lfportal:
Refactor to convert error throwing to error return via a `Result` construct inspired by Rust.


lib/src/neu/parsers/headers-parser.ts
```diff
@@ -14,28 +15,33 @@ export function parseHeaders(
...
-        name: getPropertyName(p),
-        type: parseType(p.getTypeNodeOrThrow(), typeTable, lociTable),
-        description: pDescription && pDescription.getComment(),
-        optional: p.hasQuestionToken()
-      };
-    })
-    .sort((a, b) => (b.name > a.name ? -1 : 1));
+
+  const headers = [];
+  for (const ps of type.getProperties()) {
```
committedteadrinker: what's ps? Post script?

lib/src/neu/parsers/endpoint-parser.ts
```diff
@@ -34,47 +40,74 @@ export function parseEndpoint(
...
-    .sort((a, b) => {
-      if (a.status === b.status) {
-        throw new Error(
-          `Multiple responses found for status code ${a.status} in endpoint ${endpointName}`
-        );
-      }
-      return b.status > a.status ? -1 : 1;
-    });
+  const responses = [];
+  for (const rm of responseMethods) {
```
committedteadrinker:
```suggestion
  for (const method of responseMethods) {
```

committedteadrinker (approved):

lib/src/neu/parsers/headers-parser.ts
lfportal: property signature - will update var to `propertySignature`

PR #428, opened 2019-09-09
[TT-502] Add lint rule to ensure queryParams are not objects 

medric:
**Changes** 

As per the [api guidelines](https://github.com/airtasker/api-guidelines/blob/master/guidelines/query-parameters/object.md), this adds a lint rule to ensure there's no object in query parameters definitions. 

This also adds tests to cover the changes.

### DIFF
--- a/lib/src/linting/rules/no-objects-in-query-params.spec.ts
+++ b/lib/src/linting/rules/no-objects-in-query-params.spec.ts
@@ -109,7 +109,7 @@ describe("rule: no objects in query parameters", () => {
     expect(errors).toEqual([
       {
         message:
-          "The type `somequeryparam` is of type object. Objects are not recommended in query parameters."
+          "The type `somequeryparam` in endpoint listUsers is of type object. Objects are not recommended in query parameters."
       }
     ]);
   });
@@ -155,7 +155,7 @@ describe("rule: no objects in query parameters", () => {
     expect(errors).toEqual([
       {
         message:
-          "The type `somequeryparam` is of type object. Objects are not recommended in query parameters."
+          "The type `somequeryparam` in endpoint listUsers is of type object. Objects are not recommended in query parameters."
       }
     ]);
   });
--- a/lib/src/linting/rules/no-objects-in-query-params.spec.ts
+++ b/lib/src/linting/rules/no-objects-in-query-params.spec.ts
@@ -1,8 +1,8 @@
 import { HttpMethod } from "../../models/http";
 import {
   ApiNode,
-  QueryParamNode,
   EndpointNode,
+  QueryParamNode,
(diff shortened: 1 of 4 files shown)


lib/src/linting/rules/no-objects-in-query-params.ts
```diff
@@ -0,0 +1,22 @@
...
+import { TypeNode } from "../../models/nodes";
+import { extractQueryParams } from "../../utilities/extract-endpoint-types";
+import { extractNestedObjectTypes } from "../../utilities/extract-nested-types";
+
+/**
+ * Ensure query parameters are not objects
+ */
+export const noObjectsInQueryParams: LintingRule = contract => {
+  const extractObjectTypes = (t: TypeNode) =>
+    t && extractNestedObjectTypes(t, contract.types);
```
fwouts: Is `t &&` actually needed?

```suggestion
    extractNestedObjectTypes(t, contract.types);
```

lib/src/linting/rules/no-objects-in-query-params.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+          }),
+          responses: []
+        })
+      ],
+      types: []
+    });
+    expect(errors).toEqual([
+      {
+        message:
+          "The type `somequeryparam` is of type object. Objects are not recommended in query parameters."
```
fwouts: Could we make the error message give more context, e.g. which endpoint this belongs to and what the name of the query parameter is, so it's easier to fix the issue? Otherwise, this could be hard to address when you have a large contract with many endpoints.

fwouts (changes requested):

lfportal (approved):
LGTM with @fwouts suggestions!

lib/src/linting/rules/no-objects-in-query-params.ts
```diff
@@ -0,0 +1,29 @@
...
+    extractNestedObjectTypes(t, contract.types);
+
+  const messages = contract.endpoints.map(
+    (endpoint: Locatable<EndpointNode>) => {
+      const queryParams = extractQueryParams(endpoint);
+
+      const objectTypes = queryParams.map(extractObjectTypes);
+
+      return flatten(objectTypes).map((typeNode: TypeNode) => ({
+        message: `The type \`${typeNode.name}\` in endpoint ${endpoint.value.name.value} is of type object. Objects are not recommended in query parameters.`
```
fwouts:
```suggestion
        message: `The parameter \`${typeNode.name}\` in endpoint ${endpoint.value.name.value} is of type object. Objects are not recommended in query parameters.`
```

fwouts (approved):

PR #429, opened 2019-09-09
Config parser parity

lfportal:
Replicates `@config` parsing to the raw parser.

### DIFF
--- a/lib/src/neu/definitions.ts
+++ b/lib/src/neu/definitions.ts
@@ -3,9 +3,18 @@ import { Type } from "./types";
 export interface Contract {
   name: string;
   description?: string;
+  config: Config;
   types: Array<{ name: string; type: Type }>;
   security?: SecurityHeader;
   endpoints: Endpoint[];
+}
+
+export interface Config {
+  paramSerializationStrategy: {
+    query: {
+      array: QueryParamArrayStrategy;
+    };
+  };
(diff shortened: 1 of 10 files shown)


lib/src/neu/definitions.ts
```diff
@@ -65,4 +74,19 @@ export interface Body {
...
 
+/**
+ * Supported serialization strategies for arrays in query parameters
+ *
+ *    "ampersand": ?id=3&id=4&id=5
+ *    "ampersandEscaped": ?id=3%204%205
+ *    "comma": ?id=3,4,5
+ *    "pipe": ?id=3|4|5
+ */
+export type QueryParamArrayStrategy =
```
alephao, line 85: This looks exactly like the `QueryParamArrayStrategy` defined below, why do we need to declare it again?

https://github.com/airtasker/spot/blob/32237cda11fa7a744d63422f84a1d248700d972e/lib/src/models/types/parameter-serialization-strategies.ts#L9-L14

lfportal: When the generators switch over to the raw parser everything outside of `/neu/ `directory will get deprecated, so `neu` needs to be self contained for an easy transition

alephao (approved):

committedteadrinker (approved):

PR #431, opened 2019-09-09
Indexed access type parity

lfportal:
Brings the newer raw parser's type parser to feature parity with the old type parser. This adds the final indexed access type parsing that was introduced recently.

### DIFF
--- a/lib/src/neu/parsers/type-parser.ts
+++ b/lib/src/neu/parsers/type-parser.ts
@@ -1,5 +1,6 @@
 import {
   ArrayTypeNode,
+  IndexedAccessTypeNode,
   InterfaceDeclaration,
   LiteralTypeNode,
   TypeAliasDeclaration,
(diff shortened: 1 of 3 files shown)


lib/src/neu/parsers/type-parser.ts
```diff
@@ -398,6 +402,141 @@ function parseUnionType(
...
+): Result<string[], ParserError> {
+  if (accResult.isErr()) return accResult;
+  const acc = accResult.unwrap();
+
+  const literalTypeNode = typeNode.getIndexTypeNode();
+  if (!TypeGuards.isLiteralTypeNode(literalTypeNode)) {
+    throw new Error("expected type literal");
+  }
+  const literalTypeResult = parseLiteralType(literalTypeNode);
+  if (literalTypeResult.isErr()) return literalTypeResult;
```
medric, line 523:
```suggestion
  if (literalTypeResult.isErr()) {
   return literalTypeResult;
  }
```
To preserve consistency

medric (approved):

PR #446, opened 2019-09-17
Added verifications to raw parser

lfportal:
All generator's are first parsed and then verifications are performed against the parsed content. However the raw generator uses a newer parser but currently does not perform the verifications. The raw generator would parse the contract twice (once using the older parser, once using the newer) to temporarily ensure verification is performed. This PR adds the same verifications directly into the newer parser which brings parity to the old and new parsers. As a result the raw generator no longer parses the contract twice.

### DIFF
--- a/lib/src/neu/parsers/endpoint-parser.ts
+++ b/lib/src/neu/parsers/endpoint-parser.ts
@@ -86,14 +86,14 @@ export function parseEndpoint(
   // Check request path params cover the path dynamic components
   const pathParamsInPath = getDynamicPathComponents(path);
   const pathParamsInRequest = request
-    ? request.pathParams.map(pp => pp.name)
+    ? request.pathParams.map(pathParam => pathParam.name)
     : [];
 
   const exclusivePathParamsInPath = pathParamsInPath.filter(
-    pp => !pathParamsInRequest.includes(pp)
+    pathParam => !pathParamsInRequest.includes(pathParam)
   );
   const exclusivePathParamsInRequest = pathParamsInRequest.filter(
-    pp => !pathParamsInPath.includes(pp)
+    pathParam => !pathParamsInPath.includes(pathParam)
   );
   if (exclusivePathParamsInPath.length !== 0) {
     return err(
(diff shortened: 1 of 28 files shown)


lib/src/neu/types.ts
```diff
@@ -37,6 +37,8 @@ export type Type =
   | UnionType
   | ReferenceType;
 
+export type ConcreteRootType = Exclude<Type, UnionType | ReferenceType>;
```
committedteadrinker: Please add a comment above this to explain what 'concrete' means

lib/src/neu/parsers/endpoint-parser.ts
```diff
@@ -28,118 +32,246 @@ export function parseEndpoint(
...
-    decorator
+  // Handle path
+  const pathResult = extractEndpointPath(decoratorConfig);
+  if (pathResult.isErr()) return pathResult;
+  const path = pathResult.unwrap();
+
+  // Check request path params cover the path dynamic components
+  const pathParamsInPath = getDynamicPathComponents(path);
+  const pathParamsInRequest = request
+    ? request.pathParams.map(pp => pp.name)
```
committedteadrinker: per person name/ pretty name

lib/src/neu/parsers/endpoint-parser.ts
```diff
@@ -28,118 +32,246 @@ export function parseEndpoint(
...
+): Result<string, ParserError> {
+  const pathProp = getObjLiteralPropOrThrow<EndpointConfig>(
+    decoratorConfig,
+    "path"
+  );
+  const pathLiteral = getPropValueAsStringOrThrow(pathProp);
+  const path = pathLiteral.getLiteralText();
+  const dynamicComponents = getDynamicPathComponents(path);
+
+  const duplicateDynamicComponents = [
```
committedteadrinker, line 213: `duplicated`

lib/src/neu/parsers/headers-parser.ts
```diff
@@ -25,27 +27,82 @@ export function parseHeaders(
...
 
-  // TODO: add loci information
   return ok(headers.sort((a, b) => (b.name > a.name ? -1 : 1)));
 }
+
+function extractHeaderName(
+  propertySignature: PropertySignature
+): Result<string, ParserError> {
+  const name = getPropertyName(propertySignature);
+  if (!/^[\w-]*$/.test(name)) {
```
committedteadrinker, line 61: I trust you with this regex

committedteadrinker (approved):

PR #450, opened 2019-09-20
TT-557 Add CLI command to start contract validation server

alephao:
## Description, Motivation, and Context
Initial work on the validation server feature.

* Added a function to create an express server
* The server contains only one endpoint `/health` which always returns 200
* Added a CLI command `spot serve CONTRACT_PATH [-p PORT]`
* The command generates a raw contract and saves it to `.spot/raw.json`

**Please, let me know if you have any suggestions/concerns about:**
* The command name `serve`
* The path to save the raw contract `.spot/raw.json`

I didn't put much thought on those since they are easy to change, and I wanted to hear your opinions.

### DIFF
--- a/cli/src/commands/serve.ts
+++ b/cli/src/commands/serve.ts
@@ -1,7 +1,7 @@
 import { Command, flags } from "@oclif/command";
-import { startValidationServer } from "../../../lib/src/validation-server/server";
-import { parse } from "../../../lib/src/neu/parser";
 import { outputFile } from "../../../lib/src/io/output";
+import { parse } from "../../../lib/src/neu/parser";
+import { runValidationServer } from "../../../lib/src/validation-server/server";
 
 const ARG_API = "spot_contract";
 
@@ -42,6 +42,6 @@ export default class Serve extends Command {
     outputFile(".spot", "raw.json", rawContract, true);
 
     this.log("Starting spot validation server...");
-    startValidationServer(port);
+    runValidationServer(port, this);
   }
 }
--- a/cli/src/commands/serve.ts
+++ b/cli/src/commands/serve.ts
@@ -6,10 +6,10 @@ import { outputFile } from "../../../lib/src/io/output";
 const ARG_API = "spot_contract";
 
 /**
- * oclif command to start the validation server for spot contracts
+ * oclif command to start the spot contract validation server
  */
 export default class Serve extends Command {
-  static description = "Validate a Spot contract";
+  static description = "Start the spot contract validation server";
 
   static examples = ["$ spot serve api.ts"];
 
--- a/cli/src/commands/serve.ts
+++ b/cli/src/commands/serve.ts
@@ -1,5 +1,4 @@
 import { Command, flags } from "@oclif/command";
-import { outputFile } from "../../../lib/src/io/output";
 import { parse } from "../../../lib/src/neu/parser";
 import { runValidationServer } from "../../../lib/src/validation-server/server";
 
@@ -38,12 +37,10 @@ export default class Serve extends Command {
 
     try {
       this.log("Generating raw contract...");
-      const rawContract = JSON.stringify(parse(contractPath));
-      this.debug("Outputing raw contract to ./spot");
-      outputFile(".spot", "raw.json", rawContract, true);
+      const contract = parse(contractPath);
 
       this.log("Starting validation server...");
-      await runValidationServer(port, this).defer();
+      await runValidationServer(port, contract, this).defer();
       this.log(`Validation server running on port ${port}`);
     } catch (e) {
       this.error(e, { exit: 1 });
--- a/cli/src/commands/serve.ts
+++ b/cli/src/commands/serve.ts
@@ -1,0 +1,36 @@
+import { Command, flags } from "@oclif/command";
+
+const ARG_API = "spot_contract";
+
+/**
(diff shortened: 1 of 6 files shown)


cli/src/commands/serve.ts
```diff
@@ -0,0 +1,52 @@
...
+  async run() {
+    const { args, flags } = this.parse(Serve);
+    const contractPath = args[ARG_API];
+    const { port } = flags;
+
+    try {
+      this.log("Generating raw contract...");
+      const rawContract = JSON.stringify(parse(contractPath));
+      this.debug("Outputing raw contract to ./spot");
+      outputFile(".spot", "raw.json", rawContract, true);
```
fwouts: Why do we still need to generate the raw file?

alephao: Now that you said that, I see we could just use the contract in memory instead, that's what you were thinking?

lfportal (approved):

cli/src/commands/serve.ts
```diff
@@ -0,0 +1,49 @@
+import { Command, flags } from "@oclif/command";
+import { parse } from "../../../lib/src/neu/parser";
+import { runValidationServer } from "../../../lib/src/validation-server/server";
+
+const ARG_API = "spot_contract";
+
+/**
+ * oclif command to start the spot contract validation server
+ */
+export default class Serve extends Command {
```
fwouts: We have different types of Spot server, for example `spot mock` also starts a server. Ultimately I think we should merge them under different paths. In the meantime, I'd recommend something less general e.g. `spot validation-server`.

cli/src/commands/serve.ts
```diff
@@ -0,0 +1,49 @@
...
+    const { args, flags } = this.parse(Serve);
+    const contractPath = args[ARG_API];
+    const { port } = flags;
+
+    try {
+      this.log("Generating raw contract...");
+      const contract = parse(contractPath);
+
+      this.log("Starting validation server...");
+      await runValidationServer(port, contract, this).defer();
```
fwouts: Just wondering if we can do something a bit simpler, why do we need the `refer()` indirection?

cli/src/commands/serve.ts
```diff
@@ -0,0 +1,49 @@
...
+    })
+  };
+
+  async run() {
+    const { args, flags } = this.parse(Serve);
+    const contractPath = args[ARG_API];
+    const { port } = flags;
+
+    try {
+      this.log("Generating raw contract...");
```
fwouts:
```suggestion
      this.log("Parsing contract...");
```

cli/src/commands/serve.ts
```diff
@@ -0,0 +1,49 @@
...
+      const contract = parse(contractPath);
```
fwouts: Does this validate the contract too? (I'm not familiar with the `neu` architecture)

lib/src/validation-server/server.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe("Server", () => {
+  const mockLogger = {
+    log: (message: string) => message
+  };
+
+  const contract = parse(CONTRACT_PATH);
+
+  describe("Run", () => {
+    it("/health and return 200", done => {
+      const { app } = runValidationServer(5907, contract, mockLogger);
```
fwouts: It's easy to think that the number `5907` has a specific meaning for the test. Maybe put it in a constant called `DUMMY_PORT`?

lib/src/validation-server/server.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      request(app)
+        .get("/health")
+        .expect(200)
+        .then(_ => {
```
fwouts: This should work too FYI:

```suggestion
        .then(done);
```

Or you could make the test function async, e.g:
```typescript
it(..., async () => {
  const { app } = ...
  await request(app).get("/health").expect(200);
});
```

lib/src/validation-server/server.ts
```diff
@@ -0,0 +1,23 @@
...
+    res.status(200).end();
+  });
+
+  return {
+    app,
+    defer: () => new Promise(resolve => app.listen(port, resolve))
+  };
+}
+
+export interface Logger {
```
fwouts: Is this useful to other parts of the Spot codebase? Should we consider moving it to a more central spot?

fwouts (approved):

lib/src/validation-server/server.ts
alephao: Yes, I thought about that too, I'm happy to move it

lib/src/validation-server/server.spec.ts
alephao: Much better, thanks!

alephao: The second approach, using `await` doesn't really work, the method is not `async`

cli/src/commands/serve.ts
alephao: That's a way to wait for the server to start before continuing, what would be a better way of doing it? I basically just followed the implementation on `mock.ts`

committedteadrinker (approved):

cli/src/commands/serve.ts
lfportal: Parse will throw an error if the contract is not valid.

fwouts (approved):

PR #470, opened 2019-10-08
[TT-558] Validate user inputs

medric:

This implements a module to validate user inputs. Inputs are to be strings. The motivation for using this module is to allow for contract testing and ensure inputs are valid and match a given contract.

For instance, the following input query parameters : 

`{ id: '12', name: 'medric' }` could get validated against type `{ id: int64Type(), name: stringType() }` by doing `validator.run({ id: '12', name: 'medric'}, { id: int64Type(), name: stringType() }`.

## Checklist:


lib/src/neu/verifications/validator.ts
```diff
@@ -0,0 +1,130 @@
...
+        )
+        .some(v => v === false)
+    );
+  }
+
+  validateReference(input: Input, type: ReferenceType) {
+    return this.run(input, dereferenceType(type, this.typeTable));
+  }
+
+  run(input: Input, type: Type, isMandatory: boolean = true): boolean | void {
```
lfportal: why is `run` able to return `void`?

lib/src/neu/verifications/validator.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    if (type.kind === REFERENCE) {
+      return this.validateReference(input, type);
+    }
+
+    const validator = Validator.validatorMap[type.kind];
+
+    if (typeof validator !== "function") {
+      return;
```
lfportal: what happens if `validator === undefined`? 

is this intended to capture this case - `[STRING]: Boolean`?

lfportal (changes requested):

lib/src/neu/verifications/validator.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      );
+      expect(result).toBe(true);
+    });
+
+    test("should return true when value matches a reference type", () => {
+      const result = validator.run(
+        { name: "param", value: { id: "false", name: "" } },
+        referenceType("obj1")
+      );
+      expect(result).toBe(false);
```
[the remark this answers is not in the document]
medric: My bad, I forgot to update that one.

lib/src/neu/verifications/validator.ts
medric: What do you mean by that ? If `validator === undefined` then the condition is met and the function returns undefined. Actually we should ensure we handle every types. Otherwise I can change the logic and throw an error instead.

lib/src/neu/verifications/string-validator.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  messages: string[] = [];
+  typeTable: TypeTable;
+
+  constructor(typeTable: TypeTable) {
+    this.typeTable = typeTable;
+  }
+
+  run(input: Input, type: Type, isMandatory: boolean = true): boolean | void {
```
lfportal:
```suggestion
  run(input: Input, type: Type, isMandatory: boolean = true): boolean {
```

lfportal (approved):

lib/src/neu/verifications/string-validator.ts
```diff
@@ -0,0 +1,134 @@
...
+          this.run(
+            {
+              name: `.${input.name}.${p.name}`,
+              value: `${(input.value as { [key: string]: unknown })[p.name]}`
+            },
+            p.type,
+            !p.optional
+          )
+        )
+        .some(v => v === false)
```
committedteadrinker: As discussed, please put this in another function. It was hard to read.

lib/src/neu/verifications/string-validator.ts
```diff
@@ -0,0 +1,134 @@
...
+    return (
+      Array.isArray(input.value) &&
+      !input.value
+        .map((v, index) =>
+          this.run(
+            { name: `${input.name}[${index}]`, value: `${v}` },
+            type.elementType
+          )
+        )
+        .some(v => v === false)
```
committedteadrinker: These ones as well. 👍

committedteadrinker (approved):
Other than my comments, looks good otherwise. Those comments are nitpicks.

PR #472, opened 2019-10-08
TT-567 Add a spot contract to validation-server's health check

alephao:
Adding spot contracts to the validation-server. The only endpoint we currently have is the health check endpoint, so I'm adding an @endpoint for it.

I also added another CircleCI job to validate the contract, so we can be confident that a pull-request is not breaking it.

### DIFF
--- a/lib/src/validation-server/spots/health.ts
+++ b/lib/src/validation-server/spots/health.ts
@@ -1,7 +1,7 @@
 import { endpoint, request, response } from "../../lib";
 
 @endpoint({
-  method: "POST",
+  method: "GET",
   path: "/health"
 })
 export class HealthCheck {
--- a/lib/src/validation-server/spots/health.ts
+++ b/lib/src/validation-server/spots/health.ts
@@ -1,0 +1,13 @@
+import { endpoint, request, response } from "../../lib";
+
+@endpoint({
+  method: "POST",
+  path: "/health"
+})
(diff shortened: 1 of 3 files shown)


lib/src/validation-server/spots/health.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@endpoint({
+  method: "POST",
+  path: "/health"
+})
+export class HealthCheck {
+  @request
+  request() {}
+
+  @response({ status: 200 })
```
fwouts, line 11: What's the status when it's not healthy?

alephao: This is just to check if the server was initialized or not, so if the server is not available it will return 503 I guess

lib/src/validation-server/spots/health.ts
```diff
@@ -0,0 +1,13 @@
+import { endpoint, request, response } from "../../lib";
+
+@endpoint({
+  method: "POST",
```
fwouts: Shouldn't this be GET?

fwouts (approved):

lib/src/validation-server/spots/health.ts
alephao: Yes 🤦‍♂

lfportal (approved):

PR #473, opened 2019-10-09
TT-528 Add spot contract for the validation endpoint

alephao:
Hi, this PR is adding the spot contract for the validation server's `validation` endpoint. This is the endpoint we'll use to confirm that a pair of response and request matches an endpoint definition in the contract.

I have an initial draft of the validate endpoint and I wanted to get some early feedback on the interface

I wrote some reasoning on my choices below, please let me know if you think there is a better way:

**Headers** type
- Using a String[] for the Headers type since Spot doesn't seem to support indexed types
- An example of how we would use it:
```
request: {
  headers: ["Auth-Token: 123", "Content-Type: text/html; charset=UTF-8"]
}
```

**UnprocessableEntityError /  BaseError**
- Following a pattern that we're using on other places, an interface with `error_code`, `error_messages`, and `type`

**ValidateRequest.request.body** type
- Not sure what's the best type to use here. I left it as a String, we can throw a 422 if the value is an invalid JSON

### DIFF
--- a/lib/src/validation-server/spots/utils.ts
+++ b/lib/src/validation-server/spots/utils.ts
@@ -1,0 +1,8 @@
+import { String } from "../../lib";
+
+export interface UnprocessableEntityError {
+  error_code: String;
+  error_messages: String[];
+}
+
+export type Headers = String[];
--- a/lib/src/validation-server/spots/utils.ts
+++ b/lib/src/validation-server/spots/utils.ts
@@ -9,4 +9,7 @@ export interface UnprocessableEntityError extends BaseError {
   type: "unprocessable_entity";
 }
 
-export type Headers = String[];
+export interface Header {
+  key: String;
+  value: String;
+}
--- a/lib/src/validation-server/spots/utils.ts
+++ b/lib/src/validation-server/spots/utils.ts
@@ -1,8 +1,12 @@
 import { String } from "../../lib";
 
-export interface UnprocessableEntityError {
+export interface BaseError {
   error_code: String;
   error_messages: String[];
+}
+
+export interface UnprocessableEntityError extends BaseError {
+  type: "unprocessable_entity";
 }
 
 export type Headers = String[];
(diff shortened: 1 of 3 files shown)


driesg:
> Using a String[] for the Headers type since Spot doesn't seem to support indexed types

I think this is fine. Could you create a ticket in the Spot epic to support this? This ticket should include: 
- use case
- example API (how would we use it?)
- details on why spot doesn't support it at the moment (if you know)

> UnprocessableEntityError / BaseError

👍 

> ValidateRequest.request.body type
> Not sure what's the best type to use here. I left it as a String, we can throw a 422 if the value is an invalid JSON

Wouldn't this be an Object?

lib/src/validation-server/spots/validate.ts
```diff
@@ -0,0 +1,38 @@
...
+  request: {
+    method: HttpMethod;
+    path: String;
+    headers: Headers;
+    body: String;
+  };
+  response: {
+    status: {
+      code: Integer;
+    };
```
lfportal: doesn't look like there's any benefit from wrapping `code` in a `status` object.

lib/src/validation-server/spots/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface BaseError {
+  error_code: String;
+  error_messages: String[];
+}
+
+export interface UnprocessableEntityError extends BaseError {
+  type: "unprocessable_entity";
+}
+
+export type Headers = String[];
```
lfportal: For a stronger contract we can consider something like:
```suggestion
export type Headers = Header[];
interface Header = {
  name: String;
  value: String;
}
```

alephao: Great suggestion, sounds obvious now 😂

lib/src/validation-server/spots/validate.ts
alephao: I agree, will remove

alephao:
@driesg 
- After seeing Leslie's suggestion, I think we don't really need to support it
- Spot doesn't support an "Object/any/unknown" type yet :(

lfportal (approved):

PR #476, opened 2019-10-10
[TT-543] Find contract mismatches on request query params

medric:

This PR adds validation logic to the contract testing module in order to ensure input query parameters match a given contract / endpoint definition.

- What does it do?

It takes a map of query parameters and validate it against a contract. Validation is repeated for every endpoint defined in the contract. Query parameters values are string so we leverage the string validation module to check if the inputs match the right patterns. 

## Checklist:


lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -328,6 +344,76 @@ export class ContractMismatcher {
...
+
+    return { comma };
+  }
+
+  private findMismatchOnRequestQueryParams(
+    endpoint: Endpoint,
+    userInputRequest: UserInputRequest
+  ): Result<Mismatch[], Error> {
+    const queryParamsString = url.parse(userInputRequest.path).query || "";
+    const contractQueryParams = endpoint.request!!.queryParams;
```
lfportal, line 359: There is no guarantee `request` exists here.

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -328,6 +344,76 @@ export class ContractMismatcher {
...
+    let mismatches: Mismatch[] = [];
+    for (const {
+      name: queryParamName,
+      optional,
+      type: contractQueryParamType
+    } of contractQueryParams) {
+      const requestQueryParam = queryParams[queryParamName];
+
+      // Query parameter is mandatory and hasn't been provided
+      if (typeof requestQueryParam === "undefined" && !optional) {
```
lfportal:
```suggestion
      if (requestQueryParam === undefined && !optional) {
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -328,6 +344,76 @@ export class ContractMismatcher {
...
+        mismatches.push(
+          new Mismatch(
+            `Query parameter "${queryParamName}" is required but hasn't been provided.`
+          )
+        );
+      }
+
+      if (typeof requestQueryParam === "undefined") {
```
lfportal:
```suggestion
      if (requestQueryParam === undefined) {
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -328,6 +344,76 @@ export class ContractMismatcher {
...
+        continue;
+      }
+
+      // Validate current request query param against contract
+      result = this.findMismatchOnStringContent(
+        { name: queryParamName, value: requestQueryParam },
+        contractQueryParamType
+      );
+
+      queryParams[queryParamName] = null;
```
lfportal: It is unconventional and not clear why we are mutating the `queryParams` object here. Instead can `checkForNonExistingParams` be refactored to be perform all the logic required using the unmodified `queryParams` and `contractQueryParams`?

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -365,6 +452,25 @@ export class ContractMismatcher {
     }
   }
 
+  private findMismatchOnStringContent(
+    content: StringInput,
+    contractContentTypeToCheckWith: Type
+  ): Result<Mismatch[], Error> {
+    if (!content) {
+      return ok([]);
+    }
```
lfportal, line 471: why is lack of `content` automatically valid?

lfportal (changes requested):

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -328,6 +344,76 @@ export class ContractMismatcher {
...
+        mismatches.push(
+          new Mismatch(
+            `Query parameter "${queryParamName}" is required but hasn't been provided.`
+          )
+        );
+      }
+
+      if (typeof requestQueryParam === "undefined") {
+        continue;
+      }
```
committedteadrinker: nit: this is kind of hard to read as I'd need to understand the if method before this one first.
I'd prefer putting this as a nested if statement with everything below.

Not a big deal, it's just hard to read and understand the more this codebase is going to grow later on.

lib/src/neu/verifications/contract-mismatcher.ts
medric: I'd stick to using `typeof`,  as
1. Strict equality operator can go wrong if the variable is undeclared
2. `undefined` could be shadowed or maybe redefined.

`typeof` will always work, it could never go wrong

lib/src/neu/verifications/contract-mismatcher.ts
medric: same reason as above

committedteadrinker (approved):

lib/src/neu/verifications/contract-mismatcher.ts
medric: hmm this has to be reworked then  i guess, i re-used the existing logic.

lib/src/neu/verifications/contract-mismatcher.ts
medric: hey - i prefer linear code and tend to avoid nested branching. I'll have a look and will try to make this easier to read.

lfportal (approved):

PR #480, opened 2019-10-15
Use string based validators for headers and path params

medric:

This revisits the validation logic that ensures path parameters and headers are valid according to a given contract. It leverages string validators as these kinds of inputs are strings (according to HTTP standards).

## Checklist:

- [] I've created an issue associated with this PR

### DIFF
--- a/lib/src/neu/verifications/contract-mismatcher.ts
+++ b/lib/src/neu/verifications/contract-mismatcher.ts
@@ -149,7 +149,7 @@ export class ContractMismatcher {
         const result = this.findMismatchOnStringContent(
           {
             name: userInputHeaderKey,
-            value: userInputRequest.headers[userInputHeaderKey] as string
+            value: userInputRequest.headers[userInputHeaderKey]
           },
           contractHeaderType.unwrap()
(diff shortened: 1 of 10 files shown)


lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -146,8 +146,11 @@ export class ContractMismatcher {
           return accumulator;
         }
 
-        const result = this.findMismatchOnContent(
-          userInputRequest.headers[userInputHeaderKey],
+        const result = this.findMismatchOnStringContent(
+          {
+            name: userInputHeaderKey,
+            value: userInputRequest.headers[userInputHeaderKey] as string
```
lfportal: is there a way to make this more typesafe?

medric: What do you mean by this ? Can you be bit more specific please ?

committedteadrinker (approved):
I think we've discussed most of these offline. LGTM

lfportal (approved):

PR #509, opened 2019-10-23
has-discriminator linting rule port to neu namespace

lfportal:
This PR ports the has-discriminator linting rule to the neu namespace to use the new version of the contract definitions. As it is the first linting rule to be ported, some extra setup files are also included.

### DIFF
--- a/lib/src/neu/types.ts
+++ b/lib/src/neu/types.ts
@@ -355,7 +355,7 @@ export function inferDiscriminator(
   types: Type[],
   typeTable: TypeTable
 ): string | undefined {
-  const conreteRootTypes = types.reduce<ConcreteType[]>((acc, type) => {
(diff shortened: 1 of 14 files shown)


lib/src/neu/types.ts
```diff
@@ -340,6 +345,69 @@ export function dereferenceType(
...
+ * Given a list of types, try to find a disriminator.
+ *
+ * @param types list of types
+ * @param typeTable a TypeTable
+ */
+export function inferDiscriminator(
+  types: Type[],
+  typeTable: TypeTable
+): string | undefined {
+  const conreteRootTypes = types.reduce<ConcreteType[]>((acc, type) => {
```
medric: typo - `concreteRootTypes`

medric (approved):

PR #510, opened 2019-10-24
TT-585 Fix path params matcher matching wrong paths

alephao:
The path params matcher was matching wrong paths, for example, `/z/x/w/a/0/c` would match `/a/:b/c`.

This PR fixes this issue by correcting the regexp used for matching, and add tests for a couple of cases.

There are other small changes on this PR:
- Changed the error message when the matcher fails to find an endpoint on the contract matching the request
- Refactored the regexp part of the code
- Removed an unused method

## Checklist:

### DIFF
--- a/lib/src/neu/verifications/contract-mismatcher.ts
+++ b/lib/src/neu/verifications/contract-mismatcher.ts
@@ -20,7 +20,6 @@ import {
 } from "./user-input-models";
 
 export class ContractMismatcher {
-  private readonly PATH_PARAM_REGEX = /:[^\/]*/g;
   private typeTable: TypeTable;
 
   constructor(private readonly contract: Contract) {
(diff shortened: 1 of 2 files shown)


lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -573,54 +570,44 @@ export class ContractMismatcher {
...
-  }
-
-  private getSeparatedPathFromQueries(path: string): string[] {
-    return path.split("?");
-  }
 }
 
+// Transform /a/:b/c/:d/e -> /^/a/\w+/c/\w+/e$/
+const regexForVariablePath = (path: string): RegExp => {
+  const regexp = path.replace(/:[^\/]+/g, "\\w+");
```
lfportal: The path segment of a uri may contain characters beyond `\w`. See https://stackoverflow.com/questions/4669692/valid-characters-for-directory-part-of-a-url-for-short-links

lfportal (changes requested):

lfportal (approved):

PR #524, opened 2019-10-31
[TT-598] Add support for date & date time formats in JSON Schema validator

medric:

This adds support for date and date-time formats when validating JSON schemas. 

## Checklist:


lib/src/generators/contract/__snapshots__/json-schema.spec.ts.snap
```diff
@@ -233,6 +236,8 @@ definitions:
             $ref: '#/definitions/Email'
           address:
             $ref: '#/definitions/Address'
+          createdAt:
+            type: string
```
alephao, line 240: Shouldn't the changes have generated the `format` too?
```
type: string
format: date
```

medric: hey @alephao, this is the old implementation. We should not use it anymore. I'll create a ticket to clean this up.

alephao (approved):

lfportal (approved):

PR #528, opened 2019-11-01
Add contextual data to validation server error messages

lfportal:
The messages returned from validation server currently contain very little contextual data to aid in resolving contract violations. This PR adds some contextual data including:
- Details when a type violation occurs
- The received interaction (request/response)
- The `endpoint` matched

## Checklist:

### DIFF
--- a/lib/src/neu/verifications/contract-mismatcher.ts
+++ b/lib/src/neu/verifications/contract-mismatcher.ts
@@ -116,7 +116,8 @@ export class ContractMismatcher {
             requestHeaderTypeMismatch(
               `Request header "${m.header}" type mismatch: ${m.mismatches.join(
                 ", "
-              )}`
+              )}`,
+              m.mismatches
             )
           );
           return;
@@ -154,7 +155,8 @@ export class ContractMismatcher {
             responseHeaderTypeMismatch(
               `Response header "${m.header}" type mismatch: ${m.mismatches.join(
                 ", "
-              )}`
+              )}`,
+              m.mismatches
             )
           );
           return;
@@ -184,7 +186,8 @@ export class ContractMismatcher {
             requestBodyTypeMismatch(
               `Request body type mismatch:\n${m.data}\n${m.mismatches
                 .map(mismatch => `- ${mismatch}`)
-                .join("\n")}`
+                .join("\n")}`,
+              m.mismatches
             )
           );
           return;
@@ -213,7 +216,8 @@ export class ContractMismatcher {
             responseBodyTypeMismatch(
               `Response body type mismatch:\n${m.data}\n${m.mismatches
                 .map(mismatch => `- ${mismatch}`)
-                .join("\n")}`
+                .join("\n")}`,
+              m.mismatches
             )
           );
           return;
@@ -237,7 +241,8 @@ export class ContractMismatcher {
             pathParamTypeMismatch(
               `Path param "${m.pathParam}" type mismatch: ${m.mismatches.join(
                 ", "
-              )}`
+              )}`,
+              m.mismatches
             )
           );
           return;
@@ -275,7 +280,8 @@ export class ContractMismatcher {
             queryParamTypeMismatch(
               `Query param "${m.queryParam}" type mismatch: ${m.mismatches.join(
                 ", "
-              )}`
+              )}`,
+              m.mismatches
             )
           );
           return;
--- a/lib/src/neu/verifications/contract-mismatcher.ts
+++ b/lib/src/neu/verifications/contract-mismatcher.ts
@@ -1,4 +1,5 @@
-import JsonSchemaValidator, { ErrorObject } from "ajv";
+import JsonSchemaValidator from "ajv";
+import assertNever from "assert-never";
 import qs from "qs";
 import * as url from "url";
 import {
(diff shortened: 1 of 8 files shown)


lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -140,16 +318,26 @@ export class ContractMismatcher {
...
+          iH =>
+            !contractHeaders.some(
+              header => header.name.toLowerCase() === iH.name.toLowerCase()
+            )
+        )
         .forEach(iH => {
-          mismatches.push(
-            new Mismatch(`Header "${iH.name}" not found on endpoint`)
-          );
+          mismatches.push(new UndefinedHeader(iH.name));
```
alephao: Same here
```suggestion
          mismatches.push(undefinedHeader(iH.name));
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -195,7 +383,14 @@ export class ContractMismatcher {
           { name: contractPathParam.name, value: inputPathArray[i] },
           contractPathParamType
         );
-        mismatches.push(...pathParamMismatches);
+        if (pathParamMismatches.length > 0) {
+          mismatches.push(
+            new PathParamTypeMismatch(
```
alephao:
```suggestion
            pathParamTypeMismatch(
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -205,21 +400,21 @@ export class ContractMismatcher {
...
+  ): Result<Array<UndefinedBody | BodyTypeMismatch>, Error> {
     if (contractBody === undefined) {
       if (inputBody === undefined) {
         return ok([]);
       }
       if (strict) {
-        return ok([
-          new Mismatch("contract body has no type, but a body was given")
-        ]);
+        return ok([new UndefinedBody()]);
```
alephao:
```suggestion
        return ok([undefinedBody()]);
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -292,37 +506,35 @@ export class ContractMismatcher {
...
-      Object.entries(verifiedQueryParams)
-        .filter(([_, value]) => !value)
-        .map(
-          ([key, _]) =>
-            new Mismatch(
-              `Query parameter "${key}" does not exist under the specified endpoint.`
-            )
+      if (result.length > 0) {
+        mismatches.push(
+          new QueryParamTypeMismatch(
```
alephao:
```suggestion
          queryParamTypeMismatch(
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -292,37 +506,35 @@ export class ContractMismatcher {
...
         );
+      }
+    }
 
-    mismatches = [...mismatches, ...checkForNonExistingParams()];
+    Object.entries(verifiedQueryParams)
+      .filter(([_, value]) => !value)
+      .map(([key, _]) => key)
+      .forEach(key => {
+        mismatches.push(new UndefinedQueryParam(key));
```
alephao:
```suggestion
        mismatches.push(undefinedQueryParam(key));
```

lib/src/neu/verifications/contract-mismatcher.ts
```diff
@@ -423,3 +623,60 @@ export const pathMatchesVariablePath = (
...
+  readonly kind = MismatchKind.REQUIRED_QUERY_PARAM_MISSING;
+  constructor(readonly queryParam: string) {}
+}
+
+class UndefinedQueryParam {
+  readonly kind = MismatchKind.UNDEFINED_QUERY_PARAM;
+  constructor(readonly queryParam: string) {}
+}
+
+class QueryParamTypeMismatch {
```
alephao: Should these classes be moved to `violations.ts` as well? They seem to be related

alephao (changes requested):
In some places, you're using a function to create the violation object, but in other places you're using the object initializer `new MyObject(...)`. Is there a reason for that? If not, can we use only one way for consistency?

I highlighted some examples in the comments below

lib/src/neu/verifications/mismatches.ts
```diff
@@ -0,0 +1,138 @@
...
+  HEADER_TYPE_DISPARITY = "header_type_disparity",
+  PATH_PARAM_TYPE_DISPARITY = "path_param_type_disparity",
+  REQUIRED_QUERY_PARAM_MISSING = "required_query_param_missing",
+  UNDEFINED_QUERY_PARAM = "undefined_query_param",
+  QUERY_PARAM_TYPE_DISPARITY = "query_param_type_disparity",
+  UNDEFINED_BODY = "undefined_body",
+  BODY_TYPE_DISPARITY = "body_type_disparity"
+}
+
+export interface RequiredHeaderMissingMismatch {
```
alephao, line 20: I saw you renamed `Mismatch` to `DisparityViolation`, should the interfaces in this file also be renamed?

Look the comment below for an example👇

lib/src/neu/verifications/mismatches.ts
```diff
@@ -0,0 +1,138 @@
...
+  kind: MismatchKind.REQUIRED_HEADER_MISSING;
+  header: string;
+}
+
+export interface UndefinedHeaderMismatch {
+  kind: MismatchKind.UNDEFINED_HEADER;
+  header: string;
+}
+
+export interface HeaderTypeDisparityMismatch {
```
alephao, line 30: For example, should this be `HeaderTypeDisparityViolation` instead?

lib/src/neu/verifications/mismatches.ts
```diff
@@ -0,0 +1,138 @@
...
+  typeDisparities: string[];
+}
+
+export interface PathParamTypeDisparityMismatch {
+  kind: MismatchKind.PATH_PARAM_TYPE_DISPARITY;
+  pathParam: string;
+  typeDisparities: string[];
+}
+
+export interface RequiredQueryParamMissingMismatch {
```
alephao, line 42: And here, should it be `RequiredQueryParamMissingViolation` instead of `*MissingMismatch`?

alephao (approved):
Looks good, I just have a question related to naming

lib/src/neu/verifications/mismatches.ts
alephao: explained offline, I didn't saw the comment on the file

medric (approved):
As discussed offline, we should decouple the logic here and think about refactoring this so it's easier to follow. Maybe we could have 2 separate modules, in different files. One for input mismatches, more generic that returns mismatches objects, and one more specific to a contract violations. IMO, it would help to distinguish between `Mismatch` and `Violation`.  Can you create a ticket for this ?

PR #600, opened 2019-12-11
[TT-482] Fix linting cli to output every error

medric:

`this.error` will cause the app to exit with non-0 code, hence the 1st linting error only being displayed. 

This is fixed by passing  down a flag to prevent the program from stopping. 

## Checklist:

### DIFF
--- a/cli/src/commands/lint.ts
+++ b/cli/src/commands/lint.ts
@@ -33,7 +33,7 @@ export default class Lint extends Command {
     const lintingErrors = lint(contract);
 
     lintingErrors.forEach(error => {
-      this.error(error.message);
+      this.error(error.message, { exit: false });
     });
   }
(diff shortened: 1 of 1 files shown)


cli/src/commands/lint.ts
```diff
@@ -33,7 +33,7 @@ export default class Lint extends Command {
     const lintingErrors = lint(contract);
 
     lintingErrors.forEach(error => {
-      this.error(error.message);
+      this.error(error.message, { exit: false });
```
lfportal, line 37: The `lint` command should exit with a non-0 exit code at the end of execution. This helps CI processes halt workflows.

lfportal (changes requested):

lfportal (approved):

PR #632, opened 2020-01-01
Neu OpenAPI 3 generator

lfportal:
Currently, the Spot's newer internal model is not used by the generators. This PR migrates the OpenAPI 3 generator to use the new internal model.

This generator deviates from the older generator in some ways:
- The OpenAPI version is set to `3.0.2` instead of `3.0.0`
- Path parameter `style` is unset (previous generator in some cases incorrectly sets `style` to `deepObject` or `pipeDelimited`)
- Query parameter `style` is set to `deepObject` for object types
- Nullable type literals (and same-type literal unions) translate into `nullable` schemas with `null` included in `enum` list
- All same-type literal (`boolean`, `float`, `int`, `string`) unions are converted into schemas with `enum` (previously only `string` literal unions) instead of `oneOf`

### Spectral
This PR introduces [Spectral](https://github.com/stoplightio/spectral) to programmatically check for OpenAPI errors. Spectral is a general purpose schema linter. Although the project uses Speccy for OpenAPI validation already, it is not as programmatically friendly (mostly used as a CLI) especially in TypeScript. Additionally Spectral will allow Spot to use the same validation library for any future non-OpenAPI specifications (e.g. Blueprint).

### Tests
There are a large number of tests, for convenience it may be simpler to refer to the test plan/outline:

Screen Shot 2020-01-06 at 1 20 19 pm
Screen Shot 2020-01-06 at 1 22 19 pm

## Checklist:


lib/src/neu/generators/openapi3/openapi3-schema.ts
```diff
@@ -0,0 +1,500 @@
+// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#openapi-object
+export interface OpenApiV3 {
```
medric: Out of scope of this PR - but should we move these type definitions out of the Spot codebase and create a separate package? My main concern here is that the codegen leverages these types and duplicates them. I feel like it's getting hard to maintain. I'm afraid our tools will end up not being aligned.

lfportal: As long as the definition used in any project is compatible with OpenAPI it shouldn't matter. However, it'd be a good idea to use exactly the same definitions. Perhaps after the dust settles on the generator migrations.

lib/src/neu/generators/openapi3/openapi3-type-util.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+}
+
+function booleanSchema(
+  opts: {
+    values?: boolean[];
+    nullable?: boolean;
+  } = {}
+): BooleanSchemaObject {
+  const enums =
```
medric: This logic is duplicated, it could be abstracted in a separate function - e.g. : 

```
const enum = createEnum(options)
```

lib/src/neu/generators/openapi3/openapi3-type-util.ts
```diff
@@ -0,0 +1,309 @@
...
+    opts.values && (opts.nullable ? [...opts.values, null] : opts.values);
```
medric: This looks a bit cryptic, hence hard to read. I'd go for something more declarative, something along the lines of 
```javascript
createEnum(options) {
 if (!options.values) {
   return;
 }

 if (options.nullable) {
   return [...options.values, null];
 }

 return options.value;
} 
```

lib/src/neu/generators/openapi3/openapi3-type-util.ts
```diff
@@ -0,0 +1,309 @@
...
+          oneOf: nonNullTypes.map(t =>
+            typeToSchemaOrReferenceObject(t, typeTable)
+          ),
+          discriminator: unionTypeToDiscrimintorObject(type, typeTable)
+        };
+      }
+  }
+}
+
+function unionTypeToDiscrimintorObject(
```
medric, line 241:
```suggestion
function unionTypeToDiscriminatorObject(
```

lib/src/neu/generators/openapi3/openapi3-type-util.ts
```diff
@@ -0,0 +1,309 @@
...
+  };
+}
+
+function referenceTypeToSchema(
+  type: ReferenceType,
+  nullable?: boolean
+): ReferenceObject {
+  return {
+    $ref: referenceObjectValue(type.name),
+    nullable: nullable || undefined
```
medric, line 306: if `nullable` is `false`, then it will evaluate to `nullable: undefined`, right?
`false || undefined === undefined`

lib/src/neu/types.ts
```diff
@@ -270,14 +282,32 @@ export function isBooleanLiteralType(type: Type): type is BooleanLiteralType {
...
+  types: Type[]
+): types is BooleanLiteralType[] {
+  return types.every(isBooleanLiteralType);
+}
+
 export function isStringType(type: Type): type is StringType {
   return type.kind === TypeKind.STRING;
 }
 
+export function isNotStringType<T extends Type>(
```
medric, line 295: Why creating another function when you just need to negate `isStringType` ?

lib/src/neu/generators/openapi3/openapi3-type-util.ts
lfportal: Yeap, strictly speaking we could put `nullable: false` but the default is also `false`

lib/src/neu/types.ts
lfportal: Using the negation in `.filter`: `types.filter(t => !isStringType(t))`. Whilst this will do the job, `types` still has type `Type[]`. 

Using a type guard function directly: `types.filter(isNotStringType)` allows type narrowing to occur. Now `types` is `StringType[]`.

lib/src/neu/generators/openapi3/openapi3-type-util.ts
medric: But it will be `undefined` is this case, not `false`. I don't think `undefined` makes sense here. If the default value is `false` just set it as a default parameter.

lib/src/neu/generators/openapi3/openapi3-type-util.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  return {
+    propertyName: unionType.discriminator,
+    mapping
+  };
+}
+
+function referenceTypeToSchema(
+  type: ReferenceType,
+  nullable?: boolean
```
medric, line 302: I would do 
```suggestion
  nullable = false
```

lib/src/neu/types.ts
medric: But you don't have to duplicate the logic, right ? For instance, you could do :
`export const isNotStringType = (type) => !isStringType(type)`
or, if using lodash 
`export const isNotStringType = _.negate(isStringType)`

lib/src/neu/types.ts
```diff
@@ -302,6 +338,10 @@ export function isIntLiteralType(type: Type): type is IntLiteralType {
   return type.kind === TypeKind.INT_LITERAL;
 }
 
+export function areIntLiteralTypes(types: Type[]): types is IntLiteralType[] {
+  return types.every(isIntLiteralType);
```
medric: Same suggestion here - the logic is duplicated. You could encapsulate this logic and pass the predicate (i.e. `isIntLiteralType`) as a parameter. 
`areLiteralTypes = (types, predicate) => types.every(predicate)`

lib/src/neu/generators/openapi3/openapi3.ts
```diff
@@ -0,0 +1,314 @@
...
+  };
+}
+
+function endpointRequestToParameterObjects(
+  request: Request,
+  typeTable: TypeTable,
+  config: Config
+): ParameterObject[] {
+  const pathParameters: PathParameterObject[] = request.pathParams.map(p => {
+    return {
```
medric: nit - use implicit return, `map(p => ({ name: p.name, .... }))`

committedteadrinker (approved):

lib/src/neu/generators/openapi3/openapi3-type-util.ts
lfportal: Where possible throughtout the OpenAPI spec, I've avoided including any details are aren't required. It would be preferable if this maintained that behaviour.

lib/src/neu/generators/openapi3/openapi3-type-util.ts
lfportal: as above - Where possible throughtout the OpenAPI spec, I've avoided including any details are aren't required. It would be preferable if this maintained that behaviour.

medric (approved):

PR #663, opened 2020-01-19
Array type parsing fix

lfportal:
Type aliases referencing basic arrays (non-Array constructor) are incorrectly treated as Array constructor types, which results in a parser error as Array constructor types expect type arguments (basic arrays do not). This PR fixes this behaviour by checking type arguments `.length` to determine if an array is an Array constructor. The TypeScript compiler forbids Array constructors without type arguments, so this can be safely assumed.


ZakHam (approved):
LGTM

PR #664, opened 2020-01-20
Moved json-schema generator to its own directory

lfportal:
This change moves the json-schema generator into its own directory to maintain consistency with the other generators (OpenAPI 2 & 3). Testing has also been given an upgrade - type testing is now consistent in coverage with OpenAPI type testing. Along with this comes a couple of fixes regarding the use of `const` and `enum` in json-schema.


lib/src/neu/generators/json-schema/json-schema-type-util.ts
```diff
@@ -0,0 +1,200 @@
...
+ * be used to configure whether additional properties should be allowed on
+ * object types. This can be useful for data validation purposes where
+ * property validation can be strict or lenient.
+ *
+ * @param type a contract type
+ * @param objectAdditionalProperties whether to allow additional properties for objects
+ */
+export function typeToJsonSchemaType(
+  type: Type,
+  objectAdditionalProperties: boolean = true // TODO: expose proper configuration object
```
medric, line 32: Did you create a ticket for it ? Or you can do it now if it doesn't too much time.

lfportal: No I haven't, this was existing code. There's no real reason immediately to expose a configuration object until there's more than one config option. It serves as a reminder for when someone decides to add more things.

lib/src/neu/generators/json-schema/json-schema-type-util.ts
```diff
@@ -0,0 +1,200 @@
...
+            break;
+          default:
+            oneOfElements.push({
+              type: "boolean",
+              enum: booleanLiterals.map(b => b.value)
+            });
+        }
+
+        const stringLiterals = elements.filter(isStringLiteralType);
+        switch (stringLiterals.length) {
```
medric: Looks like this logic is repeated across this function, I'd consider refactoring it - e.g. create some function that returns an array of `oneOfElements`.

medric (approved):

medric:
LGTM - just address the linting errors :)

PR #677, opened 2020-01-23
CI cleanup

lfportal:
This PR simplifies the CircleCI config by taking advantage of v2.1 features:
- job configuration is replaced by `executors`
- checkout, dependency restoration, dependency installation and dependency caching are common steps to each job and have been moved to a reusable `command` - `prepare-repository`
- the `npm publish` step should only be run once (in case of parallelism) and also should not be run when `ssh`ing for debugging - this is now prevented by conversion to a `deploy` step rather than `run` step
- the `validate-contracts` job has been removed from CI - the validation is now performed directly as part of the test suite
- the node image has been updated to Node v13
  - previous tests on CI produced consistent memory issues running on Node v10
  - Node v13 uses a different memory management strategy, which reduces these errors
- parallelism has been introduced to the jest tests
  - this brings the CI time down from *__~10min__* to *__~90sec__*


package.json
```diff
@@ -60,7 +61,7 @@
     "webpack-cli": "^3.3.10"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=12.0.0"
```
committedteadrinker, line 64: is this supposed to be 13?

.circleci/config.yml
```diff
@@ -1,111 +1,85 @@
...
-    <<: *job_configuration
-    resource_class: xlarge
+
+orbs:
+  node: circleci/node@1.1.6
+
+executors:
+  node-executor:
+    docker:
+      - image: circleci/node:13-stretch
```
committedteadrinker: should it be using `buster`? Super nit, was looking into this and the stable version currently is `buster`

committedteadrinker (approved):

committedteadrinker:
Makes me wonder whether this can be done more easily on github actions

lfportal:
> Makes me wonder whether this can be done more easily on github actions

More or less the same with different syntax, it would be easier to use GitHub actions if we wanted to move to GitHub Package Registry though

package.json
lfportal: Local development would easily fail to run all the tests without memory failures on Node v10. Node v12's improved memory management allowed tests to run consistently without memory issues. This setting ensures developers to use at least v12 in order to develop effectively. We use Node v13 on CI to take advantage of any more recent Node performance/memory improvements.

ZakHam (approved):
Nice work that's a crazy speedup

.circleci/config.yml
```diff
@@ -1,111 +1,85 @@
...
+          timings-type: classname
+          steps:
+            - run:
+                name: Run tests
+                command: yarn ci:test $TESTFILES
+                environment:
+                  JEST_JUNIT_OUTPUT_DIR: ./test-reports/jest
+                  JEST_JUNIT_OUTPUT_NAME: results.xml
+                  JEST_JUNIT_CLASSNAME: "{filepath}"
+      - store_test_results:
```
driesg, line 47: if you want to store the artifact and get more detailed info on results, you can also use `store_artifacts`

.circleci/config.yml
```diff
@@ -1,111 +1,85 @@
...
-
-      - run: ./bin/run --help
-
-  lint-check:
-    <<: *job_configuration
+jobs:
+  test:
+    executor: node-executor
+    resource_class: xlarge
+    parallelism: 8
```
driesg, line 33: 🎉

driesg (commented):
awesome work

driesg:
@lfportal just fyi, Node 13 will be EOL in June this year. So we should update it to a LTS version. Given the performance (js engine) improvements, we can wait for v14

PR #744, opened 2020-03-23
Replace tslint with eslint

lfportal:
[tslint](https://github.com/palantir/tslint) is deprecated and the recommendation is to move to [eslint](https://eslint.org/). This PR removes tslint and adds eslint. Some recommended linting changes have also been applied.


lib/src/generators/openapi2/openapi2-specification.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
fwouts: There are not too many `any`s, maybe it'd be better to have per-line disabling to prevent more from sneaking in?

lib/src/generators/openapi3/openapi3-specification.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
fwouts: Same here.

fwouts (approved):

PR #768, opened 2020-03-30
Add support for warnings and convert no-omittable-fields-within-response-bodies to warning

clintonfeng-airtasker:

The `no-omittable-fields-within-response-bodies` rule prevents us from adding additional fields to our endpoints in production, as to add a field we would have to make it optional first and the `no-omitable-fields-within-response-bodies` prevents us from doing so.

- What does it do?
Adds support for warnings and added an hardcoded config object (which can be refactored to be read from a file in the future).
Set `no-omittable-fields-within-response-bodies` to be an warning by default.

## Checklist:


lib/src/linting/handle-lint-violations.ts
```diff
@@ -0,0 +1,69 @@
+import { GroupedLintRuleViolations } from "./rule";
+import { SpotConfig } from "cli/src/commands/lint";
+
+/**
```
fwouts: This JSDoc looks like it was meant to be on `handleLintViolations()`, not on the `HandleLintViolationsDependencies` interface?

lib/src/linting/handle-lint-violations.ts
```diff
@@ -0,0 +1,69 @@
...
+    msg: string,
+    options: {
+      code?: string;
+      exit: false;
+    }
+  ) => void;
+  warn: (msg: string) => void;
+}
+
+export const handleLintViolations = (
```
fwouts: How about renaming this to `findLintViolations`?

lib/src/linting/handle-lint-violations.ts
```diff
@@ -0,0 +1,69 @@
...
+  groupedLintErrors: GroupedLintRuleViolations[],
+  spotConfig: SpotConfig,
+  { error, warn }: HandleLintViolationsDependencies
+): boolean => {
```
fwouts: Instead of returning a `boolean`, which means you always need to read the JSDoc to understand what it will represent, how about returning a more explicit object? I'm thinking `{ foundErrors: boolean }` for example. You could actually make this a bit more general such as `{ errorCount: number, warningCount: number }`.

This also means that the responsibility of deciding whether to exit or not moves onto the caller.

lib/src/linting/rules.ts
```diff
@@ -16,7 +16,7 @@ export const availableRules: LintingRules = {
   "no-inline-objects-within-unions": noInlineObjectsWithinUnions,
   "no-nullable-arrays": noNullableArrays,
   "no-nullable-fields-within-request-bodies": noNullableFieldsWithinRequestBodies,
-  "no-omiitable-fields-within-response-bodies": noOmittableFieldsWithinResponseBodies
+  "no-omitable-fields-within-response-bodies": noOmittableFieldsWithinResponseBodies
```
fwouts: It should actually be `omittable` to be consistent with the variable name 😝

lib/src/linting/handle-lint-violations.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+        rules: {}
+      };
+
+      deferExit = handleLintViolations(groupedLintErrors, spotConfig, {
+        error: errorMock,
+        warn: warnMock
+      });
+    });
+
+    it("should return deferExit true", () => {
```
fwouts: Personally I'd group these three together and include the setup code inside the test, renaming the overall thing to `it("triggers an error when a rule is not specified")`.

Something like:
```ts
beforeEach(() => {
  errorMock = jest.fn();
  warnMock = jest.fn();
});

it("triggers an error when a rule is not specified", () => {
  ...
  expect(deferExit).toBe(true);
  expect(errorMock).toBeCalled();
  expect(warnMock).not.toBeCalled();
});

..
```

Up to you though, we're in the realm of personal preference.

lib/src/linting/handle-lint-violations.spec.ts
```diff
@@ -0,0 +1,220 @@
+import { handleLintViolations } from "./handle-lint-violations";
+
+describe("handle lint violations", () => {
+  let deferExit: boolean;
+  let errorMock: jest.Mock;
+  let warnMock: jest.Mock;
+
+  describe("lint violation when rule not specified", () => {
+    beforeEach(() => {
+      errorMock = jest.fn();
```
fwouts: This is done in each `describe()` block, I think it would be better to extract it out into a top-level `beforeEach()`?

fwouts (changes requested):
Thank you!

cli/src/commands/lint.ts
```diff
@@ -1,9 +1,21 @@
 import { Command, flags } from "@oclif/command";
 import { lint } from "../../../lib/src/linting/linter";
 import { parse } from "../../../lib/src/parser";
+import { handleLintViolations } from "lib/src/linting/handle-lint-violations";
 
 const ARG_API = "spot_contract";
 
+export interface SpotConfig {
```
lfportal: nit: can we rename this to `LintConfig` for clarity?

lib/src/linting/handle-lint-violations.ts
```diff
@@ -0,0 +1,69 @@
...
+  let deferExit = false;
+
+  groupedLintErrors.forEach(lintingErrors => {
+    let ruleSetting = spotConfig["rules"][lintingErrors.name];
+    if (ruleSetting === undefined) {
+      ruleSetting = "error";
+    }
```
lfportal:
```suggestion
    const ruleSetting = spotConfig["rules"][lintingErrors.name] ?? "error";
```

lfportal (changes requested):
Thanks for doing this!

lib/src/linting/handle-lint-violations.ts
clintonfeng-airtasker: Yeah, I like your suggestion of `{ errorCount: number, warningCount: number }`.
Could potentially add something like  `max-warnings` in the future with that

lfportal (approved):
lgtm

lib/src/linting/find-lint-violations.ts
```diff
@@ -0,0 +1,73 @@
+import { GroupedLintRuleViolations } from "./rule";
+import { LintConfig } from "cli/src/commands/lint";
+
+interface FindLintViolationsDependencies {
+  error: (
+    msg: string,
+    options: {
```
fwouts: We never set `code` and we never set `exit: true` here. Could we remove `options` altogether, and instead do the following in `commands/lint.ts`:
```
    const { errorCount } = findLintViolations(groupedLintErrors, lintConfig, {
      error: (msg) => this.error(msg, { exit: false }),
      warn: this.warn
    });
```

fwouts (approved):

lib/src/linting/find-lint-violations.ts
clintonfeng-airtasker: 👍 I ended up removing options from the interface. Looking at the oclif readme, it looks like we have to specify `{exit: false }` for it to not exit but we can just pass in:
```
(msg: string) => {
        this.error(msg, { exit: false });
},
```

fwouts (approved):

Issue #652, opened 2020-01-15
no ability to add version to doc


acordiner92:
### Current behavior:
Currently there is no ability to be able to add a version to the doc (aka setting the version in the info section of open api spec).
### Desired behavior:
To have the option to be able to set the version.
A potential solution maybe in the @api component have a info object which has a version field that can be set e.g
```
@api({
  name: "My API",
  info: {
     version: 0.2.0
  }
})
```
### Steps to reproduce:
N/A

committedteadrinker:
Hi @acordiner92,

The concept of `info` is more aligned to what OpenAPI has. Would suggest to support this syntax instead:
```
@api({
  name: "My API",
  version: 0.2.0
})
```

Please create a PR for this as well for further discussion on the implementation side. Thanks!

committedteadrinker:
@theRingleman @acordiner92 I'm closing this as it has been addressed on the PR and released on version 1.1.3.

PR #892, opened 2020-06-19
Adds the abailty to specify api versions

theRingleman:

This addresses this issue: https://github.com/airtasker/spot/issues/652
As a developer using the spot system to create my api documentation, I require the need to version my docs. This allows me to trigger events in my pipelines and keep everyone in our organization consistently up to date.
- What does it do?
This provides the capability to version the API documentation from the spot doc itself, rather than manually through the generated contract which will get overwritten on the next document change.

## Checklist:

- [ x ] I've added/updated tests to cover my changes
- [ x ] I've created an issue associated with this PR
I have not created an issue due to the fact that one was already created.

### DIFF
--- a/lib/src/parsers/contract-parser.ts
+++ b/lib/src/parsers/contract-parser.ts
@@ -61,11 +61,10 @@ export function parseContract(
   const description = descriptionDoc?.getDescription().trim();
 
   // Handle Version
-  let version = "0.0.0";
   const versionProp = getObjLiteralProp<ApiConfig>(decoratorConfig, "version");
-  if (versionProp) {
-    version = getPropValueAsStringOrThrow(versionProp)?.getLiteralText().trim();
-  }
+  const version = versionProp
+    ? getPropValueAsStringOrThrow(versionProp).getLiteralText().trim()
+    : undefined;
 
   // Handle config
   const configResult = resolveConfig(klass);
--- a/lib/src/parsers/contract-parser.ts
+++ b/lib/src/parsers/contract-parser.ts
@@ -10,7 +10,7 @@ import { parseEndpoint } from "./endpoint-parser";
 import {
   getClassWithDecoratorOrThrow,
   getDecoratorConfigOrThrow,
-  getJsDoc,
+  getJsDoc, getObjLiteralProp,
   getObjLiteralPropOrThrow,
   getPropertyWithDecorator,
   getPropValueAsStringOrThrow,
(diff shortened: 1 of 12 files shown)


lib/src/parsers/contract-parser.ts
```diff
@@ -59,6 +60,13 @@ export function parseContract(
   const descriptionDoc = getJsDoc(klass);
   const description = descriptionDoc?.getDescription().trim();
 
+  // Handle Version
+  let version = "0.0.0";
+  const versionProp = getObjLiteralProp<ApiConfig>(decoratorConfig, "version");
+  if (versionProp) {
+    version = getPropValueAsStringOrThrow(versionProp)?.getLiteralText().trim();
+  }
```
lfportal: The contract definition specifies `version` as optional. In general we prefer letting the contract generators (OpenAPI 2 & 3) define the default values, rather than driving it with Spot's internal definitions. Something along the lines of:
```suggestion
  const versionProp = getObjLiteralProp<ApiConfig>(decoratorConfig, "version");
  const version = versionProp ? getPropValueAsStringOrThrow(versionProp).getLiteralText().trim() : undefined
```

lib/src/syntax/api.ts
```diff
@@ -15,4 +15,5 @@ export function api(config: ApiConfig) {
 export interface ApiConfig {
   /** Name of the API. This should be the name of the service that is being documented */
   name: string;
+  version?: string;
```
lfportal, line 19:
```suggestion
  /** Version of this document */
  version?: string;
```

lfportal (changes requested):
Welcome and thanks for the contribution @theRingleman! Just a couple of suggestions.

lib/src/parsers/contract-parser.ts
theRingleman: Excellent! Makes perfect sense to me. I will get that updated in the pull request as well as update the contract spec to reflect those changes. Thanks for the suggestion, and the rapid response!

lib/src/syntax/api.ts
theRingleman: Ugh.... Sorry about that, I always forget docs.

lfportal (approved):

committedteadrinker (approved):

committedteadrinker:
Thank you for the contribution @theRingleman I'll release the new version that contains this.

Issue #711, opened 2020-02-20
Missing meta information


hffmnn:
**Is your feature request related to a problem? Please describe.**

I am trying to convert an existing OpenApi file to `spot`. When trying to do that I am missing most of the meta-information parameters. For example:

* `description` for parameters and queries
* `summary` for the endpoint
* `example` for parameters and endpoints
*  other content types than `application/json` e.g. `application/xml` for responses
* `description` for `@interface`s

**Describe the solution you'd like**

Add these properties to the corresponding annotations (or in the case of annotations for `interface` types add new annotations.

**Describe alternatives you've considered**

Another option would be to add something like a global overwrite, that could add these properties (I guess by giving up the type-safety of TS).

As an example:

```js
@api({
    name: "API",
})

@globalUnsafeOverrides({
  "openapi3": {
    "info": {
      "title": "Hello API"
    }
  }
})
```

would give this openapi3 doc:

```json
{
  "openapi": "3.0.2",
  "info": {
    "title": "API",
    "version": "0.0.0"
  }
```

via sth. like that:

```js
const finalJson = {
  ...api,
  ...globalUnsaveOverrides
}
```

 Or the other way around (which would preserve the properties set in the annotated version)

```js
const finalJson = {
  ...globalUnsaveOverrides,
  ...api
}
```

---

As an alternative for `description` for `interface` and its properties, I wonder if it is possible (and a good idea) to use e.g. surrounding comments.

```js
// The user type
interface User {
  // the unique id
  id: string
}
```

would give

```json
"components": {
    "schemas": {
      "User": {
        "type": "object",
        "description": "The user type",
        "properties": {
          "id": {
            "type": "string"
          },
         "description": "the unique id"
        },
        "required": [
          "id"
        ]
      }
    }
  }
```

EDIT: I found https://github.com/airtasker/spot/commit/aa517d3681bb8f6c3357390f3ff8933fcf6ee23c for the `@endpoint` annotation and wonder why it got removed?

jimpudar:
Agreed, I'm definitely missing `description`, `summary`, and `example`!

lfportal:
[...]

PR #916, opened 2020-07-10
add example to pathParams

maxzirps:

I started with an example tag for pathParms as stated in #711 

I looked at the suggestion in https://github.com/airtasker/spot/issues/711#issuecomment-604226927 : 

```javascript
request(
  @pathParams
  pathParams:  {
     /**
      * An identifier
      *
      * @example <caption>Example 32</caption>
      * 32
      *
      * @example <caption>Example 64</caption>
      * 64
      */
     id: Int32
  }
  // ...
) {}
```

Currently it only accepts one example of type string, still need to improve this. Just created this draft to know if this is the way to go to contribute, since I didn't find a CONTRIBUTE.md or similar.

## Checklist:

### DIFF
--- a/lib/src/definitions.ts
+++ b/lib/src/definitions.ts
@@ -63,7 +63,12 @@ export interface PathParam {
   name: string;
   description?: string;
   type: Type;
-  examples?: Record<string, string>;
+  examples?: Example[];
+}
+
+export interface Example {
+  name: string;
+  value: string;
 }
 
 export interface QueryParam {
--- a/lib/src/definitions.ts
+++ b/lib/src/definitions.ts
@@ -63,7 +63,7 @@ export interface PathParam {
   name: string;
   description?: string;
   type: Type;
-  example?: string;
+  examples?: Record<string, string>;
 }
 
 export interface QueryParam {
--- a/lib/src/definitions.ts
+++ b/lib/src/definitions.ts
@@ -68,7 +68,7 @@ export interface PathParam {
 
 export interface Example {
   name: string;
-  value: string;
+  value: string; // TODO: encapsulate type information
 }
 
 export interface QueryParam {
--- a/lib/src/definitions.ts
+++ b/lib/src/definitions.ts
@@ -63,6 +63,7 @@ export interface PathParam {
   name: string;
   description?: string;
   type: Type;
+  example?: string;
 }
 
 export interface QueryParam {
--- a/lib/src/parsers/__spec-examples__/path-params.ts
+++ b/lib/src/parsers/__spec-examples__/path-params.ts
@@ -13,7 +13,7 @@ class PathParamsClass {
       arrayProperty: string[];
       /** property-example description
        * @example property-example
-       * property-example-value
+       * "property-example-value"
        *  */
       "property-with-example": string;
       /** property-two-examples description
@@ -71,6 +71,13 @@ class PathParamsClass {
     @pathParams
     pathParamsWithIllegalPropertyArrayType: {
       property: null[];
+    },
+    @pathParams
+    paramsWithUnquotedStringExample: {
+      /**@example
+       * invalid-property-example
+       */
+      property: string;
     },
     @pathParams
     optionalPathParams?: {
--- a/lib/src/parsers/__spec-examples__/path-params.ts
+++ b/lib/src/parsers/__spec-examples__/path-params.ts
@@ -19,9 +19,9 @@ class PathParamsClass {
       "property-with-example": string;
       /** property-two-examples description
        * @example property-example-one
-       * property-example-one-value
+       * 123
        * @example property-example-two
-       * property-example-two-value
+       * 456
        * */
       "property-with-examples": Integer;
     },
--- a/lib/src/parsers/__spec-examples__/path-params.ts
+++ b/lib/src/parsers/__spec-examples__/path-params.ts
@@ -11,6 +11,14 @@ class PathParamsClass {
       /** property description */
       "property-with-description": string;
       arrayProperty: string[];
+      /** property description
+       * @example property-example */
+      "property-with-example": string;
+    },
+    @pathParams
+    paramsWithEmptyExample: {
+      /**@example */
+      property: string;
     },
     @pathParams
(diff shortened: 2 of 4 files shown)


lfportal:
Hi @zirmax, sorry for the delay. I'm looking to set out contribution guidelines soon. You're PR is on the right track 👍

zirmax:
I have difficulties writing a guard for the examples types, so they match the types of the param.
I always get `Enums are not supported` error.

https://github.com/zirmax/spot/blob/5167bc15790c50e354f9079b00e5c919f312c618/lib/src/parsers/__spec-examples__/path-params.ts#L19-L26

https://github.com/zirmax/spot/blob/5167bc15790c50e354f9079b00e5c919f312c618/lib/src/parsers/path-params-parser.spec.ts#L63-L74


@lfportal do you have an idea which types I can use for the guard?

mahirk:
@zirmax (cc: @lfportal) would have an intention for also extending this to the request and response body for examples? I don't want to reinvent the wheel, so I was waiting for this to be complete to add those in (unless it already would be added in)

maxzirps:
> @zirmax (cc: @lfportal) would have an intention for also extending this to the request and response body for examples? I don't want to reinvent the wheel, so I was waiting for this to be complete to add those in (unless it already would be added in)

I will only add it to the path params; I am nearly done, just need to add some checks so that the example and specified type match

maxzirps:
@lfportal I am having issues getting the type of the examples, so I can compare them to the type specified:

https://github.com/airtasker/spot/blob/c60a38181ba4009b522331257505990d30df45e5/lib/src/parsers/path-params-parser.ts#L144-L163

Do you have any suggestions?

lfportal:
@maxzirps `JSDocTag` does not expose type information from the tag's comment so we'll be unable to use the existing `type-parser.ts` methods. I think our best bet here is to rely on `JSON.parse` and `typeof`:
```js
const tagType = typeof JSON.parse(tag.getComment())
```
and `switch` on the result - perhaps we can introduce a `jsdoc-example-tag-parser.ts` for this similar to `type-parser.ts`. We can keep it simple for now and support just the basic types required for path params.

maxzirps:
> @maxzirps `JSDocTag` does not expose type information from the tag's comment so we'll be unable to use the existing `type-parser.ts` methods. I think our best bet here is to rely on `JSON.parse` and `typeof`:
> 
> ```js
> const tagType = typeof JSON.parse(tag.getComment())
> ```
> 
> and `switch` on the result - perhaps we can introduce a `jsdoc-example-tag-parser.ts` for this similar to `type-parser.ts`. We can keep it simple for now and support just the basic types required for path params.

I added some checks for basic types. Propose to keep it in this file until it is needed somewhere else

lib/src/definitions.ts
```diff
@@ -63,6 +63,7 @@ export interface PathParam {
   name: string;
   description?: string;
   type: Type;
+  examples?: Record<string, string>;
```
lfportal:
```suggestion
  examples?: Example[];
```

The convention we currently use is a sorted array. There is consideration being put into changing this convention to use `Record`s. For now, use arrays for consistency.

lib/src/parsers/__spec-examples__/path-params.ts
```diff
@@ -11,6 +12,44 @@ class PathParamsClass {
...
+      property: string;
+    },
+    @pathParams
+    paramsWithDuplicateExampleName: {
+      /**@example name
+       * 123
+       * @example name
+       * 456
+       */
+      property: string;
```
lfportal:
```suggestion
      /**@example name
       * "123"
       * @example name
       * "456"
       */
      property: string;
```

To ensure forwards compatibility with object based examples, strings should be explicit

maxzirps: What if someone wants to specify a number example?

lfportal: Number examples for `string` types would be numbers in quotes i.e. `"123"`. Number examples for `number` types will be the number without quotes i.e. `123`. Some further examples:
```ts
/**
 * @example string-example
 * "myString"
 *
 * @example string-number-example
 * "123"
 */
stringProp: String;

/**
 * @example number-example
 * 123
 */
intProp: Integer;

/**
 * @example object-example
 * {
 *   "stringAttr": "myString",
 *   "intAttr": 123
 * }
 */
objectProp: {
  stringAttr: String;
  intAttr: Integer;
}
```

maxzirps: ah you're right, mistook it for the int example

lib/src/definitions.ts
```diff
@@ -63,6 +63,12 @@ export interface PathParam {
   name: string;
   description?: string;
   type: Type;
+  examples?: Example[];
+}
+
+export interface Example {
+  name: string;
+  value: string;
```
lfportal: OpenAPI examples will be output as strings, however within SPOT it will be useful to retain type information. For now we can leave this as `string` and leave a `TODO`.
```suggestion
  value: string; // TODO: encapsulate type information
```

lfportal (changes requested):
I think we're almost there, just some cleanup and a string example fix

maxzirps:
I am using `JSON.parse` now and enforce the quotes. 
Are the examples automatically written to the result or did I forget to add this somewhere?

lfportal:
> I am using `JSON.parse` now and enforce the quotes.
> Are the examples automatically written to the result or did I forget to add this somewhere?

The contract parsing phase will create the `Contract` object which is passed to one of the generators, which performs the conversion between SPOT's native contract representation to some standard one. For example [the openapi3 generator](https://github.com/airtasker/spot/blob/master/lib/src/generators/openapi3/openapi3.ts). Feel free to add the `example` conversion for the openapi generators in this PR. Otherwise we can do that in a separate PR.

@maxzirps let me know if you want to add that in, otherwise we can merge this PR as is.

lfportal (approved):

maxzirps:
I would do it in a different PR and merge this one, since this PR has been open for some months

PR #1076, opened 2020-10-16
Generate examples from contract for OpenAPI Spec

### LINKED ISSUE #1051: Add conversion of examples
Create the conversion of pathParams examples between SPOTs native contract representation to some standard one

> I am using `JSON.parse` now and enforce the quotes.
> Are the examples automatically written to the result or did I forget to add this somewhere?

The contract parsing phase will create the `Contract` object which is passed to one of the generators, which performs the conversion between SPOT's native contract representation to some standard one. For example [the openapi3 generator](https://github.com/airtasker/spot/blob/master/lib/src/generators/openapi3/openapi3.ts). Feel free to add the `example` conversion for the openapi generators in this PR. Otherwise we can do that in a separate PR.

@maxzirps let me know if you want to add that in, otherwise we can merge this PR as is.

_Originally posted by @lfportal in https://github.com/airtasker/spot/pull/916#issuecomment-705400712_

mahirk:
Resolves https://github.com/airtasker/spot/issues/1051

## Checklist:

### DIFF
--- a/lib/src/generators/openapi3/openapi3.ts
+++ b/lib/src/generators/openapi3/openapi3.ts
@@ -4,7 +4,7 @@ import {
   Config,
   Contract,
   DefaultResponse,
-  Endpoint,
+  Endpoint, Example,
   Header,
   HttpMethod,
   isSpecificResponse,
@@ -33,7 +33,7 @@ import {
   QueryParameterObject,
   RequestBodyObject,
   ResponseObject,
-  ResponsesObject
+  ResponsesObject, ExamplesSet
 } from "./openapi3-specification";
 import {
   typeToSchemaOrReferenceObject,
@@ -153,7 +153,8 @@ function endpointRequestToParameterObjects(
     in: "path",
     description: p.description,
     required: true,
-    schema: typeToSchemaOrReferenceObject(p.type, typeTable)
+    schema: typeToSchemaOrReferenceObject(p.type, typeTable),
+    examples: exampleToOpenApiExampleModel(p.examples),
   }));
 
   const queryParameters: QueryParameterObject[] = request.queryParams.map(
@@ -163,7 +164,8 @@ function endpointRequestToParameterObjects(
       description: p.description,
       ...typeToQueryParameterSerializationStrategy(p.type, typeTable, config),
       required: !p.optional,
-      schema: typeToSchemaOrReferenceObject(p.type, typeTable)
+      schema: typeToSchemaOrReferenceObject(p.type, typeTable),
+      examples: exampleToOpenApiExampleModel(p.examples),
     })
   );
 
@@ -172,7 +174,8 @@ function endpointRequestToParameterObjects(
     in: "header",
     description: p.description,
     required: !p.optional,
-    schema: typeToSchemaOrReferenceObject(p.type, typeTable)
+    schema: typeToSchemaOrReferenceObject(p.type, typeTable),
+    examples: exampleToOpenApiExampleModel(p.examples),
   }));
 
   const parameters: ParameterObject[] = [];
@@ -332,4 +335,15 @@ function httpMethodToPathItemMethod(
     default:
       assertNever(method);
   }
+}
+
+function exampleToOpenApiExampleModel(examples?: Example[]): ExamplesSet | undefined {
+  return examples?.reduce<ExamplesSet>((acc, example: Example) => {
+      acc[example.name] = {
+        value: example.value
+      };
+      return acc;
+    },
+    {}
+  );
 }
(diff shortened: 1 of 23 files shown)


mahirk:
@lfportal Would you be able to take a look at this?

lib/src/parsers/headers-parser.spec.ts
```diff
@@ -45,12 +47,40 @@ describe("headers parser", () => {
...
+        kind: TypeKind.STRING
+      },
+      optional: false
+    });
+    expect(result[4]).toStrictEqual({
+      description: "property-two-examples description",
+      examples: [
+        {
+          name: "property-example-one",
+          value: 123
```
lfportal, line 71: Given we are storing different types for `value`, can we update the corresponding definition to `value: any`:

```ts
// definitions.ts
export interface Example {
  name: string;
  value: any;
}
```

lib/src/parsers/__spec-examples__/examples.ts
```diff
@@ -0,0 +1,37 @@
...
+  property: Integer;
+};
+
+type MismatchedExampleAndStringWithQuotesType = {
+  /**
+   * @example name
+   * This_is_not_a_string_in_quotes
+   */
+  property: string;
+};
```
lfportal: Can we merge these examples into the above `type`? See `example-parser.spec.ts` for related comment.

lib/src/parsers/example-parser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+} from "./parser-helpers";
+import { JSDoc, TypeAliasDeclaration } from "ts-morph";
+import { StringType, Type, TypeKind } from "../types";
+
+type JSDocNodeMapping = {
+  ParsedExamples: JSDoc[];
+  MismatchedExampleAndIntegerType: JSDoc[];
+  MismatchedExampleAndStringWithQuotesType: JSDoc[];
+  [typeName: string]: JSDoc[];
+};
```
lfportal: I don't see much of an advantage in holding on to an object of this type. Each `test` can just deal with one `property` of the `type`. See related comment from `__spec-examples__/examples.ts`

lib/src/parsers/example-parser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+describe("example-parser", () => {
+  describe("extractJSDocExamples", () => {
+    const jsDocs: JSDocNodeMapping = {} as JSDocNodeMapping;
+    const sourceFile = createProjectFromExistingSourceFile(
+      `${__dirname}/__spec-examples__/examples.ts`
+    ).file;
+    const typeValues = sourceFile.getTypeAliases();
+
+    typeValues.forEach((typeAlias: TypeAliasDeclaration) => {
+      const properties = parseTypeReferencePropertySignaturesOrThrow(
```
lfportal: Continuing on from above comments, I think its clearer and more consistent with other tests if each test relies on this `properties` array. i.e. at the beginning of each `test` we do something along the lines of:

```
properties.find(p => p.getName() === "property-with-example").getJsDocs()
```
We can extract this into a function:
```
function getJsDocsFromProperty(properties: PropertySignature, propertyName: string): JsDoc[] {
 ...
}
```
I think this allows additional tests / examples easier to add and is simpler to reason with.

lib/src/parsers/example-parser.ts
```diff
@@ -0,0 +1,126 @@
+import { JSDoc } from "ts-morph";
+import { Type, TypeKind } from "../types";
+import { err, ok, Result } from "../util";
+import { Example } from "../definitions";
+import { ParserError } from "../errors";
+
+export function extractJSDocExamples(
```
lfportal, line 7: Awesome! thanks for consolidating this 👍

lib/src/generators/openapi3/openapi3.ts
```diff
@@ -333,3 +338,14 @@ function httpMethodToPathItemMethod(
...
+
+function exampleToOpenApiExampleModel(
+  examples?: Example[]
+): ExamplesSet | undefined {
+  return examples?.reduce<ExamplesSet>((acc, example: Example) => {
+    acc[example.name] = {
+      value: example.value
+    };
+    return acc;
+  }, {});
```
lfportal, line 350: We should add a test to cover a contract that generates an example in `openapi3.spec.ts`

lfportal (changes requested):
Thanks @mahirk! Left some comments, mostly to do with some tests.

mahirk:
Thanks for all the feedback @lfportal pushed a few updates!

lib/src/parsers/headers-parser.spec.ts
mahirk: Should this be either a generic or `string | integer | boolean`?

lfportal: I think this can be left as is until we need to encapsulate type properly. At that point I think we can replace it with some subset of `Type` or `TypeDef`.

lfportal (approved):
👍

PR #1133, opened 2020-11-17
Allow usage of Intersection Types for spec generation

### LINKED ISSUE #1123: Allow usage for Intersection Types
**Is your feature request related to a problem? Please describe.**
An intersection type is the combination of two or more types to combine multiple types into one. These are currently not supported by the parser though are logically no different than types. 

**Describe the solution you'd like**
The following spec fails:
```
class Api {}

@endpoint({
  method: "POST",
  path: "/users"
})
class CreateUser {
  @request
  request(
    @headers
    headers: HeaderAlias,
    @queryParams
    queryParams: LimitParams & OffsetParam,
    @body body: CreateUserRequest
  ) {}

  @response({ status: 201 })
  successResponse(@body body: CreateUserResponse) {}
}
interface LimitParams {
  limit?: number
}

interface OffsetParam {
  offset: string
}

interface CreateUserResponse {
  firstName: string;
  lastName: string;
  role: string;
}
```

with the error as `Error: unknown type`. 

The proposal would be to update the type-parser to allow intersection types and define them using the guidelines provided above.

mahirk:
Resolves: https://github.com/airtasker/spot/issues/1123

This PR allows for the usage of `intersection` types in contract definitions

Sample Contract:
```
import {
  api,
  endpoint,
  request,
  queryParams,
  body,
  response
} from "../lib/src/syntax";

@api({
  name: "My API"
})
class Api {}

@endpoint({
  method: "POST",
  path: "/users"
})
class CreateUser {
  @request
  request(
    @queryParams
    queryParams: LimitParams & OffsetParam,
    @body body: CreateUserResponse
  ) {}

  @response({ status: 201 })
  successResponse(@body body: bodyType) {}
}

type bodyType = CreateUserResponse & OffsetParam;

interface LimitParams {
  limit?: number;
}

interface OffsetParam {
  offset: string;
}

interface CreateUserResponse {
  firstName: string;
  lastName: string;
  role: string;
}
```

Sample Spec:
```
openapi: 3.0.2
info:
  title: My API
  version: 0.0.0
paths:
  /users:
    post:
      operationId: CreateUser
      parameters:
        - name: limit
          in: query
          required: false
          schema:
            type: number
            format: float
        - name: offset
          in: query
          required: true
          schema:
            type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/CreateUserResponse'
        required: true
      responses:
        '201':
          description: 201 response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/bodyType'
components:
  schemas:
    CreateUserResponse:
      type: object
      properties:
        firstName:
          type: string
        lastName:
          type: string
        role:
          type: string
      required:
        - firstName
        - lastName
        - role
    OffsetParam:
      type: object
      properties:
        offset:
          type: string
      required:
        - offset
    bodyType:
      allOf:
        - $ref: '#/components/schemas/CreateUserResponse'
        - $ref: '#/components/schemas/OffsetParam'
```

## Checklist:

### DIFF
--- a/lib/src/generators/openapi2/openapi2-type-util.ts
+++ b/lib/src/generators/openapi2/openapi2-type-util.ts
@@ -5,6 +5,7 @@ import {
   areIntLiteralTypes,
   areStringLiteralTypes,
   ArrayType,
+  IntersectionType,
   isNotNullType,
   isNullType,
   ObjectType,
@@ -65,6 +66,8 @@ export function typeToSchemaObject(
       return arrayTypeToSchema(type, typeTable, nullable);
     case TypeKind.UNION:
       return unionTypeToSchema(type, typeTable);
+    case TypeKind.INTERSECTION:
+      return intersectionTypeToSchema(type, typeTable);
     case TypeKind.REFERENCE:
       return referenceTypeToSchema(type, nullable);
     default:
@@ -224,6 +227,22 @@ function unionTypeToSchema(
         throw new Error(`Unions are not supported in OpenAPI 2`);
       }
   }
+}
+
+function intersectionTypeToSchema(
+  type: IntersectionType,
+  typeTable: TypeTable
+): AllOfSchemaObject {
+  // Sanity check: This should not be possible
+  if (type.types.length === 0) {
+    throw new Error("Unexpected type: intersection type with no types");
+  }
+
+  const nonNullTypes = type.types.filter(isNotNullType);
+
+  return {
+    allOf: nonNullTypes.map((t: Type) => typeToSchemaObject(t, typeTable))
+  };
 }
 
 function referenceTypeToSchema(
--- a/lib/src/http.ts
+++ b/lib/src/http.ts
@@ -28,6 +28,7 @@ export function isPathParamTypeSafe(type: Type, typeTable: TypeTable): boolean {
     case TypeKind.ARRAY:
       return isParamArrayElementTypeSafe(type.elementType, typeTable);
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
       return type.types.every(t => isPathParamTypeSafe(t, typeTable));
     case TypeKind.REFERENCE:
       return isPathParamTypeSafe(dereferenceType(type, typeTable), typeTable);
@@ -69,6 +70,7 @@ export function isQueryParamTypeSafe(
         isParamObjectPropertyTypeSafe(p.type, typeTable)
       );
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
       return type.types.every(t => isQueryParamTypeSafe(t, typeTable));
     case TypeKind.REFERENCE:
       return isQueryParamTypeSafe(dereferenceType(type, typeTable), typeTable);
@@ -103,6 +105,7 @@ export function isHeaderTypeSafe(type: Type, typeTable: TypeTable): boolean {
     case TypeKind.OBJECT:
       return false;
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
       return type.types.every(t => isHeaderTypeSafe(t, typeTable));
     case TypeKind.REFERENCE:
       return isHeaderTypeSafe(dereferenceType(type, typeTable), typeTable);
@@ -140,6 +143,7 @@ function isParamObjectPropertyTypeSafe(
     case TypeKind.OBJECT:
       return false;
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
       return type.types.every(t => isParamObjectPropertyTypeSafe(t, typeTable));
     case TypeKind.REFERENCE:
       return isParamObjectPropertyTypeSafe(
@@ -180,6 +184,7 @@ function isParamArrayElementTypeSafe(
     case TypeKind.OBJECT:
       return false;
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
       return type.types.every(t => isParamArrayElementTypeSafe(t, typeTable));
     case TypeKind.REFERENCE:
       return isParamArrayElementTypeSafe(
--- a/lib/src/linting/rules/has-discriminator.ts
+++ b/lib/src/linting/rules/has-discriminator.ts
@@ -126,7 +126,6 @@ function findDisriminatorViolations(
     case TypeKind.INT_LITERAL:
     case TypeKind.DATE:
     case TypeKind.DATE_TIME:
-    case TypeKind.INTERSECTION:
       return [];
     case TypeKind.OBJECT:
       return type.properties.reduce<string[]>((acc, prop) => {
@@ -144,6 +143,12 @@ function findDisriminatorViolations(
         typeTable,
         typePath.concat("[]")
       );
+    case TypeKind.INTERSECTION:
+      return type.types.reduce<string[]>((acc, t) => {
+        return acc.concat(
+          findDisriminatorViolations(t, typeTable, typePath.concat())
+        );
+      }, []);
     case TypeKind.UNION: {
       const violationsInUnionTypes = type.types.reduce<string[]>((acc, t) => {
         return acc.concat(
--- a/lib/src/linting/rules/has-discriminator.ts
+++ b/lib/src/linting/rules/has-discriminator.ts
@@ -126,6 +126,7 @@ function findDisriminatorViolations(
     case TypeKind.INT_LITERAL:
     case TypeKind.DATE:
     case TypeKind.DATE_TIME:
+    case TypeKind.INTERSECTION:
       return [];
     case TypeKind.OBJECT:
       return type.properties.reduce<string[]>((acc, prop) => {
--- a/lib/src/parsers/type-parser.ts
+++ b/lib/src/parsers/type-parser.ts
@@ -52,7 +52,10 @@ import {
   NullType,
   intersectionType,
   isPrimitiveType,
-  isArrayType, doesInterfaceEvaluatesToNever, possibleRootTypes, isObjectType
+  isArrayType,
+  doesInterfaceEvaluatesToNever,
+  possibleRootTypes,
+  isObjectType
 } from "../types";
 import { err, ok, Result } from "../util";
 import { getJsDoc, getPropertyName } from "./parser-helpers";
--- a/lib/src/parsers/type-parser.ts
+++ b/lib/src/parsers/type-parser.ts
@@ -52,7 +52,7 @@ import {
   NullType,
   intersectionType,
   isPrimitiveType,
-  isArrayType
+  isArrayType, doesInterfaceEvaluatesToNever, possibleRootTypes, isObjectType
 } from "../types";
 import { err, ok, Result } from "../util";
 import { getJsDoc, getPropertyName } from "./parser-helpers";
@@ -482,7 +482,8 @@ function parseIntersectionTypeNode(
     if (typeResult.isErr()) return typeResult;
     // Only allow objects, unions, intersections and references
     const typeResultType = typeResult.unwrap();
-    if (isPrimitiveType(typeResultType) || isArrayType(typeResultType)) {
+    const concreteTypes = possibleRootTypes(typeResultType, typeTable);
+    if (!concreteTypes.every(isObjectType)) {
       return err(
         new TypeNotAllowedError(
           "Cannot use primitive types in an intersection type",
@@ -495,7 +496,18 @@ function parseIntersectionTypeNode(
     }
     types.push(typeResultType);
   }
-  return ok(intersectionType(types, inferDiscriminator(types, typeTable)));
+  if (doesInterfaceEvaluatesToNever(types, typeTable)) {
+    return err(
+      new TypeNotAllowedError(
+        "intersection evaluates to never and is an illegal argument",
+        {
+          file: typeNode.getSourceFile().getFilePath(),
+          position: typeNode.getPos()
+        }
+      )
+    );
+  }
+  return ok(intersectionType(types));
 }
 
 /**
--- a/lib/src/parsers/type-parser.ts
+++ b/lib/src/parsers/type-parser.ts
@@ -50,7 +50,9 @@ import {
   TypeTable,
   unionType,
   NullType,
-  intersectionType
+  intersectionType,
+  isPrimitiveType,
+  isArrayType
 } from "../types";
 import { err, ok, Result } from "../util";
 import { getJsDoc, getPropertyName } from "./parser-helpers";
@@ -462,7 +464,7 @@ function parseUnionType(
 /**
  * Parse an intersection type node.
  *
- * @param typeNode union type node
+ * @param typeNode intersection type node
  * @param typeTable a TypeTable
  * @param lociTable a LociTable
  */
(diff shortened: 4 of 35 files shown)


mahirk:
@lfportal Would you be able to take a look at this?

lib/src/types.ts
```diff
@@ -37,7 +38,8 @@ export type Type =
   | ObjectType
   | ArrayType
   | UnionType
-  | ReferenceType;
+  | ReferenceType
+  | IntersectionType;
```
lfportal, line 42: `IntersectionType` will also need to be excluded in `ConcreteType` below, which will also require a non-trivial change to [`possibleRootTypes`](https://github.com/airtasker/spot/blob/16f5e964dff888e96dfffafeee75617d28602dc7/lib/src/types.ts#L421) - the intersection type will need to be resolved into single merged `object` type

lib/src/parsers/type-parser.ts
```diff
@@ -455,6 +459,30 @@ function parseUnionType(
   }
 }
 
+/**
+ * Parse an intersection type node.
+ *
+ * @param typeNode union type node
```
lfportal:
```suggestion
 * @param typeNode intersection type node
```

lib/src/parsers/type-parser.ts
```diff
@@ -455,6 +459,30 @@ function parseUnionType(
...
+ * @param typeTable a TypeTable
+ * @param lociTable a LociTable
+ */
+function parseIntersectionTypeNode(
```
lfportal, line 472: I'm thinking we should limit intersection type members to objects only. I can't see any use-case for intersection types that involve non-objects (e.g. `string & "a"`). This will make working with intersection types a bit easier.

lib/src/linting/rules/has-discriminator.ts
```diff
@@ -126,6 +126,7 @@ function findDisriminatorViolations(
     case TypeKind.INT_LITERAL:
     case TypeKind.DATE:
     case TypeKind.DATE_TIME:
+    case TypeKind.INTERSECTION:
```
lfportal: Check for discriminator violations in each member of the intersection type

lfportal (commented):
This is great @mahirk ! Couple of questions and suggestions.

lib/src/generators/openapi2/openapi2-type-util.ts
```diff
@@ -226,6 +229,22 @@ function unionTypeToSchema(
   }
 }
 
+function intersectionTypeToSchema(
+  type: IntersectionType,
+  typeTable: TypeTable
+): AllOfSchemaObject {
+  // Sanity check: This should not be possible
+  if (type.types.length === 0) {
```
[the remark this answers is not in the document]
mahirk: I believe it can only be an intersection if it has 2 or more otherwise it would be designated as a `type` or `interface` itself.

lib/src/http.ts
```diff
@@ -28,6 +28,7 @@ export function isPathParamTypeSafe(type: Type, typeTable: TypeTable): boolean {
     case TypeKind.ARRAY:
       return isParamArrayElementTypeSafe(type.elementType, typeTable);
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
```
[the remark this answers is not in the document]
mahirk: Yeah, I actually hit this because we have one of our APIs where we have two params say:
`v1/accounts/:accountId/purchase/:purchaseId`

So it was defined as `AccountId & PurchaseId`

lib/src/types.ts
mahirk: @lfportal would be curious when it would need to resolve into a merged `object`. I might be able to get a better understanding of what would need to be returned

lib/src/http.ts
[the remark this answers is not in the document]
lfportal: I'm not sure I follow, your example would be written similar to:

```ts
@endpoint({
  method: "GET",
  path: "/v1/accounts/:accountId/purchase/:purchaseId"
})
class Endpoint {
  @request
  request(
    @pathParams
    pathParams: AccountId & PurchaseId
  ) {}
}

interface AccountId {
  accountId: String;
}

interface PurchaseId {
  purchaseId: String;
}
```

Neither `accountId` nor `purchaseId` types would require the use of intersection type.

lib/src/types.ts
lfportal: Currently the primary use case (though there are others) for `possibleRootTypes` is to infer a discriminator property for union types see [inferDiscriminator](https://github.com/airtasker/spot/blob/16f5e964dff888e96dfffafeee75617d28602dc7/lib/src/types.ts#L453). 

`possibleRootTypes` will unwrap unions and dereference types. Consider:

```ts
interface A { x: "a", a: String; }
interface B { x: "b", b: String; }
interface C { x: "c", c: String; }

type BC = B | C;
type ABC = A | BC;

possibleRootTypes(ABC) === [{ x: "a", a: String; }, { x: "b", b: String; }, { x: "c", c: String; }]
```
A discriminator is then inferred by examining each type and checking for a common property with a unique literal value. Intersection types bring a bit more complexity to this:

```ts
interface A { x: "a", a: String; }
interface B { x: "b", b: String; }
interface C { c: String; }
interface X { x: "c"; }

type ABC = A | B | (C & X)

possibleRootTypes(ABC) === [{ x: "a", a: String; }, { x: "b", b: String; }, (C & X)] // <- current behaviour
```
Even though the type `C & X` does actually have a conforming discriminator `x: "c"`, it would currently not be explored by the `inferDiscriminator` function.

After thinking this through some more, I think we have two high level options:
1. Handle intersection types appropriately anywhere `possibleRootTypes` is called
2. Change `possibleRootTypes` to resolve intersection types into an `object` type

Keen to hear your thoughts on this @mahirk

Admittedly this issue would have been far simpler to spot if parts of the codebase adopted the exhaustive `switch case` style rather than sometimes checking for specific types and defaulting behaviour otherwise 😢

lib/src/http.ts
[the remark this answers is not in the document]
mahirk: The intersection type here just attempts to be a composite data type, allowing us to assign both `AccountId` and `PurchaseId` to the `pathParams` object. 
For example:
```
@endpoint({
  method: "GET",
  path: "/v1/accounts/:accountId/purchase/:purchaseId"
})
class Endpoint {
  @request
  request(
    @pathParams
    pathParams: EndPointParams
  ) {}
}

interface AccountId {
  accountId: String;
}

interface PurchaseId {
  purchaseId: String;
}

type EndPointParams = AccountId & PurchaseId
```

Allows it to be cleaner and also allows us to reuse `AccountId` or `PurchaseId` when we see fit.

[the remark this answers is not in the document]
lfportal: Ah I see, I didn't think of that use case 👍

lib/src/http.ts
```diff
@@ -103,6 +105,7 @@ export function isHeaderTypeSafe(type: Type, typeTable: TypeTable): boolean {
     case TypeKind.OBJECT:
       return false;
     case TypeKind.UNION:
+    case TypeKind.INTERSECTION:
```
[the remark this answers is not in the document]
lfportal: Resolved: https://github.com/airtasker/spot/pull/1133#discussion_r529087606

mahirk:
@lfportal Apologies for the delay, with holiday season coming around work had gotten busy. I'll update this PR next week and try and continue the contribution on some of the other issues opened :)

lib/src/types.ts
mahirk: I used the same method used for union types to identify the possibleRoot types. Since `allOf` does not have any discriminators, those were not added to the openAPI specification, but were inferred for when we have unions with intersections.

Given how some of the parsing is set up, the example provided above would not work as is, however it does if you change it to the following when defining the contract. Using the parenthesis throws an "unknown type" error. After the changes the following are valid contracts:
Union with Intersection
```
interface A { x: "a", a: string; }
interface B { x: "b", b: string; }
interface C { c: string; }
interface X { x: "c"; }
type CX = C & X;
type ABC = A | B | CX
```
Union with Union
```
interface A { x: "a", a: string; }
interface B { x: "b", b: string; }
interface C { x: "c"; }
interface X { x: "d"; }
type CX = C | X;
type ABC = A | B | CX
```

Finally, another valid discriminator can be boolean types which are ignored. Before adding those in, I wanted to ask if that is a missing feature and should be added?

mahirk:
@lfportal Ready for review! There is one edge case where intersections can result in `never` which I will also work on and push!

lfportal:
> @lfportal Ready for review! There is one edge case where intersections can result in `never` which I will also work on and push!

Thanks @mahirk I'll take a look today or tomorrow 👀

lib/src/types.ts
lfportal: > Finally, another valid discriminator can be boolean types which are ignored. Before adding those in, I wanted to ask if that is a missing feature and should be added?

Let's keep non-string discriminator types as separate issues if there is demand for them.

lib/src/parsers/type-parser.ts
```diff
@@ -455,6 +461,43 @@ function parseUnionType(
...
+  const allowedTargetTypes = typeNode
+    .getTypeNodes()
+    .filter(type => !type.getType().isUndefined());
+  const types: Type[] = [];
+  for (const tn of allowedTargetTypes) {
+    const typeResult = parseType(tn, typeTable, lociTable);
+    if (typeResult.isErr()) return typeResult;
+    // Only allow objects, unions, intersections and references
+    const typeResultType = typeResult.unwrap();
+    if (isPrimitiveType(typeResultType) || isArrayType(typeResultType)) {
```
lfportal: Resolve/dereference to concrete types and check if the resulting type is an object type

lib/src/types.ts
```diff
@@ -134,6 +141,12 @@ export interface UnionType {
   discriminator?: string;
 }
 
+export interface IntersectionType {
+  kind: TypeKind.INTERSECTION;
+  types: Type[];
+  discriminator?: string;
```
lfportal: Perhaps we can remove `discriminator` for `IntersectionType`. There doesn't appear to be any good use case for it.

lib/src/types.ts
```diff
@@ -431,6 +460,28 @@ export function possibleRootTypes(
...
+          properties: concreteTypes.reduce<ObjectPropertiesType[]>(
+            (acc, curr) => {
+              return acc.concat(...curr.properties);
+            },
+            []
+          )
+        }
+      ];
+    }
+    return concreteTypes;
```
lfportal: Let's assume the type parser should ensure concrete types of intersections are object types and throw an error here otherwise.

lfportal (changes requested):
Love it, just a few more small comments/adjustments

mahirk:
@lfportal Made the updates you requested. I also made an update so that intersections that evaluate to `never` throw an error

lib/src/types.ts
```diff
@@ -431,6 +456,28 @@ export function possibleRootTypes(
...
+    }, []);
+    // if they are all objects then reduce them into one object
+    // to allow for proper serialization
+    if (concreteTypes.every(isObjectType)) {
+      return [
+        {
+          kind: TypeKind.OBJECT,
+          properties: concreteTypes.reduce<ObjectPropertiesType[]>(
+            (acc, curr) => {
+              return acc.concat(...curr.properties);
```
lfportal: Noticed this might have a similar problem to `doesInterfaceEvaluatesToNever` for cases that have overlapping or duplicate properties:

```ts
type A = { x: string, y: string } & { x: string, z: string };
type B = { x: string, y: string } & { x: "abc", z: string };
```

In this case we should resolve to the narrowest type:
```ts
type A = { x: string, y: string, z: string };
type B = { x: "abc", y: string, z: string };
```

mahirk: Looks like providing a narrow type might require more changes in the `type-parser.ts` to correctly drill down the type, which might get more complicated for some of the reference types 🤔 Updating the code for the previous comment for now

lib/src/types.ts
```diff
@@ -564,6 +560,46 @@ export function inferDiscriminator(
...
+
+  for (const candidate of possibleDiscriminators.keys()) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const values = possibleDiscriminators.get(candidate)!;
+    if (values.length > 1) {
+      return true;
+    }
+  }
+
+  return false;
```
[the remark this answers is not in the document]
mahirk: There was a small change to make, but `type A` would result `false` in the above function while `type B` would result in `true`

The logic in this is similar to the `inferDiscriminator` (and maybe could be reused at some point). We would need to add some additional logic for the case with competing `StringLiteralType` values, I'll start to add that, along with narrowing the types later.

lib/src/types.ts
mahirk: hey @lfportal I am able to find the narrow type and drill down to it, however when it starts being a reference chain, it starts to get a little complicated and starts to include elements which are not meant to be there. Specifically see the following example:

```
type ABC = M
type M = P & S;
type P = Z;
type Z = { x: string, y: string }
type S = T;
type T = O;
interface O { x: "abc", z: string };
```

results in:
```
    ABC:
      $ref: '#/components/schemas/M'
    M:
      allOf:
        - type: object
          properties:
            x:
              type: string
              enum:
                - abc
            'y':
              type: string
            z:
              type: string
          required:
            - x
            - 'y'
            - z
    O:
      type: object
      properties:
        x:
          type: string
          enum:
            - abc
        z:
          type: string
      required:
        - x
        - z
    P:
      $ref: '#/components/schemas/Z'
    S:
      $ref: '#/components/schemas/T'
    T:
      $ref: '#/components/schemas/O'
    Z:
      type: object
      properties:
        x:
          type: string
        'y':
          type: string
      required:
        - x
        - 'y'
```

Ideally, since M is the narrowed down type, the other elements are not required. 

If you have suggestions on how to do that, that would be great, though it is a little interesting for me to updating what the user may have entered and creating a new object out of it. 

I have pushed the code until the fixes, excluding the code which executes the narrow types above.

lfportal: I'm not sure I understand, for the purposes of `possibleRootTypes`, we should narrow types. However, we don't need to narrow types for OpenAPI output. To clarify I would still expect:
```
  M:
      allOf:
        - $ref: '#/components/schemas/P'
        - $ref: '#/components/schemas/S'
```

lib/src/types.ts
```diff
@@ -580,24 +572,37 @@ export function doesInterfaceEvaluatesToNever(
...
+    if (allPropertiesAreStringOrLiteral) {
+      // check for any potential conflicts in string-literal types
+      // StringLiteralType("abc") and StringLiteralType("abc") are compatible.
+      // Similarly StringLiteralType("abc") and StringType are compatible.
+      // However StringLiteralType("abc") and StringLiteralType("def") are not.
+      return (
+        new Set(propertyTypeList.filter(isStringLiteralType).map(v => v.value))
+          .size > 1
+      );
     }
```
lfportal: Let's repeat the same for `number` and `boolean`

mahirk:
Done for float, number and booleans. This is also extended to the "never" identification and "discriminator"/"rootTypes"

lib/src/types.ts
lfportal: @mahirk tagging in case you missed ☝️

mahirk: Yup! sorry @lfportal thank you. I updated it so it does not update the output, which made it a lot easier. Now it only combines them for `possibleRootTypes` which allows us to correctly manage the scenarios in https://github.com/airtasker/spot/pull/1133#discussion_r551862324

lfportal (approved):
Thanks @mahirk!, this is an awesome addition to the Spot DSL. I'll merge this and publish a release over the weekend.

Issue #1050, opened 2020-10-08
Allow generation of servers and base paths in `api` 


mahirk:
**Is your feature request related to a problem? Please describe.**
OpenAPI3 allows defining servers (https://swagger.io/docs/specification/api-host-and-base-path/) which can help define the scheme, description, and variables which form a server. This request is to enable spot to generate that for the OpenAPI Spec.

**Describe the solution you'd like**
Allow the definition of `server` inside the `@api` module of spot, with the following conversions:
```ts
@api({
  name: "My API",
  version: '1',
})
class Api {
  @servers
  server(
    @url
    url: 'https://{environment}.example.com/v2',
    @variables
    variables: {
      /** API description for server variables */
      environment: 'api' | 'api.dev' | 'api.staging'
    }
  ) {}
}
```

```yaml
servers:
  - url: https://{environment}.example.com/v2
    variables:
      environment:
        default: api 
        description: API description for environment variables
        enum:
          - api         
          - api.dev  
          - api.staging
```

**Describe alternatives you've considered**
Haven't thought of any other designs at this moment, likely will need to think about if it would be useful to add (a) a default value, which could be the `first` value?  and (b) multiple servers.

lfportal:
Server objects are OpenAPI 3 spec specific metadata. I'm thinking spec specific definitions should be prefixed for example `@oa3server` or `@openapi3server` rather than `@server`. 

Using the following example from [Server Object Example
](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#server-object-example)

```
{
  "servers": [
    {
      "url": "https://{username}.gigantic-server.com:{port}/{basePath}",
      "description": "The production API server",
      "variables": {
        "username": {
          "default": "demo",
          "description": "this value is assigned by the service provider, in this example `gigantic-server.com`"
        },
        "port": {
          "enum": [
            "8443",
            "443"
          ],
          "default": "8443"
        },
        "basePath": {
          "default": "v2"
        }
      }
    }
  ]
}
```

What do you think of the following syntax?

```ts
@api({
  name: "My API",
  version: '1'
})
class Api {
  /**
   * The production API server
   */
  @oa3server({ url: "https://{username}.gigantic-server.com:{port}/{basePath}" })
  productionServer(
    @oa3serverVariables
    variables: {
      /**
       * this value is assigned by the service provider, in this example `gigantic-server.com`
       * 
       * @default "demo"
       */
      username: String,
      /**
       * @default "8443"
[...]

PR #1376, opened 2021-04-21
airtasker#1367 Implementation for SchemaProps

### LINKED ISSUE #1367: Implement the OASv3 schema properties for types
**OASv3 schema properties for types needs to be implemented**

The OASv3 has some properties for types that is not implemented currently in Spot :
minLength, maxLength, pattern, minimum, maximum, default, ...

```javascript
        {
          description: "property-schemaprop description for string",
          name: "status",
          in: "header",
          required: true,
          schema: {
            type: "string",
            minLength: 12,
            maxLength: 20,
            pattern: "^[0-9]{3}-[0-9]{8}$"
          }
        },
        {
          description: "property-schemaprop description for integer",
          name: "size",
          in: "header",
          required: true,
          schema: {
            format: "int32",
            type: "integer",
            minimum: 1,
            maximum: 100,
            default: 42
          }
        }
```

**Solution**

Create a new annotation in the description of attributes like it's implemented with **@example** 

**@schemaprop** could be the name, then put the key on the first line and the value on the second line

```ts
/** property-schemaprop description
   * @schemaprop pattern
   * "property-schemaprop-value"
   *  */
  "property-with-schemaprop": string;
  /** property-two-schemaprops description
   * @schemaprop minimum
   * 123
   * @schemaprop default
   * 456
   * */
  "property-with-schemaprops": Integer;
```

Some checks should be done to be compliant with OASv3

**OASv3 schema properties are already there waiting for implementation**

lib/src/generators/openapi3/openapi3-specification.ts
```ts
interface NumberSchemaObjectBase {
  maximum?: number;
  exclusiveMaximum?: boolean;
  minimum?: number;
  exclusiveMinimum?: boolean;
  multipleOf?: number;
  enum?: (number | null)[];
  default?: number;
}

export interface StringSchemaObject extends SchemaObjectBase {
  type: "string";
  maxLength?: number;
  minLength?: number;
  /**
   * OpenAPI allows custom formats. We constrain the format here to those
   * that OpenAPI has defined and custom formats that Spot may produce.
   */
  format?: "date" | "date-time" | "password" | "byte" | "binary";
  pattern?: string;
  enum?: (string | null)[];
  default?: string;
}
```

Thodin3:
Please see #1367
The OASv3 has some properties for types :
minLength, maxLength, pattern, minimum, maximum, default, ...
It's possible to implement it in comments in the contract parameters.

## Checklist:

- [X] I've created an issue associated with this PR


lib/src/types.ts
```diff
@@ -824,4 +826,5 @@ export class TypeTable {
 export interface TypeDef {
   type: Type;
   description?: string;
+  schemaProps?: SchemaProp[];
```
lfportal: Not all attributes in `SchemaProp` are applicable to each `Type`. Can we move the attributes to each individual type?

lib/src/definitions.ts
```diff
@@ -59,26 +59,34 @@ export interface Header {
...
+  value: any; // TODO: encapsulate type information
+}
+
 export interface QueryParam {
   name: string;
   description?: string;
   type: Type;
   optional: boolean;
   examples?: Example[];
+  schemaProps?: SchemaProp[];
```
lfportal: These attributes conceptually are a part of a `Type`. Can we encapsulate these to the `Type` definitions?

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
```diff
@@ -0,0 +1,47 @@
...
+       * "^[0-9a-z_]+$"
+       *  */
+      status: String;
+      /** property-schemaprop description for integer
+       * @schemaprop minimum
+       * 1
+       * @schemaprop maximum
+       * 100
+       * @schemaprop default
+       * 42
```
lfportal: We are currently using `@default` tag to convey default values in [`@oa3serverVariables`](https://github.com/airtasker/spot/wiki/Spot-syntax#openapi-3-servers). I think it makes sense for `@default` to be used generically wherever we need default values.

For any OpenAPI specific tags, the convention we approached this with is to prefix with:

* `@oa*` for OpenAPI 2 & 3
* `@oa2*` for OpenAPI 2
* `@oa3*` for OpenAPI 3

🙇‍♂️ Sorry we hadn't documented this - see https://github.com/airtasker/spot/issues/1050 for more context.

For these schema props we can prefix with `oa*` since they'll be applicable for both OpenAPI 2 & 3.

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
```diff
@@ -0,0 +1,47 @@
...
+  @request
+  request(
+    @headers
+    headers: {
+      /** property-schemaprop description for string
+       * @schemaprop minLength
+       * 12
+       * @schemaprop maxLength
+       * 20
+       * @schemaprop pattern
```
lfportal: It may be a bit more effort to implement, but what do you think of using separate tags for each schema prop? i.e.

```ts
/**
  * @oaSchemaMinLength 12
  * @oaSchemaMaxLength 20
  * @oaSchemaPattern "^[0-9a-z_]+$"
  */
```

lfportal (commented):
Thanks @Thodin3, just a couple of comments.

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
Thodin3: Hello @lfportal,
It's a good idea to rename it `@oaSchemaProp` and make `@default` tag works in schema properties as well.

Do you agree?

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
Thodin3: Hi @lfportal ,

This change is a huge refactor, so could we do like `@oa3serverVariables` for generic name like you sugested on the other comment?
So to rename it like `@oaSchemaProp`

What do you think?

lib/src/types.ts
Thodin3: Yes, good idea

lib/src/definitions.ts
Thodin3: Yes, if we move it to `Type`, it will be easier

lib/src/generators/openapi3/openapi3.ts
```diff
@@ -309,7 +309,8 @@ function headerToHeaderObject(
   return {
     description: header.description,
     required: !header.optional,
-    schema: typeToSchemaOrReferenceObject(header.type, typeTable)
+    schema: typeToSchemaOrReferenceObject(header.type, typeTable),
+    examples: exampleToOpenApiExampleSet(header.examples)
```
lfportal, line 313: Thanks 👍

lib/src/types.ts
```diff
@@ -128,27 +146,32 @@ export interface ObjectPropertiesType {
...
 export interface IntersectionType {
   kind: TypeKind.INTERSECTION;
   types: Type[];
+  schemaProps?: SchemaProp[];
 }
 
 export interface ReferenceType {
   kind: TypeKind.REFERENCE;
   name: string;
+  schemaProps?: SchemaProp[];
```
lfportal: Do any of these non-primitive types require `schemaProps`?

lib/src/types.ts
```diff
@@ -64,58 +69,71 @@ export type LiteralType =
 
 export interface NullType {
   kind: TypeKind.NULL;
+  schemaProps?: SchemaProp[];
```
lfportal: Does `null` require `schemaProps`?

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
lfportal: @Thodin3 makes sense, this approach is more practical for now. A larger refactor can be considered later. 👍

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
lfportal: @Thodin3 makes sense, this approach is more practical for now. A larger refactor can be considered later. 👍

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
lfportal: Yup agreed 👍

lib/src/types.ts
```diff
@@ -128,22 +145,26 @@ export interface ObjectPropertiesType {
...
   kind: TypeKind.UNION;
   types: Type[];
   discriminator?: string;
+  schemaProps?: SchemaProp[];
 }
 
 export interface IntersectionType {
   kind: TypeKind.INTERSECTION;
   types: Type[];
+  schemaProps?: SchemaProp[];
```
lfportal, line 167: I don't think `UnionType` and `IntersectionType` have any need for `schemaProps` - doesn't look like OpenAPI supports any props for `allOf` or `oneOf`.

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
Thodin3: ok cool

lib/src/generators/openapi3/__spec-examples__/contract-with-schemaprops.ts
Thodin3: ok

lib/src/types.ts
Thodin3: For `allOf` or `oneOf`, we can see that it inherits from `SchemaObjectBase`, so it includes the following attributes:
- `title`
- `example`
- `deprecated` (only for OAS3)

For `UnionType`, we will use it for `String` enum for example.

I added a config table for all combinations in `lib/src/parsers/schemaprop-parser.ts`
Based on the list of properties in this config table, there is some new contract examples in `lib/src/generators/openapiX/__spec-examples__/contract-with-schemaprops.ts` parsed by unit tests.
Then, I added also the OAS2 implementation for SchemaProperties.
Let me know.

lib/src/generators/openapi3/openapi3.ts
Thodin3: Yes, little fix for examples 👍

lib/src/types.ts
Thodin3: Fixed for `ReferenceType` & `NullType`

lib/src/types.ts
Thodin3: Fixed

lib/src/types.ts
lfportal: 👍 thanks @Thodin3

lib/src/parsers/schemaprop-parser.ts
```diff
@@ -0,0 +1,264 @@
...
+  if (rawDefaults && rawDefaults.length > 0) {
+    rawDefaults.every(defaultValue => {
+      rawSchemaProps.push("default\n" + defaultValue);
+    });
+  }
+
+  if (rawSchemaProps && rawSchemaProps.length > 0) {
+    const schemaProps: SchemaProp[] = [];
+    let schemaPropError;
+    const propTypeMap = new Map<
```
lfportal: We can move this `Map` to a constant outside the `extractJSDocSchemaProps` function for readability.

Thodin3: ok, done

lfportal (approved):

Thodin3:
Hello @lfportal,
Thanks.
I can see that the CI-Test is failing on master now.
This is because of the new feature: property.with.dots

Possible fix is:
add `schemaProps: undefined` in `type: {kind: TypeKind.STRING}` in `lib/src/parsers/query-params-parser.spec.ts:128`

PR #1419, opened 2021-05-11
airtasker#1417 @oaSchemaProp example fails with numerical strings, Date, DateTime, Object, Array, Union or Intersection

### LINKED ISSUE #1417: @oaSchemaProp example fails with numerical strings
**Describe the bug**

When using `@oaSchemaProp example`, providing a numerical string like `"123.3"` for a `String` type fails:

```javascript
 /**
   * @oaSchemaProp example
   * "123.3"
   */
  amount: String;
```

```
Error: property type is wrong or property not allowed
    at Object.extractJSDocSchemaProps (.../api-spec/node_modules/@airtasker/spot/build/lib/src/parsers/schemaprop-parser.js:116:31)
```

As `"123.3"` *is* a String, this should work.

Thodin3:

Please see #1417
When using `@oaSchemaProp example`, providing a numerical string like "123.3" for a String type fails.
When using `@oaSchemaProp example` for `Date`, `DateTime`, Object, Array, Union or Intersection type it fails.

Fix in this PR :
in `function extractJSDocSchemaProps(`
`const typeOf` is refactored and will use the `JSON.parse` real type instead of regexp filtering.

## Checklist:

- [X] I've created an issue associated with this PR

### DIFF
--- a/lib/src/generators/openapi2/__snapshots__/openapi2.spec.ts.snap
+++ b/lib/src/generators/openapi2/__snapshots__/openapi2.spec.ts.snap
@@ -934,6 +934,9 @@ exports[`OpenAPI 2 generator schemaprops contract with schemaprops parses correc
                     ],
                     \\"minProperties\\": 1,
                     \\"maxProperties\\": 100,
+                    \\"example\\": {
+                      \\"price\\": 3.14
+                    },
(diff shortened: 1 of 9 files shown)


lfportal (approved):
LGTM, thanks @Thodin3

PR #1705, opened 2021-11-20
docs: Fixup readme command links from build/js to src/ts

janbrasna:

- Right now the usage command source links in README here on GH lead to 404 resources… as the build output is not a part of the hosted repo here.
- Seems like a good candidate for such links would be the ts source, instead of trying to reference the hypothetical js built outputs?
- _(Not sure if the README content is generated by a release script or similar tooling, couldn't find any such hint as all release versioning changes apparently seem to be done manually by a human(?) — hence proposing this here directly for README.md file itself… if the right place to correct these links lies elsewhere, I'd be happy to fix it there if you point me to the right direction.)_

## Checklist:

- [ ] ~I've added/updated tests to cover my changes~
- [ ] ~I've created an issue associated with this PR~

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ EXAMPLE
   $ spot checksum api.ts
 ```
 
-_See code: [build/cli/src/commands/checksum.js](https://github.com/airtasker/spot/blob/v1.9.0/build/cli/src/commands/checksum.js)_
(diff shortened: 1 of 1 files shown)


janbrasna:
_(still passing w/o conflict as there hasn't been any release published since, so the versions in diff still match…)_

janbrasna:
If anyone wants me to update this to current release version links I'm happy to resolve the conflict.

The questions in the main PR description are still open.

PR #2062, opened 2023-09-27
[PSC-1988] Fix `--proxyBaseUrl` argument to `spot mock`

timdawborn:

There exists functionality in `spot` to be able to run a mock server with an optional proxy. When the proxy isn't supplied, all endpoints that are defined in the given contact will have a stubbed response returned, with random, structurally valid data being returned. When the optional proxy argument is supplied, any non-draft SPOT contracts will be proxied through to the upstream instead of having mocked data returned.

The intent of this proxy argument is to allow you to perform local development against a stubbed new endpoint that doesn't yet have an implementation upstream.

When attempting to use this though, I found that the proxy was never being used, even though the code looked like it was meant to be doing the right thing. Upon further digging, I found that the proxy argument was never setup to be used correctly due to a bug in the code that TypeScript did not pick up on, due to the use of a generic object of options being passed to an object.

This PR:
1. Fixes the aforementioned bug, and now actually passes the proxy configuration through from the CLI parser to the mock server function call.
2. Updates the proxying logic to set the `host` header correctly when invoking HTTPs endpoints.

## Checklist:

### DIFF
--- a/cli/src/commands/mock.ts
+++ b/cli/src/commands/mock.ts
@@ -50,7 +50,7 @@ export default class Mock extends Command {
       await runMockServer(contract, {
         port,
         pathPrefix: pathPrefix ?? "",
-        ...proxyConfig,
+        proxyConfig,
         logger: this
       }).defer();
       this.log(`Mock server is running on port ${port}.`);
--- a/cli/src/common/infer-proxy-config.ts
+++ b/cli/src/common/infer-proxy-config.ts
@@ -7,7 +7,7 @@ export default function inferProxyConfig(
     return null;
   }
 
-  const url = new URL(proxyBaseUrl)
+  const url = new URL(proxyBaseUrl);
   if (url.protocol !== "http:" && url.protocol !== "https:") {
     throw new Error(
       'Could not infer protocol from proxy base url, should be either "http" or "https".'
@@ -17,6 +17,6 @@ export default function inferProxyConfig(
   return {
     isHttps: url.protocol === "https:",
     host: url.host,
-    path: url.pathname,
+    path: url.pathname
   };
 }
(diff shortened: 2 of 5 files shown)


cli/src/commands/mock.ts
```diff
@@ -50,7 +50,7 @@ export default class Mock extends Command {
       await runMockServer(contract, {
         port,
         pathPrefix: pathPrefix ?? "",
-        ...proxyConfig,
+        proxyConfig,
```
timdawborn, line 53: This was a bug. The options passed to `runMockServer` expected a `ProxyConfig` object, but this call was exploding the proxy configuration instead of just passing it. This resulted in the proxy configuration never being used.

cli/src/common/infer-proxy-config.ts
```diff
@@ -7,16 +7,16 @@ export default function inferProxyConfig(
     return null;
   }
 
-  const [protocol] = proxyBaseUrl.split("://");
-
-  if (protocol !== "http" && protocol !== "https") {
+  const url = new URL(proxyBaseUrl);
```
timdawborn, line 10: Update the `ProxyConfig` object to have the relevant separated pieces of information to be able to mutate the proxied request accordingly (specifically, the host).

flavray (approved):
Neat ✨

lib/src/mock-server/proxy.ts
```diff
@@ -1,27 +1,38 @@
...
+          ? 443
+          : 80
+        : proxyConfig.port,
+    path: proxyConfig.path + incomingRequest.path,
+    headers: {
+      ...incomingRequest.headers,
+      host: proxyConfig.host
+    }
   };
+  console.log(options);
```
njday: **Question**: is it intentional to keep this log here?

njday (approved):
LGTM - just that comment about the log which can be done as follow up if unintentional

lib/src/mock-server/proxy.ts
timdawborn: Oops, no. That wasn't intended. I'll remove them now.

PR #2063, opened 2023-10-03
[PSC-1988] More `spot mock` fixes

timdawborn:

Three misc fixes for `spot mock` after attempting to use it more thoroughly.

## Checklist:

### DIFF
--- a/lib/src/mock-server/proxy.ts
+++ b/lib/src/mock-server/proxy.ts
@@ -35,6 +35,11 @@ export function proxyRequest({
     response.writeHead(res.statusCode ?? response.statusCode, res.headers);
     res.pipe(response);
   });
+  proxyRequest.on("error", e => {
+    console.error(`Failed to proxy request: ${e}`, e.stack);
+    response.statusCode = 500;
+    response.send();
+  });
 
   if (incomingRequest.body) {
     proxyRequest.write(incomingRequest.body);
--- a/lib/src/mock-server/proxy.ts
+++ b/lib/src/mock-server/proxy.ts
@@ -41,7 +41,7 @@ export function proxyRequest({
     response.send();
   });
 
-  if (incomingRequest.body) {
+  if (incomingRequest.body && incomingRequest.headers["content-length"]) {
     proxyRequest.write(incomingRequest.body);
   }
 
--- a/lib/src/mock-server/proxy.ts
+++ b/lib/src/mock-server/proxy.ts
@@ -41,7 +41,7 @@ export function proxyRequest({
     response.send();
   });
 
-  if (incomingRequest.body && incomingRequest.headers["content-length"]) {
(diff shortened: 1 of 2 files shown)


lib/src/mock-server/server.ts
```diff
@@ -33,6 +33,7 @@ export function runMockServer(
   }
 ) {
   const app = express();
+  app.use(express.raw({type: "*/*"}));
```
timdawborn, line 36: This is to force `express` to always populate the `request.body` attribute, event on `GET` requests. Before this change, received `GET` requests with bodies did had a `undefined` `request.body`.

lib/src/mock-server/server.ts
```diff
@@ -72,7 +73,10 @@ export function runMockServer(
         return;
       }
     }
+
     logger.error(`No match for request ${req.method} at ${req.path}.`);
+    resp.status(404);
+    resp.send();
```
timdawborn, line 79: Upon not finding a valid route, we still need to respond with a HTTP response. Add a 404 response.

lib/src/mock-server/proxy.ts
```diff
@@ -35,6 +35,11 @@ export function proxyRequest({
     response.writeHead(res.statusCode ?? response.statusCode, res.headers);
     res.pipe(response);
   });
+  proxyRequest.on("error", e => {
+    console.error(`Failed to proxy request: ${e}`, e.stack);
+    response.statusCode = 500;
+    response.send();
+  });
```
timdawborn, line 42: We previously were not handling any form of network or connection error on the proxied request. Any errors would cause the mock server to die. This instead logs the error and returns a 500 response.

lib/src/mock-server/server.ts
```diff
@@ -33,6 +33,7 @@ export function runMockServer(
   }
 ) {
   const app = express();
+  app.use(express.raw({ type: () => true }));
```
timdawborn, line 36: This is to force `express` to always populate the `request.body` attribute, event on `GET` requests. Before this change, received `GET` requests with bodies did had a `undefined` `request.body`. This meant that the request body wasn't being proxied through, causing the upstream to reject the request or hang.

lib/src/mock-server/proxy.ts
```diff
@@ -35,8 +35,13 @@ export function proxyRequest({
...
 
-  if (incomingRequest.body) {
+  if (incomingRequest.body && Buffer.isBuffer(incomingRequest.body)) {
```
timdawborn, line 44: After the change below with `express.raw` as [per the documentation for this middleware](https://expressjs.com/de/api.html#express.raw), if there is no request body, `request.body` will be populated with `{}`, otherwise with a `Buffer` object. Why, no idea. Regardless, we only want to proxy the body through if it's actually a `Buffer` object.

njday (approved):
LGTM

PR #2066, opened 2023-10-04
[PSC-1988] Enable `spot mock` to have a 404 fallback proxy

timdawborn:

In order to make the mock server more useful in the case where you have multiple different upstream servers on the same hostname, this PR adds a new optional fallback proxy to `spot mock`, which is used instead of instantly 404'ing the received request if the request does not match any of the given contracts. If provided, instead of 404'ing, it will send the request through to the fallback proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,12 +283,15 @@ ARGUMENTS
   SPOT_CONTRACT  path to Spot contract
 
 OPTIONS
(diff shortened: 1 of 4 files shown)


njday (approved):
LGTM 👍

flavray (approved):

PR #2068, opened 2023-10-06
[NT] Normalise paths before proxying.

timdawborn:

Due to various layers of proxying, sometimes a proxied request ends up with multiple leading slashes. Some web servers do not not normalise request paths, leading to 404's.

This PR normalises the received request path on the mock server before matching against contract paths, and normalises the received request paths when proxying the request.

## Checklist:


flavray (approved):
LGTM 👍

Sheep It

PR #2069, opened 2023-10-10
[NT] Correct the hostname for proxy configurations with port numbers

timdawborn:

This PR fixes a bug in the proxy hostname parsing, where we currently accidentally use the wrong field from the parsed `URL` object to refer to the hostname of the proxy. https://developer.mozilla.org/en-US/docs/Web/API/URL — the `host` attribute is of the form `<host>:<port>` if present, whereas `hostname` is just `<host>`. How we use the parsed attribute expects just a hostname without a port number.

This PR fixes this bug by using the `hostname` attribute of the parsed URL instead of the `host` attribute. While I was here, I also added tests for this proxy URL parsing function given there were none before.

## Checklist:

### DIFF
--- a/cli/src/common/infer-proxy-config.spec.ts
+++ b/cli/src/common/infer-proxy-config.spec.ts
@@ -1,0 +1,69 @@
+import inferProxyConfig from "./infer-proxy-config";
+
+describe("inferProxyConfig", () => {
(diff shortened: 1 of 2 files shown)


flavray (approved):
LGTM - The test suite is a nice addition too 👍

PR #2512, opened 2026-02-04
[ACQ-6253] Add GitHub Package Registry publishing to workflow

tory24:

**Why is this change required?** 
Some repos expect @airtasker/spot to be available in the Github Registry, however this was never an issue as their lock files have conitnued to resolve using npmjs. With a clean install or migration to another package manager, they now throw a 404 when trying to find spot in Github.

The current workflow only publishes to npm registry. To make the package available on GitHub Package Registry as well, we need to add publishing steps for both registries.

**What does it do?**
  - Adds a separate publish step for GitHub Package Registry
  - Configures authentication for both npm and GitHub Package Registry
  - Adds required `packages: write` permission to the publish job
  - Uses `pnpm publish` with `--no-git-checks` flag for both registries
  - Maintains backward compatibility with existing npm publishing

## Changes Made

1. **Added permissions**: Added `packages: write` permission to the publish job to allow publishing to GitHub Package Registry
2. **NPM Publishing**: Updated to use `pnpm publish` instead of `npm publish` for consistency
3. **GitHub Registry Publishing**: Added authentication and publish steps for GitHub Package Registry using `GITHUB_TOKEN`
4. **Registry Configuration**: Properly configures the `.npmrc` file for each registry before publishing

## Checklist:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -93,14 +93,7 @@ jobs:
           EOF
 
       - name: Publish to NPM
-        id: npm-publish
         run: pnpm publish --access=public
-
-      - name: Verify NPM publish
-        run: |
-          echo "Waiting for NPM registry propagation..."
-          sleep 10
-          npm view @airtasker/spot@${{ github.event.release.tag_name }} version
 
       - name: Authenticate with GitHub Package Registry
         run: |
@@ -113,14 +106,6 @@ jobs:
         id: gpr-publish
         run: pnpm publish --registry=https://npm.pkg.github.com --access=public
         continue-on-error: true
-
-      - name: Verify GPR publish
-        if: steps.gpr-publish.outcome == 'success'
-        run: |
-          echo "Waiting for GitHub Package Registry propagation..."
-          sleep 10
-          npm view @airtasker/spot@${{ github.event.release.tag_name }} version \
-            --registry=https://npm.pkg.github.com
 
       - name: Report GPR failure
         if: steps.gpr-publish.outcome == 'failure'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -10,9 +10,6 @@ on:
   release:
     types: [published]
 
-env:
-  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
 jobs:
   test:
     name: test-node-${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -73,12 +73,23 @@ jobs:
...
       - uses: actions/checkout@v5
       - uses: ./.github/actions/prepare-repository
 
       - name: Authenticate with NPM registry
         run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
 
-      - name: Publish
-        run: npm publish --access=public
+      - name: Publish to NPM
+        run: pnpm publish --access=public --no-git-checks
```
justinnais: why not git checks?

tory24: Yep overlooked that change, removing

.github/workflows/build-and-test.yml
```diff
@@ -73,12 +79,37 @@ jobs:
...
+        run: |
+          cat > ~/.npmrc << 'EOF'
+          //npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}
+          @airtasker:registry=https://npm.pkg.github.com
+          EOF
+
+      - name: Publish to GitHub Package Registry
+        id: gpr-publish
+        run: pnpm publish --registry=https://npm.pkg.github.com --access=public
+        continue-on-error: true
```
tory24, line 102: This was added because if Github fails but NPM didn't then the release would be flagged as having failed.
Unsure if this is the best approach.

.github/workflows/build-and-test.yml
```diff
@@ -73,12 +79,37 @@ jobs:
...
+
+      - name: Report GPR failure
```
tory24: Wondering if this should actually be some kind of slack alert

.github/workflows/build-and-test.yml
```diff
@@ -73,12 +79,37 @@ jobs:
...
-        run: npm publish --access=public
+      - name: Publish to NPM
+        run: pnpm publish --access=public
+
+      - name: Authenticate with GitHub Package Registry
+        run: |
+          cat > ~/.npmrc << 'EOF'
+          //npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}
+          @airtasker:registry=https://npm.pkg.github.com
+          EOF
```
justinnais: Any reason for these to override each other? Any downside to having a single step to create the npmrc for both registries?

.github/workflows/build-and-test.yml
```diff
@@ -73,12 +79,37 @@ jobs:
...
+      packages: write
     steps:
       - uses: actions/checkout@v5
       - uses: ./.github/actions/prepare-repository
 
       - name: Authenticate with NPM registry
-        run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
+        run: |
+          cat > ~/.npmrc << 'EOF'
+          //registry.npmjs.org/:_authToken=${{ env.NPM_TOKEN }}
```
justinnais: should this be:
```suggestion
          //registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}
```

or do we need to add NPM_TOKEN to the job envs

timdawborn (dismissed):
LGTM 👍

.github/workflows/build-and-test.yml
tory24: Ah not necessarily, it was just already available, but can simplify for sure

.github/workflows/build-and-test.yml
tory24: No reason, I don't think this would cause any issues, but I've combined the two in a new commit

vlj91 (approved):
LGTM.

justinnais (approved):
bing

PR #2521, opened 2026-02-12
[ACQ-6298] Fix pnpm-workspace.yaml configuration structure

justinnais:
## Summary
- Fixed invalid `settings:` nesting in pnpm-workspace.yaml
- Removed unnecessary configuration options (`autoInstallPeers`, `strictPeerDependencies`)
- Moved `minimumReleaseAge` to root level as per pnpm specification
- Added `minimumReleaseAgeExclude` for `@airtasker/*` packages

## Related Tickets
- ACQ-6298

## Technical Details
The pnpm-workspace.yaml file had an invalid `settings:` wrapper that is not supported by pnpm's workspace configuration. Configuration options like `minimumReleaseAge` should be at the root level of the file, not nested under `settings:`.

Removed options:
- `autoInstallPeers: true` - default behavior, not needed
- `strictPeerDependencies: false` - default behavior, not needed

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,12 +1,7 @@
 packages:
   - 'docs'
 
-settings:
-  # Automatically install peer dependencies
-  autoInstallPeers: true
-
-  # Don't fail on unmet peer dependencies (migration compatibility)
-  strictPeerDependencies: false
-
-  # Security: 7-day delay before using newly published packages (10,080 minutes)
-  minimumReleaseAge: 10080
+# Security: 7-day delay before using newly published packages (10,080 minutes)
+minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+  - '@airtasker/*'


pnpm-workspace.yaml
```diff
@@ -1,12 +1,7 @@
...
-
-  # Don't fail on unmet peer dependencies (migration compatibility)
-  strictPeerDependencies: false
-
-  # Security: 7-day delay before using newly published packages (10,080 minutes)
-  minimumReleaseAge: 10080
+# Security: 7-day delay before using newly published packages (10,080 minutes)
+minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+  - '@airtasker/*'
```
tory24, line 7: Ahh, i think we need this for the web PR, I think it was an array

pnpm-workspace.yaml
```diff
@@ -1,12 +1,7 @@
...
-  autoInstallPeers: true
-
-  # Don't fail on unmet peer dependencies (migration compatibility)
-  strictPeerDependencies: false
-
-  # Security: 7-day delay before using newly published packages (10,080 minutes)
-  minimumReleaseAge: 10080
+# Security: 7-day delay before using newly published packages (10,080 minutes)
+minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
```
tory24, line 6: Any chance we can add the description just for inline documentation purposes:


```suggestion
# (Default: undefined) - A list of dependencies that always need to install the newest version immediately
minimumReleaseAgeExclude:
```

tory24 (commented):
Can we a

pnpm-workspace.yaml
justinnais: could you elaborate here? web is the same no?

tory24 (approved):

PR #2593, opened 2026-04-23
fix(RVNGR-1577): emit all leaf discriminators for nested-union members in unionTypeToDiscrimintorObject

mirzahaiderali:

- **Why is this change required?**  
  `unionTypeToDiscrimintorObject` in `lib/src/generators/openapi3/openapi3-type-util.ts` used `.find()` to resolve the discriminator property from a union member's concrete leaves. When `possibleRootTypes(t)` returned multiple leaves (i.e. `t` is a reference to a union type), only the first leaf's discriminator value was ever emitted into the `mapping`. Every subsequent leaf was silently dropped.

  This caused `ads_slider_double` to be absent from the generated `api.yml` for `BrowseTasksFeedListFilterType`, which in turn meant the iOS Swift decoder and Android Kotlin decoder had no `case "ads_slider_double":` branch — falling through to `.undecodable` and crashing at runtime when the BFF returned a double-slider filter.

- **What does it do?**  
  Replaces the single `.find()` + one `acc[value] = ref` write with a `for...of` loop that iterates **all** concrete leaves returned by `possibleRootTypes()`, emitting one mapping entry per leaf. Each entry maps back to the outer reference name (`t.name`) so downstream consumers still decode via the wrapper schema — matching the semantic intent of the original code.

  This is the minimal change required. The `objectReferenceOnly` guard and all sanity-check `throw`s are preserved. No other callsites are affected.

**Reproduces:** `BrowseTasksFeedListFilterType` with `AdsSlider = AdsSliderSingle | AdsSliderDouble` — before this fix, only `ads_slider_single` was emitted; `ads_slider_double` was missing.  
**Fixes:** [RVNGR-1577](https://airtasker.atlassian.net/browse/RVNGR-1577)  
**Related mobile PR (manual workaround):** airtasker/mobile#586

## Checklist:

[RVNGR-1577]: https://airtasker.atlassian.net/browse/RVNGR-1577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/generators/openapi3/openapi3-type-util.spec.ts
+++ b/lib/src/generators/openapi3/openapi3-type-util.spec.ts
@@ -777,43 +777,41 @@ describe("OpenAPI 3 type util", () => {
       });
     });
 
-    test("emits mapping entries for all three leaves when two union members are both nested unions", () => {
+    test("emits all four leaves when both union members are themselves nested unions", () => {
+      // Symmetric case: UnionA = A1 | A2, UnionB = B1 | B2
+      // Catches off-by-one issues in the loop across multiple nested members.
       const typeTable = new TypeTable();
-      typeTable.add("UnionASingle", {
+      typeTable.add("UnionA1", {
         type: objectType([
-          {
-            name: "type",
-            type: stringLiteralType("union_a_single"),
-            optional: false
-          }
+          { name: "type", type: stringLiteralType("union_a_one"), optional: false }
         ])
       });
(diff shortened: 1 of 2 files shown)


timdawborn (approved):
Makes sense. LGTM 👍

PR #2708, opened 2026-08-18
fix(COMPASS-17): resolve @airtasker/spot from spot's own tree

timdawborn:
## Ticket

## What

Fix the `@airtasker/spot` path mapping that the contract project is built with, and move the project construction into `lib/src/ts-project.ts`.

## Why

Spot cannot currently run without a consumer `node_modules`. The contract project mapped `@airtasker/spot` to `<parser dir>/../lib`, which resolves to `lib/lib` from `lib/src/parser.ts` and `build/lib/lib` from the compiled output. Neither path exists, so a contract's own `import ... from "@airtasker/spot"` only ever resolved through a copy of the package installed beside the contract. `parse()` throws on any TypeScript diagnostic, so an unresolved import fails every command that parses a contract.

That is fine while every consumer installs the npm package next to its contracts. It blocks the rest of COMPASS-17, where spot is invoked as a container against a repo that has no Node toolchain at all.

Correcting the mapping to `<parser dir>/lib` resolves `lib/src/lib.ts` from source and `build/lib/src/lib.d.ts` from the compiled output — the pattern already used in `lib/src/spec-helpers/helper.ts`, which sits one directory deeper and so spells the same target as `../lib`.

**This changes which declarations win when a consumer has `@airtasker/spot` installed.** TypeScript consults `paths` before `node_modules`. On `master` the mapped target did not exist, so the entry silently fell through and an installed copy always won. Correcting the target makes the mapping resolve, so from here a contract is always type-checked against the declarations of the Spot that is *running*, never against a different version installed beside it.

That is the intended rule — the parser and the declarations it checks against should be one version, and the container work requires it — but it is a real behaviour change, not a no-op. A repo pinned to an older `@airtasker/spot` that runs a newer Spot (`npx @airtasker/spot@latest`, a hoisted copy in a monorepo, or the image) now type-checks its contracts against the newer declarations. Since `parse()` throws on any diagnostic, a symbol removed between the two versions fails the command rather than resolving quietly. The precedence is recorded in the `ts-project.ts` docstring.

## Changes

- `lib/src/ts-project.ts` (new) — `contractCompilerOptions` and `createContractProject()`, with the corrected mapping. One home for the options, which `ts-lint` will also need in order to type-check contracts under exactly the options `parse()` uses.
- `lib/src/parser.ts` — consumes `createContractProject()`; the local `createProject()` is gone.
- `lib/src/parser.spec.ts` (new) — regression coverage, see below.
- `lib/src/__spec-examples__/contract-with-type-error.ts` (new) — a contract that is valid apart from one type error.
- `test-fixtures/contract/` (new) — a multi-file contract that imports `@airtasker/spot`. Consumed by the spec here, and by the docker image-parity script in a later PR of this ticket, which runs it through both the local CLI and the image.
- `.eslintrc.js` — `test-fixtures/**/*.ts` joins the existing contract-syntax rule-off list alongside `**/__spec-examples__/**/*.ts`. Contract files are empty-bodied decorated methods; without this, `no-empty-function` errors on every response handler.

`lib/src/spec-helpers/helper.ts` keeps its own `createProject()` — it is a spec-only helper with a different option set, and folding the two together is not this change.

## How this was verified

- `pnpm test` — 529 tests across 48 suites pass. `pnpm lint:check` clean (25 pre-existing warnings, 0 errors).
- **Fail-before / pass-after on the spec.** There is no `node_modules/@airtasker/spot` in this repository, so the fixture's import can only resolve through the mapping under test. Reverting just the path back to `../lib` fails the first case with `Cannot find module '@airtasker/spot' or its corresponding type declarations`; restoring it passes. The second case parses `contract-with-type-error.ts` and asserts the throw, so the guard stays fail-closed — otherwise a mapping that silently stopped type-checking would pass the first case too.
- **End-to-end, no `node_modules` on any parent path.** Built the CLI, copied `test-fixtures/contract/` into a scratch directory outside the repo, checked no `node_modules` exists on any ancestor of it, and parsed `api.ts` there against the built output. Fails with the same unresolved-import error before the change; returns the parsed `widget-api` contract after.

## Notes for the follow-up PRs

`bin/run` requires `@oclif/errors/handle`, but `@oclif/errors` is not a declared dependency — it is reached only by hoisting. Under pnpm's default isolated layout `node bin/run <cmd>` fails with `MODULE_NOT_FOUND` before it reaches any command. This is pre-existing on `master` and unrelated to this change, but the image build in the docker PR installs with pnpm and will hit it, so it gets fixed there.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,8 @@ module.exports = {
     {
       files: [
         "**/__spec-examples__/**/*.ts",
-        "lib/src/validation-server/spots/**/*.ts"
+        "lib/src/validation-server/spots/**/*.ts",
+        "test-fixtures/**/*.ts"
       ],
       rules: {
         "@typescript-eslint/ban-types": "off",
--- a/lib/src/__spec-examples__/contract-with-type-error.ts
+++ b/lib/src/__spec-examples__/contract-with-type-error.ts
@@ -1,0 +1,33 @@
+import { api, body, endpoint, request, response } from "@airtasker/spot";
+
+@api({ name: "type-error-api" })
+class TypeErrorApi {}
+
+@endpoint({
+  method: "POST",
+  path: "/widgets"
+})
+class CreateWidget {
+  @request
+  request(
+    /** request body */
+    @body body: WidgetBody
+  ) {}
+
+  /** The widget was created */
+  @response({ status: 201 })
+  successResponse(
+    /** Widget response body */
+    @body body: WidgetBody
+  ) {}
+}
+
+/** Widget request and response body */
+interface WidgetBody {
+  /** price in cents */
+  priceCents: number;
+}
+
+// The contract is otherwise valid; this assignment is the sole type error, so a
+// parse that succeeds here has stopped type-checking the project altogether.
+const malformed: WidgetBody = { priceCents: "free" };
--- a/lib/src/parser.spec.ts
+++ b/lib/src/parser.spec.ts
@@ -26,13 +26,17 @@ describe("parse", () => {
   });
 
   test("rejects a contract whose project has a type error", () => {
-    expect(() => parse(TYPE_ERROR_FIXTURE)).toThrow(
-      /not assignable to type 'number'/
-    );
-    // Diagnostics are joined into one message, so the assertion above matches
-    // on a broken mapping too — the unresolved import simply adds a line
-    // alongside the type error. Ruling that line out is what makes this case
-    // fail-closed: it reports a type error, and only a type error.
-    expect(() => parse(TYPE_ERROR_FIXTURE)).not.toThrow(/Cannot find module/);
+    let message = "";
+    try {
+      parse(TYPE_ERROR_FIXTURE);
+    } catch (e) {
+      message = (e as Error).message;
+    }
+
+    expect(message).toMatch(/not assignable to type 'number'/);
+    // Diagnostics arrive joined, so the assertion above matches a broken
+    // mapping too — the unresolved import adds a line beside the type error.
+    // Ruling it out is what makes this case report a type error and only that.
+    expect(message).not.toMatch(/Cannot find module/);
   });
 });
--- a/lib/src/parser.spec.ts
+++ b/lib/src/parser.spec.ts
@@ -1,0 +1,33 @@
+import * as path from "path";
+import { parse } from "./parser";
+
+const CONTRACT_FIXTURE = path.join(
+  __dirname,
+  "../../test-fixtures/contract/api.ts"
+);
+
+const TYPE_ERROR_FIXTURE = path.join(
+  __dirname,
+  "__spec-examples__/contract-with-type-error.ts"
+);
+
+describe("parse", () => {
+  test("resolves @airtasker/spot without an installed copy beside the contract", () => {
+    // There is no node_modules/@airtasker/spot in this repository, so the
+    // contract's own `import ... from "@airtasker/spot"` can only resolve
+    // through the path mapping Spot configures for the contract project.
+    const contract = parse(CONTRACT_FIXTURE);
+
+    expect(contract.name).toBe("widget-api");
+    expect(contract.endpoints.map(e => e.name).sort()).toEqual([
+      "CreateWidget",
+      "ListWidgets"
+    ]);
+  });
+
+  test("rejects a contract whose project has a type error", () => {
+    expect(() => parse(TYPE_ERROR_FIXTURE)).toThrow(
+      /not assignable to type 'number'/
+    );
+  });
+});
--- a/lib/src/parser.spec.ts
+++ b/lib/src/parser.spec.ts
@@ -29,5 +29,10 @@ describe("parse", () => {
     expect(() => parse(TYPE_ERROR_FIXTURE)).toThrow(
       /not assignable to type 'number'/
     );
+    // Diagnostics are joined into one message, so the assertion above matches
+    // on a broken mapping too — the unresolved import simply adds a line
+    // alongside the type error. Ruling that line out is what makes this case
+    // fail-closed: it reports a type error, and only a type error.
+    expect(() => parse(TYPE_ERROR_FIXTURE)).not.toThrow(/Cannot find module/);
   });
 });
--- a/lib/src/parser.ts
+++ b/lib/src/parser.ts
@@ -1,10 +1,10 @@
-import * as path from "path";
-import { CompilerOptions, Project, ts } from "ts-morph";
+import { Project } from "ts-morph";
 import { Contract } from "./definitions";
 import { parseContract } from "./parsers/contract-parser";
+import { createContractProject } from "./ts-project";
 
 export function parse(sourcePath: string): Contract {
-  const project = createProject();
+  const project = createContractProject();
 
   // Add all dependent files that the project requires
   const sourceFile = project.addSourceFileAtPath(sourcePath);
@@ -19,35 +19,6 @@ export function parse(sourcePath: string): Contract {
   if (result.isErr()) throw result.unwrapErr();
 
   return result.unwrap().contract;
-}
-
-/**
- * Create a new project configured for Spot
- */
-function createProject(): Project {
-  const compilerOptions: CompilerOptions = {
-    target: ts.ScriptTarget.ESNext,
-    module: ts.ModuleKind.CommonJS,
-    strict: true,
-    noImplicitAny: true,
-    strictNullChecks: true,
-    strictFunctionTypes: true,
-    strictPropertyInitialization: true,
-    noImplicitThis: true,
-    resolveJsonModule: true,
-    alwaysStrict: true,
-    noImplicitReturns: true,
-    noFallthroughCasesInSwitch: true,
-    moduleResolution: ts.ModuleResolutionKind.NodeJs,
-    experimentalDecorators: true,
-    baseUrl: "./",
-    paths: {
-      "@airtasker/spot": [path.join(__dirname, "../lib")]
-    }
-  };
-
-  // Creates a new typescript program in memory
-  return new Project({ compilerOptions });
 }
 
 /**
--- a/lib/src/ts-project.ts
+++ b/lib/src/ts-project.ts
@@ -1,17 +1,34 @@
+import * as fs from "fs";
 import * as path from "path";
 import { CompilerOptions, Project, ts } from "ts-morph";
 
 /**
+ * The declarations the `@airtasker/spot` path mapping resolves to: `lib.ts`
+ * when running from source (ts-jest, ts-node), `lib.d.ts` once compiled. Both
+ * are siblings of this file, so the mapping is relative to this module's
+ * directory and must stay so.
+ */
+const SPOT_DECLARATIONS = path.join(__dirname, "lib");
+
+/**
  * The compiler options every Spot contract is parsed and type-checked under.
  *
- * The `@airtasker/spot` path mapping resolves the contract's own import of the
+ * The `@airtasker/spot` path mapping resolves a contract's own import of the
  * package to Spot's copy of the syntax declarations, so a contract type-checks
- * without an installed `node_modules/@airtasker/spot` next to it. It is
- * relative to this module's directory and must stay so: `lib/src/lib.ts` when
- * running from source (ts-jest, ts-node), `build/lib/src/lib.d.ts` once
- * compiled — both siblings of this file.
+ * with no installed `node_modules/@airtasker/spot` beside it.
+ *
+ * TypeScript consults `paths` before `node_modules`, so this mapping also takes
+ * precedence over an installed copy when there is one: a contract is always
+ * type-checked against the declarations of the Spot that is running, never
+ * against a different version installed next to it. That is the intended rule —
+ * the parser and the declarations it checks against are one version — and it
+ * means an older pinned `@airtasker/spot` beside a contract no longer changes
+ * how that contract parses.
+ *
+ * Frozen because `ts-lint` shares this object: a mutation anywhere would change
+ * every subsequent `parse()` in the process.
  */
-export const contractCompilerOptions: CompilerOptions = {
+export const contractCompilerOptions: CompilerOptions = Object.freeze({
   target: ts.ScriptTarget.ESNext,
   module: ts.ModuleKind.CommonJS,
   strict: true,
@@ -28,13 +45,37 @@ export const contractCompilerOptions: CompilerOptions = {
   experimentalDecorators: true,
   baseUrl: "./",
   paths: {
-    "@airtasker/spot": [path.join(__dirname, "lib")]
+    "@airtasker/spot": [SPOT_DECLARATIONS]
   }
-};
+});
+
+/**
+ * Fail if the mapping target does not exist.
+ *
+ * TypeScript treats `paths` as a hint: an entry naming a file that is not there
+ * falls back to node resolution with no diagnostic. So a mapping broken by a
+ * change to the emitted layout would not report as a broken mapping — a
+ * contract would resolve against an installed copy, or fail with a bare
+ * "Cannot find module", depending on what happens to sit beside it. Checking
+ * the target directly is what turns that into an error naming the real cause.
+ */
+function assertSpotDeclarationsResolve(): void {
+  const candidates = [".ts", ".d.ts"].map(ext => `${SPOT_DECLARATIONS}${ext}`);
+  if (candidates.some(candidate => fs.existsSync(candidate))) {
+    return;
+  }
+  throw new Error(
+    `Spot's own @airtasker/spot declarations are missing: expected ` +
+      `${candidates.join(" or ")}. Contracts cannot be type-checked against ` +
+      `the running Spot without them. This usually means the emitted layout ` +
+      `moved — check outDir and rootDir in tsconfig.json.`
+  );
+}
 
 /**
  * Create an in-memory TypeScript project configured for Spot contracts.
  */
 export function createContractProject(): Project {
+  assertSpotDeclarationsResolve();
   return new Project({ compilerOptions: contractCompilerOptions });
 }
--- a/lib/src/ts-project.ts
+++ b/lib/src/ts-project.ts
@@ -1,34 +1,20 @@
-import * as fs from "fs";
 import * as path from "path";
 import { CompilerOptions, Project, ts } from "ts-morph";
 
 /**
- * The declarations the `@airtasker/spot` path mapping resolves to: `lib.ts`
- * when running from source (ts-jest, ts-node), `lib.d.ts` once compiled. Both
- * are siblings of this file, so the mapping is relative to this module's
- * directory and must stay so.
- */
-const SPOT_DECLARATIONS = path.join(__dirname, "lib");
-
-/**
  * The compiler options every Spot contract is parsed and type-checked under.
  *
- * The `@airtasker/spot` path mapping resolves a contract's own import of the
- * package to Spot's copy of the syntax declarations, so a contract type-checks
- * with no installed `node_modules/@airtasker/spot` beside it.
+ * The `@airtasker/spot` mapping resolves a contract's own import of the package
+ * to Spot's copy of the syntax declarations — `lib.ts` from source, `lib.d.ts`
+ * once compiled, both siblings of this file — so a contract type-checks with no
+ * installed `node_modules/@airtasker/spot` beside it. TypeScript consults
+ * `paths` before `node_modules`, so an installed copy does not take precedence
+ * over the running Spot's declarations for this specifier.
  *
- * TypeScript consults `paths` before `node_modules`, so this mapping also takes
- * precedence over an installed copy when there is one: a contract is always
- * type-checked against the declarations of the Spot that is running, never
- * against a different version installed next to it. That is the intended rule —
- * the parser and the declarations it checks against are one version — and it
- * means an older pinned `@airtasker/spot` beside a contract no longer changes
- * how that contract parses.
- *
- * Frozen because `ts-lint` shares this object: a mutation anywhere would change
- * every subsequent `parse()` in the process.
+ * `skipLibCheck` because a declaration file's own errors are not the contract
+ * author's to fix and no `--fix` can reach them.
  */
-export const contractCompilerOptions: CompilerOptions = Object.freeze({
+export const contractCompilerOptions: CompilerOptions = {
   target: ts.ScriptTarget.ESNext,
   module: ts.ModuleKind.CommonJS,
   strict: true,
@@ -41,41 +27,18 @@ export const contractCompilerOptions: CompilerOptions = Object.freeze({
   alwaysStrict: true,
   noImplicitReturns: true,
   noFallthroughCasesInSwitch: true,
+  skipLibCheck: true,
   moduleResolution: ts.ModuleResolutionKind.NodeJs,
   experimentalDecorators: true,
   baseUrl: "./",
   paths: {
-    "@airtasker/spot": [SPOT_DECLARATIONS]
+    "@airtasker/spot": [path.join(__dirname, "lib")]
   }
-});
-
-/**
- * Fail if the mapping target does not exist.
- *
- * TypeScript treats `paths` as a hint: an entry naming a file that is not there
- * falls back to node resolution with no diagnostic. So a mapping broken by a
- * change to the emitted layout would not report as a broken mapping — a
- * contract would resolve against an installed copy, or fail with a bare
- * "Cannot find module", depending on what happens to sit beside it. Checking
- * the target directly is what turns that into an error naming the real cause.
- */
-function assertSpotDeclarationsResolve(): void {
-  const candidates = [".ts", ".d.ts"].map(ext => `${SPOT_DECLARATIONS}${ext}`);
-  if (candidates.some(candidate => fs.existsSync(candidate))) {
-    return;
-  }
-  throw new Error(
-    `Spot's own @airtasker/spot declarations are missing: expected ` +
-      `${candidates.join(" or ")}. Contracts cannot be type-checked against ` +
-      `the running Spot without them. This usually means the emitted layout ` +
-      `moved — check outDir and rootDir in tsconfig.json.`
-  );
-}
+};
 
 /**
  * Create an in-memory TypeScript project configured for Spot contracts.
  */
 export function createContractProject(): Project {
-  assertSpotDeclarationsResolve();
   return new Project({ compilerOptions: contractCompilerOptions });
 }
--- a/lib/src/ts-project.ts
+++ b/lib/src/ts-project.ts
@@ -1,0 +1,40 @@
+import * as path from "path";
+import { CompilerOptions, Project, ts } from "ts-morph";
+
+/**
+ * The compiler options every Spot contract is parsed and type-checked under.
+ *
+ * The `@airtasker/spot` path mapping resolves the contract's own import of the
+ * package to Spot's copy of the syntax declarations, so a contract type-checks
+ * without an installed `node_modules/@airtasker/spot` next to it. It is
+ * relative to this module's directory and must stay so: `lib/src/lib.ts` when
+ * running from source (ts-jest, ts-node), `build/lib/src/lib.d.ts` once
+ * compiled — both siblings of this file.
+ */
+export const contractCompilerOptions: CompilerOptions = {
+  target: ts.ScriptTarget.ESNext,
+  module: ts.ModuleKind.CommonJS,
+  strict: true,
+  noImplicitAny: true,
+  strictNullChecks: true,
+  strictFunctionTypes: true,
+  strictPropertyInitialization: true,
+  noImplicitThis: true,
+  resolveJsonModule: true,
+  alwaysStrict: true,
+  noImplicitReturns: true,
+  noFallthroughCasesInSwitch: true,
+  moduleResolution: ts.ModuleResolutionKind.NodeJs,
+  experimentalDecorators: true,
+  baseUrl: "./",
+  paths: {
+    "@airtasker/spot": [path.join(__dirname, "lib")]
+  }
+};
+
+/**
+ * Create an in-memory TypeScript project configured for Spot contracts.
+ */
+export function createContractProject(): Project {
+  return new Project({ compilerOptions: contractCompilerOptions });
+}
--- a/test-fixtures/contract/api.ts
+++ b/test-fixtures/contract/api.ts
@@ -1,0 +1,87 @@
+import {
+  api,
+  body,
+  DateTime,
+  defaultResponse,
+  endpoint,
+  Float,
+  headers,
+  pathParams,
+  request,
+  response,
+  securityHeader,
+  String
+} from "@airtasker/spot";
+import "./list-widgets";
+import { ErrorBody, WidgetBody } from "./models";
+
+/** A widget catalogue. */
+@api({ name: "widget-api" })
+class WidgetApi {
+  @securityHeader
+  "x-auth-token": String;
+}
+
+/** Creates a widget in a catalogue */
+@endpoint({
+  method: "POST",
+  path: "/catalogues/:catalogueId/widgets",
+  tags: ["Widget"]
+})
+class CreateWidget {
+  @request
+  request(
+    @pathParams
+    pathParams: {
+      /** catalogue identifier */
+      catalogueId: String;
+    },
+    @headers
+    headers: {
+      /** Auth header */
+      "x-auth-token": String;
+    },
+    /** request body */
+    @body body: CreateWidgetRequestBody
+  ) {}
+
+  /** The widget was created */
+  @response({ status: 201 })
+  successResponse(
+    @headers
+    headers: {
+      /** Location of the created widget */
+      Location: String;
+    },
+    /** Widget response body */
+    @body body: WidgetBody
+  ) {}
+
+  /** The request was malformed */
+  @response({ status: 400 })
+  badRequestResponse(
+    /** Error response body */
+    @body body: ErrorBody
+  ) {}
+
+  @defaultResponse
+  unexpectedResponse(
+    /** Error response body */
+    @body body: ErrorBody
+  ) {}
+}
+
+/** Widget creation request body */
+interface CreateWidgetRequestBody {
+  /** data wrapper */
+  data: {
+    /** display name */
+    name: String;
+    /** price in cents */
+    priceCents: Float;
+    /** when the widget becomes orderable */
+    availableFrom?: DateTime;
+    /** free-form labels */
+    tags: String[];
+  };
+}
--- a/test-fixtures/contract/list-widgets.ts
+++ b/test-fixtures/contract/list-widgets.ts
@@ -1,0 +1,54 @@
+import {
+  body,
+  endpoint,
+  headers,
+  pathParams,
+  queryParams,
+  request,
+  response,
+  String
+} from "@airtasker/spot";
+import { ErrorBody, WidgetListBody } from "./models";
+
+/** Lists the widgets in a catalogue */
+@endpoint({
+  method: "GET",
+  path: "/catalogues/:catalogueId/widgets",
+  tags: ["Widget"]
+})
+class ListWidgets {
+  @request
+  request(
+    @pathParams
+    pathParams: {
+      /** catalogue identifier */
+      catalogueId: String;
+    },
+    @headers
+    headers: {
+      /** Auth header */
+      "x-auth-token": String;
+    },
+    @queryParams
+    queryParams: {
+      /** only return widgets carrying this tag */
+      tag?: String;
+      /** maximum number of widgets to return */
+      limit?: number;
+    }
+  ) {}
+
+  /** The matching widgets */
+  @response({ status: 200 })
+  successResponse(
+    /** Widget collection response body */
+    @body body: WidgetListBody
+  ) {}
+
+  /** The catalogue does not exist */
+  @response({ status: 404 })
+  notFoundResponse(
+    /** Error response body */
+    @body body: ErrorBody
+  ) {}
+}
--- a/test-fixtures/contract/models.ts
+++ b/test-fixtures/contract/models.ts
@@ -1,0 +1,33 @@
+/** A widget */
+export interface Widget {
+  /** widget identifier */
+  id: string;
+  /** display name */
+  name: string;
+  /** price in cents */
+  priceCents: number;
+  /** whether the widget is orderable */
+  available: boolean;
+  /** free-form labels */
+  tags: string[];
+}
+
+/** Widget response body */
+export interface WidgetBody {
+  /** data wrapper */
+  data: Widget;
+}
+
+/** Widget collection response body */
+export interface WidgetListBody {
+  /** data wrapper */
+  data: Widget[];
+}
+
+/** Error body */
+export interface ErrorBody {
+  /** error name */
+  name: string;
+  /** error messages */
+  messages: string[];
+}


timdawborn:
# Stack review — COMPASS-17 (#2708 → #2714)

Automated multi-agent review of all seven stacked PRs (`/pr-review-toolkit:review-pr`): 14 agents across general review, test coverage, error handling, type design and comment accuracy. Each PR was reviewed against **its own base**, so findings are scoped to that PR's diff.

Findings marked **[verified]** were reproduced by building the branch and running the code — mutation tests, sentinel-package experiments, `--prod` installs, piped-stdout runs — not inferred from reading.

This is a genuinely strong stack. The PR bodies are the most thorough I've reviewed in this repo, and several findings below exist *because* the bodies made falsifiable claims precise enough to test. The two Critical items are both "a gate that cannot fail", which is the failure mode this stack is explicitly trying to eliminate.

---

## Summary

| PR | Critical | Important | Headline |
|---|---|---|---|
| #2708 | 1 | 2 | Path mapping now **overrides** a consumer's installed copy; body says no behaviour change |
| #2709 | 1 | 4 | `.js` files silently lose `no-undef` + 15 core rules |
| #2710 | – | 2 | The new smoke test doesn't test the property the PR fixes |
| #2711 | 2 | 7 | `ts-lint` on an empty dir exits 0; report truncated when piped |
| #2712 | – | 2 | TTY predicate can be deleted with the suite green |
| #2713 | – | 4 | `AGENTS.md` documents a ghcr publish that doesn't exist yet |
| #2714 | – | 5 | README pins `2.1.0`; arm64 never parity-checked |

**Blocking before merge:** the two #2711 Criticals, the #2709 Critical, and the cross-cutting `engines` conflict.

---

## Cross-cutting

**C1. `engines.node: ">=18.12.0"` is false once the lint stack becomes a runtime dependency.** [verified] — *two agents, independently*

`package.json` at #2711 declares `engines: {"node": ">=18.12.0"}` while shipping as **`dependencies`**:

- `eslint-plugin-jest@29.16.1` → `engines: {node: ^20.12.0 || ^22.0.0 || >=24.0.0}`

As devDependencies (#2709) this never reached consumers. As runtime dependencies (#2711) it does. A consumer on Node 18 — a version this package still advertises — gets `EBADENGINE`, or a hard failure under `engine-strict`. Meanwhile `test-node-18` still runs in the matrix, and the `lint-check` job's `setup-node` has no `node-version:` input, so it silently inherits the runner default.

Either bump `engines` to `>=20.19.0` and drop 18 from the matrix (a breaking change, so it wants a major), or keep the lint stack out of `dependencies`. This needs a decision before the stack merges — it's the one finding that breaks installs for existing external npm users.

**C2. Every PR body says CI won't report. It does.**

All six stacked bodies state `build-and-test` is `on: pull_request: branches: [master]` and "will not report here until the stack merges". In fact all seven PRs have green `test-node-{18,20,22,24}` + `lint-check`, and #2713/#2714 have green `docker-parity`. Only `dependency-review` fails (pre-existing repo-settings issue, correctly diagnosed in #2709).

Worth correcting, because reviewers were told to discount signal that actually ran — and because #2710's smoke-test step and #2713's parity gate *did* execute, which is useful evidence in their favour.

**C3. `docker-parity` is added but never promoted to a required status check.**

The only required check on `master` is `test-summary-publish`. #2713 adds `docker-parity` and wires it into `publish.needs`, which gates the *release* but not *merges*. Per `pull-requests.md` → "Introducing a new required CI check": a later PR breaks the Dockerfile, `docker-parity` goes red on `master`, nothing blocks that merge, and every subsequent PR inherits a red result until a release fails. Ship the branch-protection change, or add a "Steps to merge" item naming the owner and ticket and stating the check is advisory until promoted. (No job triggers on `merge_group`; confirm there's no merge queue rather than assuming.)

**C4. arm64 ships having never been executed.** [verified]

`docker-parity` builds `load: true` with no `platforms:` — amd64 only. `docker-publish` builds `linux/amd64,linux/arm64` under QEMU. The arm64 manifest's first ever build is the release run. The two reasons #2714's own comment gives for arm64 mattering (Apple Silicon engineers, bff-client's arm64 devcontainer runner) are exactly who receives an unverified image. Minimum: build multi-arch in `docker-parity` with `push: false` so the emulated build is exercised per-PR.

---

## #2708 — resolve `@airtasker/spot` from spot's own tree

**Important — the mapping now overrides a consumer's installed copy, and the body claims the opposite.** [verified] — *two agents, independently, with sentinel packages*

The body states: *"**No behaviour change for npm consumers.** A consumer that has `@airtasker/spot` installed resolves the mapping to the same declarations it was already getting from `node_modules`."*

Both agents built a consumer directory with a divergent `node_modules/@airtasker/spot` and parsed under each mapping:

| mapping | result |
|---|---|
| old `../lib` (never existed) | parses — **consumer's `node_modules` copy won** |
| new `lib` (exists) | `has no exported member 'SENTINEL…'` — **spot's own copy won** |

TypeScript consults `paths` *before* `node_modules` whenever the mapped target exists. On master the target never existed, so it always fell through. This PR makes it exist, so `node_modules` is now never consulted.

This is very likely the *correct* semantics — the parser and the declarations it type-checks against should be one version, and the container work requires it. The problem is that it's undocumented and actively denied. Concretely: `npx @airtasker/spot@latest` against a repo pinned to an older devDependency now type-checks contracts against the newer declarations, and since `parse()` throws on any diagnostic, one removed export hard-fails `generate`, `lint`, `validate` and `mock`.

Please rewrite that paragraph to state the intended rule (the running spot's declarations always win), and add a sentence to the `ts-project.ts` docstring naming the precedence — it currently describes only the fill-in half.

**Important — the compiled layout is asserted in a docstring and tested nowhere.**

The docstring claims a two-layout invariant (`lib/src/lib.ts` from source, `build/lib/src/lib.d.ts` compiled). ts-jest only ever exercises the first. `tsconfig.json` has `rootDir` commented out, so the emit layout is *inferred* from the input set — any `include` change shifts it, `path.join(__dirname, "lib")` names a directory that no longer exists, `pnpm test` stays green, and every published CLI and container invocation fails with the exact error this PR fixes. #2712 adds a second consumer of `contractCompilerOptions`, growing the blast radius.

One line in CI after `pnpm build` closes it:
```
node -e 'const{contractCompilerOptions:o}=require("./build/lib/src/ts-project");const p=o.paths["@airtasker/spot"][0];if(!require("fs").existsSync(p+".d.ts"))throw new Error("mapping broken in build layout: "+p)'
```

**Suggestions**
- `parser.spec.ts:29` asserts by substring on a joined diagnostic string, so it passes under *both* mappings [verified] — the broken mapping just adds a `Cannot find module` line alongside the expected one. Add `expect(...).not.toThrow(/Cannot find module/)`.
- `contractCompilerOptions` is a shared mutable object handed by reference to every `Project`; with #2711 adding a second consumer, `Object.freeze` is cheap now and awkward later.
- Fixtures now live in both `__spec-examples__/` and `test-fixtures/` with no recorded rule for which is which. Worth a line in `CLAUDE.md`.

**Positive:** the fail-before/pass-after test is real and mutation-verified; on master *zero* tests exercised contract-import resolution, so this is net-new coverage of a path that was silently broken.

---

## #2709 — eslint flat config

**Critical — `.js` files silently lose `no-undef` and 15 other core rules.** [verified]

`eslint.config.js:25-26` applies `extends: [js.configs.recommended, tseslint.configs.recommended]` to `files: ["**/*.js", "**/*.ts", "**/*.tsx"]`. `tseslint.configs.recommended` bundles the `eslint-recommended` block that turns off core rules TypeScript already enforces — upstream scopes that to `**/*.{ts,tsx,mts,cts}`, but `tseslint.config()`'s `extends` **overwrites the extended config's own `files` with the parent's**. So those disables now apply to `.js` too.

Proven with a root `probe.js` containing `module.exports = someUndefinedGlobal;`:
- parent branch: `error 'someUndefinedGlobal' is not defined  no-undef`
- this branch: **no output**

Lost on `.js`: `no-undef`, `no-const-assign`, `no-dupe-args`, `no-dupe-keys`, `no-dupe-class-members`, `no-func-assign`, `no-import-assign`, `no-obj-calls`, `no-redeclare`, `no-setter-return`, `no-this-before-super`, `no-unreachable`, `no-unsafe-negation`, `getter-return`, `constructor-super`, `no-new-symbol`.

Exposure today is one file, which is why it won't be caught — but this stack adds JS surface downstream. Fix: split the block so `.js` gets `js.configs.recommended` only.

**Important — `no-non-null-assertion` is dropped, and the body's baseline is wrong.**

The rule was in typescript-eslint v4 `recommended` (warn); v8 moved it to `strict`, so nothing enables it now. That's 23 live warnings disappearing: `example-parser.spec.ts` (8), `schemaprop-parser.spec.ts` (14), `types.ts:622` (1).

The body claims *"exactly the same 2 `no-explicit-any` warnings the old stack reported"*. The old stack reported **25** — and #2708's own body says so ("25 pre-existing warnings, 0 errors"). The two bodies in this stack contradict each other, and that inaccurate baseline is what let the drop through. Relatedly, the removed directive at `types.ts:588` is described as "not even reachable under the old one" — it was reachable and was actively suppressing.

**Important — `no-empty-object-type: "off"` also drops `no-empty-interface`.** The translation table treats `ban-types` → three rules as one-for-one. v8's `no-empty-object-type` *also* replaces `no-empty-interface`, which was on in v4 `recommended` and was **not** covered by the old `ban-types: "off"`. `interface Foo extends Bar {}` now lints clean repo-wide — a real footgun in a contract codebase. Faithful translation is `["error", { allowObjectTypes: "always" }]`.

**Important — see C1** (`engines` / Node 18).

**Important — deferred-decision comment with no follow-up ticket.** `eslint.config.js:40-41`: *"Tightening it to an error is a separate piece of work from this migration."* Per `pull-requests.md`, this needs its own ticket inline, and not COMPASS-17 (which closes when the stack merges). Same applies if `no-non-null-assertion` is deliberately left off.

**Suggestions**
- `ignores: ["build/"]` anchors to the config base path, so it no longer matches nested `build/` dirs the way `.eslintignore`'s bare `build` did [verified: 282 → 283 files with a `lib/build/inner.js` present]. Use `"**/build/"`. Note the body's "282 files each" check couldn't have caught this — a fresh checkout has no `build/`.
- Jest rules narrowed from repo-wide to `**/*.spec.ts` — defensible, but a fourth deviation the "one-for-one apart from three changes" table doesn't list.
- Several `"off"` entries are now no-ops (`no-empty-function`, `explicit-function-return-type` are in neither v4 nor v8 `recommended`); the PR removed matching dead `eslint-disable` comments but kept these.

**Positive:** verifying file coverage with a real `eslint -f json` diff rather than assertion is exactly the right instinct for this migration, and the finding above is a refinement of it, not a refutation. Correctly established that no TypeScript bump is forced.

---

## #2710 — declare oclif runtime deps

**Important — the regression guard doesn't guard the property the PR fixes.** [verified]

The smoke test runs after `prepare-repository`'s **full** `pnpm install --frozen-lockfile`. Under pnpm's isolated layout both `dependencies` and `devDependencies` are symlinked into `node_modules/`, so the step proves "declared *somewhere*", not "declared as a runtime dependency" — and the latter is the whole point (title: "needs at **runtime**"; motivation: the image runs `--prod`).

Reproduced by moving both packages back to `devDependencies`:
```
$ pnpm install --frozen-lockfile && pnpm build && node bin/run --help
VERSION @airtasker/spot/2.0.3 …          # step is GREEN

$ pnpm install --frozen-lockfile --prod && node bin/run --help
node:internal/modules/cjs/loader:1522
  throw err;                              # dead on every invocation
```

Fix — make the guard install the way the image does (`typescript` is already a prod dep, so build first then prune):
```yaml
- run: |-
    set -uex
    pnpm install --frozen-lockfile --prod
    node bin/run --help
    node bin/run validate test-fixtures/contract/api.ts
    pnpm install --frozen-lockfile   # restore devDeps for jest
```

**Important — the step comment claims more than the check enforces.** It reads as "this reproduces the docker image's failure mode"; it reproduces only the isolated-layout half. Also "and the docker image built from it" is a forward reference to something two PRs away with no navigable referent.

**Suggestion:** the step runs 16× (4 nodes × 4 shards); only the Node axis carries signal. `if: inputs.shard == '1'`.

**Positive:** the diagnosis is correct and non-obvious, and the body is honest about why 529 green tests were never evidence. An agent independently confirmed the `@oclif/help` half is genuinely load-bearing (its own `lib/util.js` re-resolves `@oclif/help` rooted at the consuming CLI's package dir), and that neither package is over-declared. Every reproducible claim in the body checked out, including the `checksum` digest.

---

## #2711 — `ts-lint` command

**Critical — a directory with no `.ts` files exits 0.** [verified] — *four agents, independently*

`lib/src/ts-lint/ts-lint.ts:37-42` returns `ok: true`, and the CLI prints both lines and exits 0:
```
$ node bin/run ts-lint /tmp/emptydir
No TypeScript files found under /tmp/emptydir
No problems found
EXIT=0
```
This command is about to become the contract gate for three consumer repos via a container invocation with a hardcoded path. The day someone renames `spots/` → `contracts/`, or a bind-mount lands empty, or a sparse checkout omits the tree, CI is permanently green and nothing is ever checked again. Note the asymmetry: a *missing* directory correctly fails (ENOENT); the more likely real-world variant is the softer one. Return `ok: false`, or require an explicit `--allow-empty`.

**Critical — `process.exit(1)` truncates the report whenever stdout is a pipe.** [verified] — *three agents*

`this.log()` writes are asynchronous when stdout is a pipe; `process.exit` kills the process before the buffer drains. Three identical runs on a 40-file tree producing a 1601-line report:
```
$ node bin/run ts-lint $BIG | cat > out
run1 lines=770    run2 lines=1601    run3 lines=577
```
Every consumer invocation in this stack's target environment is a container in CI, i.e. a pipe. The exit code is right so CI still goes red, but the contributor sees an arbitrary prefix, fixes those, and the next run surfaces "new" errors that were always there — and the header lies (`Lint (1600):` with 700 lines following). Fix: `process.exitCode = 1` (verified: 1601/1601/1601 and exit code preserved).

**Important — the test guarding the design's central claim is vacuous on its ESLint half.** [verified]

The body says of `overrideConfigFile: true`: *"That is the load-bearing claim of the in-code-config design, so it has a test rather than a comment."* The Prettier half holds. The ESLint half doesn't — running the clean fixture plus the test's own `eslint.config.mjs` with `overrideConfigFile: true` **removed** gives an identical result (0 messages both ways), because the leftover config sets `no-undef: "error"` and `tseslint.configs.recommended` inside `overrideConfig` turns `no-undef` off again. Delete that line from `ts-lint.ts:130` and the suite stays green. Discriminating fixture (verified 0 vs 5 messages):
```js
export default [{ rules: { 'no-restricted-syntax': ['error',
  { selector: 'TSInterfaceDeclaration', message: 'leftover config was read' }] } }];
```

**Important — five of the eight rule disablements are dead to the fixtures.** [verified]

Flipping each `"off"` → `"error"` in turn: `no-empty-function`, `explicit-function-return-type`, `no-unused-vars` trip; `no-inferrable-types`, `no-use-before-define`, `no-empty-object-type`, `no-wrapper-object-types`, `no-unsafe-function-type` **do not**.

`no-wrapper-object-types` is the sharp one: the body singles it out as the disablement that "matters, because `String` and `Boolean` are Spot's own contract vocabulary", and the code comment says the same — but turned on against the clean fixture (which imports and uses `String` in path params, headers and query params) it reports **nothing**. The import shadows the global, so the rule may never see a wrapper type at all. Per the repo's own rule on invariants named in comments, this needs a fixture that isolates it or a corrected comment. Separately, deleting `js.configs.recommended` and `eslint-config-prettier` entirely also leaves the suite green — the whole ESLint surface exercised is one rule at one line.

**Important — a syntax error in one file discards every finding from the other steps, and under `--fix` leaves a partially rewritten tree.** [verified] `format()` from `prettier/standalone` throws on unparseable input at `ts-lint.ts:95`, after ESLint has completed and before the type check runs. No catch, so the ESLint report is thrown away, the type check never runs, and oclif prints a raw stack. The error names no file (no `filepath` in `prettierConfig`) — on a 345-file tree you get `SyntaxError: ',' expected. (12:7)` and bisect by hand. Under `--fix`, files already written stay written and `fixed` never reaches the caller. This directly defeats the PR's headline invariant that every step runs.

**Important — the type step isn't bounded by the walked file list.** [verified] `collectContractFiles` documents skipping `node_modules`; `runTypeCheck` re-imports it via `resolveSourceFileDependencies()`, and `contractCompilerOptions` doesn't set `skipLibCheck`. A `node_modules/junk/index.d.ts` with a bad declaration produces `Types (2): …/node_modules/junk/index.d.ts:2 - …` and `ok: false`. A consumer fails on a file they didn't ask to check, can't edit, and `--fix` can't touch.

**Important — the documented ESLint-before-Prettier ordering isn't load-bearing in any test.** [verified] Swapping the two calls leaves all six tests passing, because neither fixture contains a single auto-fixable ESLint error (`output === undefined` for every file in both trees). The comment states the invariant precisely — it just needs a fixture: one auto-fixable error whose fix produces code needing re-layout.

**Important — under `--fix`, lint `line:col` point at pre-Prettier text.** [verified] ESLint messages are computed before Prettier rewrites the same files. Reported at line 5, actually at line 3. Editor and CI error parsers jump to the wrong place, compounding across a large tree.

**Important — no test of the CLI exit code.** There is no spec under `cli/src/commands/` for `ts-lint`. `process.exit(1)` on findings and 0 on clean is the entire contract CI consumes, verified only by hand in the body. Spawn `bin/run ts-lint` against both fixtures and assert 0 / 1.

**Suggestions**
- The warning path is unreachable (no rule is `"warn"`) and self-contradictory if reached: it would print `Lint (2): … warning …` then `No problems found` and exit 0.
- `fixed: Set<string>` is a mutable out-parameter threaded into two helpers; `fixed` can be non-empty with `fix: false` — an invalid state only convention prevents.
- Dot-directories are skipped, so a contract in `.generated/` passes by being invisible — contradicting the stated "must not pass by being invisible" principle.
- `test-fixtures/ts-lint/clean/` duplicates `test-fixtures/contract/` byte-for-byte modulo quote style, with nothing recording the relationship.
- `ts-lint.ts:163` docstring: "a tree that passes here is a tree `generate` and `validate` can read" — `parse` also runs `parseContract`, which returns its own errors (missing `@api`, unsupported type). Type-clean ≠ readable.

**Positive:** `lib/src/types.ts` and both generator specification files are **provably** behaviour-preserving — `diff -rq build/` against the parent branch is empty except the new files, which is the strongest evidence available for a 200-line reformat across public API. `LiteralType` is a whitespace-only union re-wrap; no major version is warranted by type changes. Every runtime `require` is a declared `dependency` with satisfied peers, so the #2710 trap does not recur. Three claims most likely to be asserted-and-untested each have an isolating test, and the Prettier config options are genuinely load-bearing (mutating `singleQuote`, `tabWidth`, `semi`, `trailingComma`, `printWidth`, `bracketSpacing` all break the clean fixture).

---

## #2712 — fail fast without a terminal

**Important — the TTY predicate can be deleted and the suite stays green.** [verified by mutation] — *two agents*

All four `generate` tests call `withoutTerminal()` first; nothing ever sets `isTTY` truthy. Mutating `if (!process.stdin.isTTY)` → `if (true)` leaves all 4 passing in 6.2s. That mutation destroys every interactive invocation — the prompting the body explicitly promises is "unchanged" — and ships green. Inversion *is* caught, but only by a 60s timeout (suite 6s → 184s) with a diagnostic naming no invariant. One test with `isTTY = true` and a mocked inquirer collapses that to a ~1s assertion.

**Important — `validation-server.ts:44-47` declares a cross-repo string contract with no enforcement.** The comment says rewording `Validation server running` hangs bff-client's Gradle `ExecFork` until timeout with no other signal. Nothing in this repo pins the string; the only thing that does is a `grep` in #2713's parity script — a *different* PR. If #2712 lands and #2713 slips, a style-motivated reword ships green here and breaks a downstream build silently. Hoist to an exported constant and assert it. (This is precisely the "if you change X, Z breaks" → that's a spec for a test rule.)

**Suggestions**
- `README.md:67` and `:216` still show `npx @airtasker/spot generate --contract api.ts` — correct interactively, but #2714 reframes the README around Docker where it now exits 2, and #2714 doesn't touch those lines. `--help` doesn't mention the flags are required non-interactively either.
- The guard comment says "hangs or exits on a silently empty answer"; inquirer 8 sets `skipTTYChecks: true`, so at EOF it hangs. State the one verified mode.
- Naming bff-client's ExecFork task is provenance shape — state the invariant ("machine-parsed readiness signal") and keep the consumer as an example.

**Positive — the predicate is right, and this was the main risk.** An agent read `inquirer@8.2.6`'s source: it defaults `input` to `process.stdin` and sets `skipTTYChecks: true`, so guarding on `process.stdin.isTTY` tests the actual precondition, not the `stdout` correlate. No fallback-to-default (which would have turned a hang into silently wrong output — the `-o` prompt carries `default: "."`), exit really is non-zero end to end via `@oclif/errors/handle`, the message names every missing flag in one pass, and the shared-code sweep holds: `inquirer` is imported by exactly one command, and the only other `outputFile` caller (`init.ts`) passes bare filenames so `path.join` → `path.resolve` is identical. The all-flags-supplied test is the only thing stopping an over-aggressive guard — worth renaming so nobody deletes its `withoutTerminal()` call as scaffolding.

---

## #2713 — docker image and parity gate

**Important — `AGENTS.md:38` states a behaviour the tree does not have.** [verified] *"Both distributions are published from one GitHub Release event, so a version means the same thing on npm and on ghcr."* At this commit `ghcr` appears nowhere in the workflow — the publish arrives in #2714. Two more sites in the same PR: the `docker-parity` comment "Multi-arch … is covered on publish" (nothing covers it at this commit), and the `Dockerfile:8-10` usage example running an unpublished image with no note that it will be private. Move these to #2714, or restate them as forward-looking.

**Important — the `Dockerfile:47-49` comment's stated mechanism is wrong.** [verified] It says `tsc` is used "rather than `pnpm build`, to stay off the `prepack` script". But `"build": "tsc"` — `pnpm build` and `pnpm exec tsc` are the same command, and `prepack` is a pack/publish lifecycle script that `pnpm build` never triggers. The *fact* is right (you don't want `prepack`'s webpack/redoc bundle); the mechanism named is not. A maintainer who "simplifies" this to `pnpm build` will think they introduced a webpack dependency they didn't.

**Important — `compare_run` never asserts the expected exit status.** [verified] — *two agents*

It compares CLI-vs-image only. It **does** catch the `HOME` regression today (image 0, CLI 2 → mismatch). The weakness is durability: if the fixture stops failing on *both* sides — someone edits `contract-with-type-error.ts`, which is shared with #2708's `parser.spec.ts`, or `validate` stops type-checking — both exit 0, the comparison passes, and the case silently degrades into a seventh always-succeeding check while its comment still claims it covers the oclif error path. Add an expected-status parameter.

**Important — see C3** (`docker-parity` not a required check).

**Suggestions**
- No `USER` instruction. `WORKDIR /workspace` is created `root:root 0755`, so under the mandated `--user "$(id -u):$(id -g)"` the documented default workdir is unwritable; without `--user` you get exactly the root-owned output the header says breaks rails-monolith and bff-client. The parity script always passes both flags, so this gap is invisible to the gate.
- Base image pinned by mutable tag, not digest — two builds of the same commit can differ.
- The generator-coverage `sed` uses `[a-z0-9-]`, which covers all four current keys (`raw`, `json-schema`, `openapi2`, `openapi3`) but would silently miss a future `"openapi3.1"` — it wouldn't enter `DECLARED`, `COVERED` also wouldn't have it, the strings still compare equal, and the gate passes. Widen the class, or state the restriction in the comment.
- The detached `validation-server` container isn't in the `EXIT` trap; port 5600 is hardcoded, so an unrelated local process on 5600 makes `curl` report a false OK.
- The `node_modules`-on-any-parent walk exits before checking `/`, so `/node_modules` is never tested — narrow the comment or check it.
- `rm -f build/cli/src/commands/{mock,docs,init}.*` can't fail, including when it matches nothing; if the emit layout moves, the image silently ships the three commands it documents as absent.
- `AGENTS.md:61` "no manifest shipped" is true of the image but stated repo-wide — the npm package *does* ship `oclif.manifest.json` via `prepack`.
- `AGENTS.md:29` documents `ts-lint spots --fix` under the description "Check…" — `--fix` writes.
- `.dockerignore:4` justifies a committed allowlist by `.claude/worktrees/`, which is untracked and developer-local.

**Positive:** `set -ue -o pipefail` present; the empty-parse trap at `:68` is explicitly closed with its reasoning recorded; self-containment asserts its own precondition *before* running so it can't pass vacuously, and checks both exit code and produced artifact; `curl -fsS` catches a listening-but-broken server; the absent-command check asserts a message and not merely non-zero; both `|| true`s discard genuinely signal-free exit codes. The `.dockerignore` allowlist matches the `COPY` set exactly (all nine `!` patterns verified). Version alignment across `.tool-versions` / Dockerfile / `packageManager` is exact and documented as coupled. Every other `AGENTS.md` path and command claim verified against the tree, including the `parser/` → `parsers/` correction. The `HOME=/tmp` discovery, and the honesty about why the *first* version of the gate missed it, is the best thing in this stack.

---

## #2714 — publish on release, docker-first README

**Important — the README pins an image tag that will not exist.** `package.json` is `2.0.3`; the README's headline and validation-server commands say `ghcr.io/airtasker/spot:2.1.0`. The first thing a reader copies out of the public README fails with `manifest unknown`, and it re-stales every release. #2713's Dockerfile header already gets this right with `:<version>`.

**Important — npm and ghcr can diverge, and the README asserts they can't.** `publish` and `docker-publish` are siblings with no ordering. If the multi-arch build fails — per C4, the step most likely to — npm has `2.1.0` and ghcr doesn't. The README says the one event publishes all three "so a version means the same thing everywhere". One event *triggering* both is not both *succeeding*. Either `publish: needs: docker-publish`, or soften the README and document recovery.

**Important — see C4** (arm64 unverified).

**Important — the public README's "recommended" path is a private package.** Spot is a public repo with external consumers and an npm downloads badge. The README calls Docker "the recommended way to run Spot" and then says the package is private and needs a `read:packages` token — an external user gets `denied`. Nothing keeps it private either: the first push from a public repo creates a **public** package, the flip is a manual rollout step, and no automated check re-asserts it. Say plainly that the image is Airtasker-internal and npm is the external distribution.

**Important — the Docker Hub login guard was copied from a job where it means something.** In `docker-parity` (#2713) `if: env.DOCKERHUB_TOKEN != ''` correctly means "fork PR, no secrets", and the comment says so. `docker-publish` runs only on `release`, where secrets are always present — so the condition can only mean "the secret was renamed, deleted or rotated to empty", and it responds by silently skipping login. The multi-arch build then pulls anonymously across two platforms and three stages and can hit the rate limit, failing the release with an error pointing nowhere near the missing secret. Make it unconditional here.

**Suggestions**
- Rollout narration in a permanent file (`# The rollout for this rehearses with a v2.1.0-rc.1 prerelease…`) — keep the invariant, drop the event. Two accuracy issues in the same comment: `latest=auto` is `metadata-action`'s **default** (so "auto, not true" reads as a deviation that isn't one), and withholding `{{major}}`/`{{major}}.{{minor}}` from prereleases belongs to `type=semver`, not to `flavor:`.
- No `concurrency:` group — two Releases in quick succession race on `latest`, `2`, `2.1`.
- The `--out` outside-the-mount claim ("exits 0, file never appears") is narrower than stated: under the mandated `--user`, most out-of-mount paths fail loudly with EACCES. The reliably silent case is a *writable* out-of-mount path, i.e. `HOME=/tmp`, which is the next bullet.
- The new "what the image carries" table lists `ts-lint`, but the oclif-generated `<!-- commands -->` block still doesn't — two tables in one README disagree about which commands exist.
- Third-party actions are tagged, not SHA-pinned, in the one job holding `packages: write`. Consistent with repo convention, so not a regression — but a publish job is where pinning earns its cost.

**Positive:** the publish gating is genuinely least-privilege and unreachable from a fork or PR (`if: github.event_name == 'release'`, `contents: read` + `packages: write`, no `id-token`), and `needs: docker-parity` means the compiled-CLI comparison blocks the push. `latest=auto` over the copy-pasted `latest=true` is the right call. The README's verifiable claims check out against source: `-p 5600` is passed explicitly because the CLI default is `5907`, `GET /health` exists, the readiness line is printed after `listen` resolves, and `generate` does log an absolute path.

---

## Suggested order

1. **Decide C1** (`engines` vs eslint 10 as a runtime dep) — it's the only finding that breaks installs for existing external users, and it may force a major.
2. **#2711 Criticals** — empty-dir `ok: false`, and `process.exitCode = 1`. Both are one-liners.
3. **#2709 Critical** — split the `.js` block; re-baseline the warning count and decide `no-non-null-assertion` explicitly.
4. **#2710** — point the smoke test at a `--prod` install, since that's the property being fixed.
5. **C3** — promote `docker-parity`, or state it's advisory with an owner and ticket.
6. **Doc corrections** — #2708's "no behaviour change" paragraph, #2713's `AGENTS.md:38` and the `prepack` comment, #2714's `2.1.0` tag.
7. Tests that turn the good comments into mechanisms: the TTY-present case, the readiness string, the `overrideConfigFile` discriminator, the five dead rule disablements.

Nothing here argues against the design. The stack's architecture — fix resolution, migrate the lint stack alone, declare runtime deps, add the command, harden non-interactive use, containerise, publish — is well-sequenced and each PR is genuinely atomic.

timdawborn:
# Review — #2708 `fix(COMPASS-17): resolve @airtasker/spot from spot's own tree`

Automated multi-agent review (`/pr-review-toolkit:review-pr`): general code review, test-coverage analysis, and error-handling analysis, scoped to this PR's diff against `master`. Findings marked **[verified]** were reproduced by building the branch and running the code, not inferred from reading.

Reviewed as PR 1 of the seven-PR COMPASS-17 stack (#2708 → #2714); each PR reviewed separately against its own base.

---

## Important

### 1. The mapping now *overrides* a consumer's installed copy, and the PR body claims the opposite
**[verified — two agents independently, with sentinel packages]** · `lib/src/ts-project.ts:31`

The body states:

> **No behaviour change for npm consumers.** A consumer that has `@airtasker/spot` installed resolves the mapping to the same declarations it was already getting from `node_modules`.

Both agents built a consumer directory with its own divergent `node_modules/@airtasker/spot` (exporting a sentinel symbol spot's own `lib` lacks) and parsed it under each mapping:

| mapping | result |
|---|---|
| old `../lib` (never existed) | parses — **consumer's `node_modules` copy won** |
| new `lib` (exists) | `has no exported member 'SENTINEL_FROM_CONSUMER_COPY'` — **spot's own copy won** |

TypeScript consults `paths` *before* `node_modules` whenever the mapped target exists. On `master` the target never existed, so the entry silently fell through and `node_modules` always won. This PR makes the target exist, so `paths` now wins outright and the installed copy is never consulted.

This is very likely the **correct** semantics — the parser and the declarations it type-checks against should be one version, and the container work requires it. The problem is that it is undocumented and actively denied.

The claim holds only when the installed version equals the running version. It breaks for the realistic cases: `npx @airtasker/spot@latest` against a repo pinned to an older devDependency, a monorepo with a hoisted older copy, or a contract using a type removed in the newer spot. Because `parse()` throws on *any* diagnostic, one removed export hard-fails `generate`, `lint`, `validate` and `mock`.

**Ask:** rewrite that paragraph to state the intended rule (the running spot's declarations always win), and add a sentence to the `ts-project.ts` docstring naming the precedence — it currently describes only the fill-in half, not the override half.

### 2. The compiled layout is asserted in a docstring and tested nowhere
`lib/src/ts-project.ts:6-11`

The docstring asserts a two-layout invariant: `lib/src/lib.ts` from source, `build/lib/src/lib.d.ts` once compiled. ts-jest only ever exercises the first; the body confirms the compiled check was done by hand.

`tsconfig.json:20` has `rootDir` commented out, so the emitted layout is *inferred* from the common root of the input set. Any change to `include` that adds a file outside `cli/src`/`lib/src`, or moving `ts-project.ts` into a subdirectory, shifts the output path — `path.join(__dirname, "lib")` then names a directory that does not exist, `pnpm test` stays fully green, and every published CLI and container invocation fails with `Cannot find module '@airtasker/spot'`, i.e. exactly the bug this PR fixes. #2712 adds a second consumer of `contractCompilerOptions`, so the blast radius grows.

One line in CI after `pnpm build` closes it:

```
node -e 'const{contractCompilerOptions:o}=require("./build/lib/src/ts-project");const p=o.paths["@airtasker/spot"][0];if(!require("fs").existsSync(p+".d.ts"))throw new Error("mapping broken in build layout: "+p)'
```

(#2713's parity script may cover this end-to-end — if so, say so here, and note it should be a required check. See the review on #2713.)

---

## Suggestions

- **The type-error assertion passes under both mappings.** **[verified]** `parser.spec.ts:29` asserts `.toThrow(/not assignable to type 'number'/)`. `validateProject` joins *all* diagnostics with `\n` and Jest's `toThrow(regex)` is a substring match, so the broken mapping simply adds a `Cannot find module` line alongside the expected one and the test still passes. It cannot distinguish "type-checking is on and the import resolved" from "type-checking is on and the import is broken". Add `expect(() => parse(TYPE_ERROR_FIXTURE)).not.toThrow(/Cannot find module/)`, or assert the message exactly.

- **Case (b) is structurally untestable today.** `test-fixtures/contract/` sits in a tree with no `@airtasker/spot` anywhere, so the precedence question above cannot be reached by the existing test. A fixture with its own deliberately-divergent `node_modules/@airtasker/spot` would pin whichever decision you land on in finding 1 — including the realistic skew where the consumer copy is an older *subset*.

- **`contractCompilerOptions` is a shared mutable object** handed by reference to every `Project`. With #2711 adding `ts-lint` as a second consumer, a single `contractCompilerOptions.skipLibCheck = true` anywhere silently changes every subsequent `parse()` in the process. `Object.freeze` costs nothing now and is awkward to retrofit.

- **Most compiler flags are unpinned.** **[verified]** Setting each to `false` in turn: `strictNullChecks` and `noImplicitAny` break a test (incidentally, via the fixture's optional fields); `strict`, `noImplicitReturns`, `noFallthroughCasesInSwitch`, `resolveJsonModule`, `alwaysStrict`, `strictFunctionTypes`, `strictPropertyInitialization`, `noImplicitThis` all leave the suite green. Since the stated reason for extracting this object is that `ts-lint` shares it so contracts are checked "under exactly the options `parse()` uses", a snapshot assertion on `contractCompilerOptions` makes any weakening visible in review.

- **Two fixture conventions, no recorded rule.** `CLAUDE.md`'s Testing section says fixtures live in `__spec-examples__/`; this PR adds fixtures in both there and `test-fixtures/`. The distinction is real (the latter is a whole-directory contract, also consumed by #2713's parity script, and deliberately outside tsconfig `include`) but nothing records it. Worth a line in `CLAUDE.md`.

- **`validateProject` discards file and line.** It joins only `getMessageText()`, so every contract error reaches users without a location. #2711's `runTypeCheck` already formats these correctly — worth backporting. Not for this PR.

---

## Verified clean

- **No silent-failure shapes in the new code.** `ts-project.ts` has no `try`/`catch`, no `?? []`, no `|| default`, no optional-chaining escape. `validateProject` throws on any pre-emit diagnostic with no filtering, so an unresolved module (TS2307) aborts the parse rather than degrading to `any`. This PR *removes* a silent cross-version fallback rather than adding one.
- No ticket key, `TODO`, `FIXME`, provenance or change-commentary text in any new file.
- `pnpm build` stays green (the deliberate type error is excluded via `tsconfig.json` `exclude`), `test-fixtures/` is not emitted into `build/` and not in `package.json` `files`, so it is not published.
- Pulling spot's own 54 source files into the contract project is **pre-existing**, not introduced here — `lib/src/spec-helpers/helper.ts:47` already maps to the identical target on `master`.

## Strengths

- The fail-before/pass-after test is real and mutation-verified: reverting only the path gives `Cannot find module '@airtasker/spot'`. On `master` *zero* tests exercised contract-import resolution — all 527 were green against a mapping that was wrong in both layouts — so this is net-new coverage of a genuinely silent bug.
- The fail-closed second case is the right instinct. A mapping that silently stopped type-checking would pass the positive case alone; the type-error fixture closes that hole (modulo the substring issue above).
- The fixture's breadth is load-bearing rather than decorative: all 12 symbols imported in `api.ts` must exist in the mapped target, so a *partial* declaration target is caught, not just a missing one.
- Extracting `contractCompilerOptions` while explicitly leaving `spec-helpers/helper.ts` alone keeps the PR atomic instead of bundling a speculative de-duplication.

timdawborn:
# Second-pass review — #2708

Multi-agent re-review (`/pr-review-toolkit:review-pr`) of both commits, with the follow-up `717d717` reviewed adversarially. Three agents worked this PR; every finding below was reproduced by building the branch and running code, not inferred from the diff.

**Verdict: the original fix is correct and verified. The follow-up is directionally right and unsound as written** — the guard tests a proxy rather than the value the compiler consumes, the freeze does not prevent the mutation its own comment names, and the added assertion is redundant while costing ~8s per CI run.

---

## Important

### 1. `Object.freeze` is shallow, and the guard reads a different value from the one the compiler uses — so neither closes the failure the comment claims
`lib/src/ts-project.ts:28-29`, `:31`, `:62-73` — *two agents independently*

`paths` is not frozen. Reproduced in-process:

```
frozen: true
paths frozen: false
mutated paths value: {"@airtasker/spot":["/nonexistent/broken"]}
parse after mutating paths: Error: Cannot find module '@airtasker/spot' or its corresponding type declarations.
```

`contractCompilerOptions.paths!["@airtasker/spot"][0] = …` succeeds silently. `assertSpotDeclarationsResolve()` then still passes, because it reads the `SPOT_DECLARATIONS` constant and never the live `paths` entry the compiler is handed. So there are two sources of truth, and after that mutation the guard reports everything fine while every `parse()` in the process fails with exactly the bare `Cannot find module` the comment says the guard converts into "an error naming the real cause".

Separately, `export const contractCompilerOptions: CompilerOptions = Object.freeze({…})` discards the `Readonly<T>` that `Object.freeze` returns, so `contractCompilerOptions.strict = false` type-checks and throws only at runtime.

Both of these tests fail against the current code:

```ts
test("the shared options object cannot be mutated", () => {
  const paths = contractCompilerOptions.paths as Record<string, string[]>;
  expect(() => { paths["@airtasker/spot"][0] = "/broken"; }).toThrow();
});

test("the mapping the parser actually uses points at real declarations", () => {
  (contractCompilerOptions.paths as Record<string, string[]>)["@airtasker/spot"][0] = "/broken";
  expect(() => createContractProject()).toThrow(/declarations are missing/);
});
```

**Fix:** derive the guard from `contractCompilerOptions.paths["@airtasker/spot"][0]` so there is one source of truth, and freeze `paths` and the array too — or build the options in a factory, which sidesteps both. One change makes both tests above pass and gives the guard body its first real coverage.

### 2. The guard's trigger is a two-extension proxy, so it hard-fails a layout that resolves correctly
`lib/src/ts-project.ts:63-64`

Node-style resolution of a `paths` target also accepts the directory form. Verified — a target directory containing `lib/index.d.ts` resolves the contract's import with **0 diagnostics**, while both `existsSync` checks are false:

```
assert-style existsSync(.ts): false  (.d.ts): false
diagnostics: 0 []
```

**Failure scenario:** `lib/src/lib.ts` grows and someone splits it into `lib/src/lib/index.ts` — a routine refactor that keeps the mapping working — and every Spot command dies with *"Spot's own @airtasker/spot declarations are missing … the emitted layout moved — check outDir and rootDir in tsconfig.json"*. Nothing about the emitted layout is wrong; the message asserts a cause that is not the cause. `.tsx` is likewise accepted by resolution and rejected by the guard, and the message misdiagnoses a bundled consumer (`ncc`/webpack) too.

**Fix, which also dissolves finding 1:** ask the compiler instead of the filesystem.

```ts
ts.resolveModuleName("@airtasker/spot", <a file in this dir>, contractCompilerOptions, ts.sys)
```

Fail when `resolvedModule` is undefined. That tests the failure's actual precondition and makes the message honest.

### 3. The new error path has zero coverage
`lib/src/ts-project.ts:62-73`

`assertSpotDeclarationsResolve` is module-private and reads a module-private constant, so nothing can exercise either arm. **Confirmed: replacing the body with `return;` leaves 529 tests / 48 suites passing.** A typo in the candidate list — `".dts"`, or dropping `".d.ts"` and breaking every compiled install — ships green. The same commit that removes a real signal from the suite (finding 4) adds an unverified one.

Export it with an injectable target and cover both arms, asserting the thrown message.

---

## Suggestions

- **The added assertion is redundant, and its comment describes a mutation that can no longer reach it.** `lib/src/parser.spec.ts:36`. It *does* discriminate — breaking the mapping **key** (`"@airtasker/spot"` → `"@airtasker/spot-typo"`) fails it while line 29 still passes. But the mutation it was written for, a broken **target**, now trips `assertSpotDeclarationsResolve` first, so line 29 fails on the assert's message and line 36 never evaluates. Every mutation that reaches line 36 also fails the test above it, so it is caught-but-never-uniquely. It is not free: it re-runs `parse()`, rebuilding the whole program — measured 10.4s and 18.4s for the two tests, so roughly **8s of CI per run**. Either delete it, or narrow the comment to "a mapping broken by its key or by `moduleResolution`/`baseUrl`".

- **The freeze rationale names a consumer that does not exist.** `ts-project.ts:28` — "Frozen because `ts-lint` shares this object". `git grep contractCompilerOptions` across the stack returns two hits, both inside `ts-project.ts`: `ts-lint` imports `createContractProject`, and neither `core.ts` nor `lib.ts` re-exports the constant. Delete the sentence, or state the invariant without naming a consumer.

- **The precedence docstring is right but absolute.** `ts-project.ts:20-26` — "**always** … **never** against a different version installed next to it". Verified correct for the bare specifier (with a divergent `node_modules/@airtasker/spot` present, the mapping won with 0 diagnostics; without the mapping, the installed copy produced 11 `has no exported member` errors). But there is no `"@airtasker/spot/*"` wildcard, so a **subpath** import still resolves through `node_modules`. Scope the sentence to the bare specifier, or add the wildcard.

- **Comment density on this file went 32% → 45%.** The 18-line docstring on a config object is the largest block added anywhere in the stack. Three lines carry it once findings 1 and 2 are addressed. Related: `parser.spec.ts:32-35` is four lines of review reply above a one-line assertion, reproduced word-for-word in the commit body — the commit message is the right home.

- **`fs.existsSync` cost is a non-issue** — two calls are microseconds against a measured 499ms compiled `parse()`. Do not add memoisation.

---

## Verified clean

- No silent-failure shapes in the new code: no `try`/`catch`, no `?? []`, no `|| default`, no optional-chaining escape. `validateProject` throws on any pre-emit diagnostic with no filtering.
- No bare ticket keys, no `TODO`/`FIXME`, no provenance or change-commentary text.
- `eslint` 0 errors / 25 pre-existing warnings, `prettier --list-different` clean, `pnpm build` clean.
- No caller left behind: only `ts-project.ts` and `spec-helpers/helper.ts` construct a `Project`, and the latter's `../lib` really is correct from `spec-helpers/`.

## Pre-existing, worth knowing before the container work lands

The from-source arm drags Spot's implementation into the contract program: **57 files (54 of them Spot's own source)** versus 31 from the compiled `.d.ts` arm. Because `baseUrl: "./"` makes the `@types` lookup process-cwd-relative, this is cwd-sensitive:

```
cwd: /private/tmp   → ["src/ts-project.ts: Cannot find module 'fs' …", "… 'path' …", "… Cannot find name '__dirname'."]
cwd: <repo root>    → []
```

Any from-source invocation whose cwd has no `@types/node` above it reports Spot's own `ts-project.ts` as the reason the user's contract is invalid. The new spec is green only because jest's cwd is the repo root. Filtering `validateProject`'s diagnostics to files outside Spot's own tree would make the fail-closed gate cover the contract and only the contract.

## Strengths

- The core fix is real and independently reproducible: the built output parses `test-fixtures/contract/` from a directory outside the repo in 499ms, and the precedence the rewritten body now claims is exactly what a divergent installed copy demonstrates. The description matches the code.
- The type-error fixture is a genuine fail-closed guard on the right axis — a mapping that silently stopped type-checking would pass the resolution test alone.
- Correcting the body's "no behaviour change for npm consumers" rather than defending it was the right call; the precedence change is real and intended.

vlj91 (approved):
LGTM.

PR #2709, opened 2026-08-18
build(COMPASS-17): migrate spot's lint stack to eslint flat config

timdawborn:
## Ticket

> **Stacked on #2708** — review that one first; this branch is based on it. Base will retarget to `master` once #2708 merges.

## What

Move spot's own lint stack from `.eslintrc.js` on eslint 7 + `@typescript-eslint` 4 to `eslint.config.js` on eslint 10 + `typescript-eslint` 8.

## Why

The `ts-lint` command later in this ticket bundles eslint as a **runtime dependency** of the published package, so the docker image can lint a consumer's contract tree with no eslint installed beside it. A package cannot carry eslint 7 in `devDependencies` and eslint 10 in `dependencies` at the same time — one version wins the `node_modules/.bin` entry. So spot's own stack goes onto the target major first, on its own, where a regression is attributable.

Nothing outside the lint setup changes shape here. The repo's own config stays deliberately **separate** from the canonical contract config `ts-lint` will bundle: that one has to be an in-code object so its plugins resolve from the image's own `node_modules` rather than from whatever sits next to the contract being linted. Merging the two would defeat that.

## Changes

- `.eslintrc.js` → `eslint.config.js`. Eslint 9 and later do not read eslintrc.
- `.eslintignore` deleted; its entries move into the config's `ignores`. Eslint 9 and later do not read `.eslintignore` either — leaving it in place would have silently started linting `build/`.
- `eslint`: `^7.32.0` → `^10.8.1`; `@typescript-eslint/{parser,eslint-plugin}`: `^4.31.2` → the unified `typescript-eslint@^8.67.0`; `eslint-plugin-jest`: `^27.6.3` → `^29.16.1`; `@eslint/js` and `globals` added (flat config imports `eslint:recommended` and the node globals as modules).
- `eslint:check` / `eslint:fix` drop `--ext`, removed from the CLI in eslint 9. Flat config names the extensions in `files` instead.
- Five source files, see below.

**No TypeScript bump is needed.** `typescript-eslint@8`'s peer range is `>=4.8.4 <6.1.0`, and spot is on `^4.9.5`. The plan for this ticket assumed a TS 5 bump would be forced here; it is not, so it stays out of this PR and out of the next one unless something else demands it.

### Rule translation

One-for-one, apart from three changes the upgrade forces:

| Old | New | Why |
|---|---|---|
| `@typescript-eslint/ban-types: off` | `no-unsafe-function-type` and `no-wrapper-object-types` `off`; `no-empty-object-type` as `["error", { allowObjectTypes: "always" }]` | typescript-eslint 8 split the rule. `String`, `Date` etc. are Spot's own contract vocabulary, so the wrapper-types rule in particular has to stay off. `no-empty-object-type` also absorbed `no-empty-interface`, which the old `ban-types: off` did **not** cover — so it is enabled with `{}` allowed, keeping empty interfaces reported. |
| `no-explicit-any` at recommended severity (warn under v4) | `warn`, explicitly | v8 recommended makes it an error. The two sites are deliberate (`Example.value` carries an untyped JSON literal; `typeOf()` takes anything). Tightening the rule and retyping them is separate work from this migration. |
| `no-non-null-assertion` at recommended severity (warn under v4) | `warn`, explicitly | v8 moved it from `recommended` to `strict`. Named explicitly so the upgrade does not silently stop reporting its 23 sites. |
| — | `js.configs.recommended` only on `**/*.js` | `tseslint.config()` rewrites an extended config's `files` to the parent's, so applying the TypeScript preset to `.js` alongside `.ts` would switch off the core checks the compiler performs — on files that are never compiled. |
| — | `no-require-imports: off` for `**/*.js` | The only linted `.js` file is `eslint.config.js` itself, and the package is CommonJS. |

The three `overrides` blocks (contract-syntax files, `lib/src/syntax/**`, `*.tsx`, `*.spec.ts`) carry across unchanged as flat-config entries.

### Code changes falling out of stricter defaults

None behavioural:

- `no-unused-vars` now covers caught errors (`caughtErrors` defaults to `all` in v8). Two `catch (e)` blocks that never read the binding become `catch` — optional catch binding, ES2019, and the target is `es2019`.
- Unused `eslint-disable` directives are now reported. Two were dead, both for `explicit-function-return-type`, which is in neither the old nor the new recommended set. (A third, for `no-non-null-assertion`, is *not* dead — see the rule-translation note below — and is retained.)

## How this was verified

- `pnpm lint:check` clean — 0 errors, and exactly the same **25** warnings the old stack reported (2 `no-explicit-any`, 23 `no-non-null-assertion`).
- **File coverage is unchanged.** The risk in an eslintrc → flat-config migration is quietly linting fewer files: `--ext` is gone, `.eslintignore` is no longer read, and flat config defaults to `.js` only. Ran `eslint -f json` on both this branch and its parent and diffed the resulting file lists — **282 files each**, differing only in the config file's own name (`.eslintrc.js` → `eslint.config.js`).
- `pnpm test` — 529 tests across 48 suites pass.
- **`build/` output diff** against the parent branch is confined to the five touched files and contains only the dropped `catch` bindings and the removed directive comments. No other emitted file changes.
- Swept for stale references to the deleted files with `rg --hidden` and `git grep` (which covers dotfiles): none.

## Note on CI

`build-and-test` is `on: pull_request: branches: [master]`, so it was expected not to run while this PR targets #2708's branch. It **does** run and report here — `test-node-{18,20,22,24}` and `lint-check` are green on this PR. Everything under *How this was verified* was also run locally.

## Note on `dependency-review`

That check fails on this PR, as it does on every PR and every Dependabot branch in this repo, with `Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled`. It is a repo settings issue, unrelated to this change, and predates it — but a permanently-red check is worth someone turning the setting on or removing the workflow.


timdawborn:
# Review — #2709 `build(COMPASS-17): migrate spot's lint stack to eslint flat config`

Automated multi-agent review (`/pr-review-toolkit:review-pr`), scoped to this PR's diff against its base (#2708). Findings marked **[verified]** were reproduced by installing both branches and diffing `eslint --print-config` and `eslint -f json` output, not inferred from reading.

---

## Critical

### 1. `.js` files silently lose `no-undef` and 15 other core correctness rules
**[verified]** · `eslint.config.js:25-26`

```js
files: ["**/*.js", "**/*.ts", "**/*.tsx"],
extends: [js.configs.recommended, tseslint.configs.recommended],
```

`tseslint.configs.recommended` bundles typescript-eslint's `eslint-recommended` block, which turns off the core rules TypeScript already enforces. Upstream scopes that block to `**/*.{ts,tsx,mts,cts}`. But `tseslint.config()`'s `extends` **overwrites the extended config's own `files` with the parent's**, so those "TypeScript handles this" disables now apply to `**/*.js` as well.

Proven with a root `probe.js` containing `module.exports = someUndefinedGlobal;`:

- parent branch: `error 'someUndefinedGlobal' is not defined  no-undef`
- this branch: **no output**

Full set lost on `.js` vs the parent branch: `no-undef`, `no-const-assign`, `no-dupe-args`, `no-dupe-keys`, `no-dupe-class-members`, `no-func-assign`, `no-import-assign`, `no-obj-calls`, `no-redeclare`, `no-setter-return`, `no-this-before-super`, `no-unreachable`, `no-unsafe-negation`, `getter-return`, `constructor-super`, `no-new-symbol`. All were active on `.js` under `.eslintrc`.

Exposure today is one file (`eslint.config.js`), which is exactly why it will not be caught — but this stack adds JS surface downstream. A typo'd identifier in a future `.js` script passes `pnpm lint:check` clean and fails at require-time in the image.

**Fix:** split the block so `.js` gets `js.configs.recommended` only, and `.ts`/`.tsx` get both.

---

## Important

### 2. `@typescript-eslint/no-non-null-assertion` is silently dropped — and the PR body's baseline is wrong
`eslint.config.js:26`, `lib/src/types.ts:588`

The rule was in typescript-eslint v4's `recommended` (warn); v8 moved it to `strict`, so nothing enables it now. That is 23 live warnings disappearing:

| file | count |
|---|---|
| `lib/src/parsers/example-parser.spec.ts` | 8 |
| `lib/src/parsers/schemaprop-parser.spec.ts` | 14 |
| `lib/src/types.ts` (line 622) | 1 |

The body claims *"exactly the same 2 `no-explicit-any` warnings the old stack reported"*. The old stack reported **25** — and **#2708's own body says so** ("25 pre-existing warnings, 0 errors"). The two PR bodies in this stack contradict each other, and that inaccurate baseline is what let this drop through unnoticed.

Relatedly, the body says the removed directive at `types.ts:588` "was not even reachable under the old one". It was reachable — the rule was live (see `types.ts:622`) and the directive was actively suppressing the `!` on the following line.

**Ask:** either re-enable explicitly (`"@typescript-eslint/no-non-null-assertion": "warn"`) and keep the directive, or state in the PR that the rule is being dropped and why. Either way the "25 warnings" line needs correcting.

### 3. `no-empty-object-type: "off"` drops more than `ban-types` did — `no-empty-interface` goes with it
`eslint.config.js:34`

The rule-translation table treats `ban-types → {no-empty-object-type, no-unsafe-function-type, no-wrapper-object-types}` as one-for-one. It is not: v8's `no-empty-object-type` **also replaces `@typescript-eslint/no-empty-interface`**, which was on in v4 `recommended` and was *not* covered by the old `ban-types: "off"`. Confirmed — `no-empty-interface` appears in the lost-rule diff for every file scope.

`interface Foo extends Bar {}` now lints clean anywhere in the repo — a real footgun in a contract-definition codebase, since it is not a nominal type. Faithful translation is `["error", { allowObjectTypes: "always" }]` rather than `"off"`.

### 4. `engines.node: ">=18.12.0"` is now false, and the CI matrix still builds on Node 18
`package.json:53`, `.github/workflows/build-and-test.yml:22`

From the lockfile this branch ships:

- `eslint-plugin-jest@29.16.1` → `engines: {node: ^20.12.0 || ^22.0.0 || >=24.0.0}`

It survives today only because there is no `engine-strict=true` (pnpm warns rather than fails) and because the `lint-check` job's `setup-node` has **no `node-version:` input**, so it inherits the runner image default. Two ways that breaks: an engineer on Node 18 — a version this repo claims to support — runs `pnpm lint:check` and gets a hard failure; or the `ubuntu-latest` default Node moves and `lint-check` starts running on an unintended version.

**This gets materially worse in #2711**, where the same packages move into `dependencies` and reach consumers. Worth deciding here: bump `engines` to `>=20.19.0` and drop 18 from the matrix, or pin `node-version` on `lint-check`. See the review on #2711.

### 5. Deferred-decision comment with no follow-up ticket
`eslint.config.js:40-41`

```js
// Warn, the severity it carried before the upgrade. Tightening it to an
// error is a separate piece of work from this migration.
```

Per `pull-requests.md`, a deferred-decision label ("separate piece of work") must name its own follow-up ticket inline, and it must not be COMPASS-17 — that closes when this stack merges, making the pointer dead exactly when discoverability matters. Rewrite as `// TODO(<follow-up-ticket>): raise to "error" once Example.value and typeOf() are retyped.` Same applies if `no-non-null-assertion` is deliberately left off per finding 2.

---

## Suggestions

- **`"build/"` no longer ignores nested build directories.** **[verified]** Flat-config global-ignore patterns anchor to the config base path, so `"build/"` matches only the root `build/`; the deleted `.eslintignore`'s bare `build` matched at any depth. With `build/emitted.js`, `build/nested/deep.js` and `lib/build/inner.js` present, the parent branch lints 282 files and this branch lints **283** — `lib/build/inner.js` is now linted. Latent today (nothing emits a nested `build/`). Use `"**/build/"` for parity. Note the body's "282 files each" check could not have exercised this — a fresh checkout has no `build/` at all.

- **Jest rules narrowed from repo-wide to `**/*.spec.ts`.** `eslint.config.js:55-56`. The old `extends: ["plugin:jest/recommended"]` was top-level, so all 19 jest rules ran on every file. Defensible (it matches `jest.config.js` `testMatch`), but it is a fourth deviation the "one-for-one, apart from three changes" table does not list. `jest/no-commented-out-tests` and `jest/no-focused-tests` no longer cover non-spec files.

- **Dead config lines, and a comment that now overclaims.** `@typescript-eslint/no-empty-function` moved to v8's `stylistic` preset and `explicit-function-return-type` was never in v4 or v8 `recommended`, so those `"off"` entries are now no-ops. The PR removed the matching `eslint-disable` comments as dead but kept these — inconsistent. Relatedly, the `CONTRACT_SYNTAX_FILES` comment says those rules are off "wherever contract syntax lives **rather than repo-wide**"; for `no-empty-function` that distinction no longer exists.

- **Provenance narration in a new file.** `eslint.config.js:31-33` — "…the old config turned that rule off…" narrates the migration rather than the invariant. State the decision: `// String, Date etc. are Spot's own contract vocabulary, so the wrapper/empty-object/unsafe-function checks stay off.` The lineage belongs in the commit message.

- **Undocumented severity change.** `no-unused-vars` goes warn → error under v8 recommended. A tightening, and it passes clean — but the PR deliberately pinned `no-explicit-any` back to `warn` for exactly this reason and did not mention this one. Worth a table row so the asymmetry reads as a choice.

---

## Verified clean

- **All five source edits are behaviour-preserving**, confirmed line by line: `example-parser.ts:73` / `schemaprop-parser.ts:113` are `catch (e)` → `catch` with the binding unread in both blocks (optional catch binding is ES2019, matching `"target": "es2019"`); `mock-server/server.ts:107`, `validation-server/server.ts:65`, `types.ts:588` are comment-only deletions.
- `pnpm` used exclusively; no `package-lock.json`/`yarn.lock` introduced.
- No stale references to the deleted `.eslintrc.js` / `.eslintignore` anywhere, checked with `git grep` (which covers dotfiles).
- `pnpm prettier:check` and `pnpm eslint:check` both exit 0 on the branch; the scripts work without `--ext`.
- No `COMPASS-17` key anywhere in committed code.

## Note on CI

The body says `build-and-test` will not report here. **It has** — `test-node-{18,20,22,24}` and `lint-check` are all green on this PR. Only `dependency-review` fails, which is the pre-existing repo-settings issue the body correctly diagnoses. Worth correcting, since reviewers were told to discount signal that actually ran.

## Strengths

- Genuinely verified the risk that matters most in an eslintrc → flat migration (silently linting fewer files) with a real `eslint -f json` diff rather than an assertion — 282 files on both sides, independently reproduced. Finding 1 is a refinement of that check, not a refutation of it.
- Keeping the repo's own config deliberately separate from the canonical contract config `ts-lint` will bundle is the right call, with the reason stated, and avoids a much worse coupling downstream.
- Correctly established that no TypeScript bump is forced (`typescript-eslint@8.67.0` peers `>=4.8.4 <6.1.0`, confirmed in the lockfile) and kept it out of scope rather than bundling it.

timdawborn:
# Second-pass review — #2709

Multi-agent re-review of both commits, with the follow-up `a52e12e` reviewed adversarially. Every claim in the follow-up was measured in isolated worktrees of `a52e12e` and its parent, by diffing resolved configs and rule-level output rather than reading the diff.

**Verdict: no functional defects. All five claims hold, and two of them are load-bearing rather than cosmetic** — unusual for a config migration of this size. Everything below is accuracy and hygiene.

---

## Verification results

| Claim | Verdict | Evidence |
|---|---|---|
| `.js` split; core preset only; `no-require-imports` deletion safe | **Holds** | `eslint .` 0 errors; `eslint.config.js` itself lints clean. `--print-config`: `sourceType: commonjs`, `@typescript-eslint/no-require-imports = undefined` (plugin absent, so the exemption is genuinely unreachable), `no-undef = [2]` live. Probe `.js` with an undefined global, a redeclare and unreachable code → 6 errors. Rule-diff for a `.js` file against the pre-fix commit: **19 core rules restored** — the regression was real, and larger than the commit message's "a dozen more" implies. |
| 25 warnings, matches parent, directive not unused | **Holds exactly** | Both branches emit 25; `diff` of sorted `file:line severity rule` is **byte-identical** (2 `no-explicit-any` + 23 `no-non-null-assertion`). `reportUnusedDisableDirectives` confirmed active via a bogus probe directive; no 26th warning appears. Deleting `types.ts:591` yields 26 → the restored directive is load-bearing. |
| `no-empty-object-type` with `allowObjectTypes: "always"` | **Holds** | `interface ExtendsOne extends Base {}` reports; bare `interface Bare {}` reports; `type T = {}` and `const v: {}` do not. Option name is schema-validated. **Load-bearing**: removing it surfaces 3 real errors, including `AnyValue: {}` at `openapi2-specification.ts:473` and `openapi3-specification.ts:236` — so the comment's "`{}` appears as a contract type" is true. |
| `["**/build/"]` | **Holds** | With `lib/src/nested/build/*.ts` planted: bare `build/` lints it, `**/build/` ignores it. Clean count **282**, identical to parent; file-list diff differs only by the config filename. |
| Three `off` entries were dead | **Holds** | `--print-config` on a contract-syntax file, a `lib/src/syntax` file and a `.spec.ts`: `@typescript-eslint/no-empty-function`, `no-empty-function` and `explicit-function-return-type` all **`undefined`**. `pnpm lint:check` and `pnpm build` exit 0. |

Also verified clean: peer ranges all satisfied with no install warnings; `**/*.spec.ts` gets 19 jest rules; `docs/*.tsx` linted on both branches. The `.js` coverage delta versus `master` (`no-extra-semi`, `no-mixed-spaces-and-tabs`) is an upstream consequence of the eslint 7→10 major, not this split — `no-var`/`prefer-const` were never on `.js` under the old stack either.

---

## Important

### 1. The PR body is stale in three places after the follow-up
Each is independently checkable, and the first has a real consequence.

- **Rule-translation table, last row** still lists `no-require-imports: off` for `**/*.js` as a change this migration makes. The follow-up **deleted** that entry — `git show a52e12e:eslint.config.js | grep no-require-imports` returns nothing. A reviewer working the table cannot find it and may conclude `.js` still needs a plugin-rule exemption, then later re-add a `.js` block that extends `tseslint.configs.recommended` — silently re-dropping the 19 core rules this commit just restored. Replace the row with one stating `.js` receives no plugin rules at all, so no exemption is needed.

- **"The three `overrides` blocks (…) carry across unchanged as flat-config entries"** is wrong three ways: four blocks are listed, not three; they did not carry across unchanged (five `off` entries were deleted across them); and `*.spec.ts` least of all — its only rule was deleted and it now carries `extends: [jest.configs["flat/recommended"]]`, which under `.eslintrc.js` was a repo-wide `extends`. That last point is a genuine scope narrowing (jest rules went from repo-wide to `**/*.spec.ts`) presented as "unchanged". Harmless today — `git grep -lE '^\s*(describe|it|test)\(' -- '*.ts'` returns nothing outside `*.spec.ts` — but it should be disclosed.

- **"Five source files"** and **"the `build/` output diff … is confined to the five touched files"** — it is four. `git diff 717d717..a52e12e -- lib/src/types.ts` is **empty**: restoring the directive made `types.ts` byte-identical to the parent. A reviewer reproducing the build-diff claim counts four and cannot reconcile it.

### 2. Three new comments narrate the migration rather than the invariant, and one names the wrong rule set
`eslint.config.js:18-19`, `:47-49`, `:49-51`, `:65-66`

The repo's own rule ("Describe the abstraction, not its provenance, and not the current change") lists `port`/`this PR`/`not yet` in its pre-push grep. Apply its test — *would a reader who never saw this PR need this line before editing the surrounding code?*

- `:18-19` — "`.eslintignore` matched at any depth" is provenance pointing at a file **this same PR deletes**. Dead on merge. The first sentence is the load-bearing half and stands alone.
- `:49-51` — "`no-empty-interface`, which v8 folded into `no-empty-object-type`, and it was never part of the allowance" is change commentary referencing a `ban-types` allowance no longer in the file. State what the config enforces.
- `:65-66` — "v8 moved this out of `recommended` into `strict`. Named explicitly so **the upgrade** does not silently stop reporting…". There is no "upgrade" once this merges, and "v8" becomes wrong at the next major.

**And a factual error at `:47-49`:** "`String`, `Boolean` and `Date` are Spot's own contract vocabulary, not the JavaScript wrapper objects these rules exist to catch". Read from the installed rule source (`@typescript-eslint/eslint-plugin@8.67.0/dist/rules/no-wrapper-object-types.js`), the caught set is `BigInt, Boolean, Number, Object, String, Symbol`. **`Date` is not caught by any of the three rules named**, and `Number`, which is, was dropped. It also disagrees with the sibling list in #2711 (`String`, `Number`, `Boolean`). Name no list: "Spot's contract vocabulary reuses these identifiers as types; `{}` also appears as a contract type."

The mechanism comment at `:27-32` **passes** the test and should stay — a future author merging the two blocks back together would reintroduce exactly the bug this commit fixes.

### 3. The `**/build/` parity fix was applied to one of four entries, while the commit message claims parity
`eslint.config.js:20-23`

The commit message states the goal as "ignore `build/` at any depth, **as `.eslintignore` did**". Under `.eslintignore`, `jest.config.js`, `jest.ci.config.js` and `webpack.config.js` were bare patterns that also matched at any depth; in flat config all three are now path-anchored. So three of the four entries still carry precisely the footgun the follow-up fixed for the fourth.

No behaviour delta today — the linted count is 282 either way. **Failure scenario:** someone adds `packages/x/jest.config.js` or a second `webpack.config.js`; it is linted under `**/*.js` with `sourceType: commonjs` + `globals.node`, and being a config file will likely trip `no-undef` or `no-unused-vars` on an unrelated PR, three commits after the cause. Either make all four `**/`-prefixed, or drop the "as `.eslintignore` did" framing and say the anchoring is deliberate for these three.

---

## Suggestions

- **Counts and versions in comments.** `:30` "and **a dozen more**" — the measured figure is 19 restored, so 16 more; and it is falsified by a typescript-eslint release with nobody watching. Use "…and other core checks the compiler would otherwise cover". Same for "v8" at `:50` and `:65` — a version's home is the manifest.

- **A two-item call-site inventory.** `:61-63` — "**The two sites** are deliberate: `Example.value` … and `typeOf()` …". Accurate today (verified: exactly 2, at `lib/src/definitions.ts:74` and `lib/src/parsers/schemaprop-parser.ts:155`) and exactly the shape the rule forbids — the third `any` falsifies it silently. Cut to "Warn: the remaining sites are deliberate."

- **Two stories about warnings, in one commit.** The comment above says warn is chosen so sites "report without needing individual suppressions", while the same commit adds `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` at `types.ts:591` — an individual suppression of a warning. Pick one.

- **Undocumented severity change.** `no-unused-vars` goes warn → error under v8 recommended. It passes clean, so no action — but the PR deliberately pinned `no-explicit-any` back to `warn` for exactly this reason and did not mention this one. A table row makes the asymmetry a choice.

- **Comment ratio on this file went 9% → 22%.** The additions are the largest proportional jump in the stack after `ts-project.ts`.

---

## Strengths

- **Every number in the commit message is reproducible.** 0 errors, 25 warnings byte-identical to the parent, 282 files on both sides — not a single overstated figure, and the two riskiest items (`allowObjectTypes` semantics, `**/build/` depth) are both load-bearing.
- **The `.js`/`.ts` split diagnoses a genuinely subtle failure.** `tseslint.config()` rewriting an extended config's `files` to the parent's is precisely the kind of thing that ships green: the 19 rules it silently disabled produce no errors on a clean tree, so nothing but reading the resolved config would have caught it.
- **The eslintrc → flat translation of basename-matching overrides is correct.** `.eslintrc.js`'s `files: ["*.spec.ts"]` matched at any depth; `**/*.spec.ts` reproduces that. Writing `*.spec.ts` in flat config (matching only the root) is the single most common silent-coverage-loss bug in these migrations, and the 282/282 parity confirms it was avoided.

vlj91 (approved):
🚀

PR #2710, opened 2026-08-18
fix(COMPASS-17): declare the oclif packages bin/run needs at runtime

timdawborn:
## Ticket

> **Stacked on #2709 → #2708.** Review those first. `build-and-test` does report on this PR — `test-node-{18,20,22,24}` and `lint-check` are green, so the new smoke-test step has run on all four Node versions.

## What

Declare `@oclif/errors` and `@oclif/help` as dependencies, and smoke-test the built CLI in CI.

## Why

**`spot`'s CLI cannot start under pnpm's default `node_modules` layout.** `bin/run` requires `@oclif/errors/handle`, and `@oclif/command` reaches for `@oclif/help` when it renders help or an error. Neither is declared in `package.json`; both have only ever resolved because a package manager hoisted them out of `@oclif/command`'s own tree.

```
$ node bin/run validate api.ts
Error: Cannot find module '@oclif/errors/handle'
```

That is on `master` today, in a clean `pnpm install --frozen-lockfile` checkout, before any change in this stack.

npm consumers are fine — npm's flat layout has been covering for the missing declarations, which is why nobody has hit it. It matters now because the docker image later in this ticket installs with `pnpm install --frozen-lockfile --prod`. Without this, the image would build green and then fail on every single invocation.

## The gap this exposes

The whole jest suite — 529 tests — passes against a CLI that cannot start. Every suite imports the library directly (`lib/src/...`); nothing loads `bin/run`, so nothing exercises the oclif entrypoint or the package's own dependency closure. Green CI was never evidence the binary worked.

Nor is a unit test the right guard: `@oclif/errors` is required by `bin/run` itself, but `@oclif/help` is pulled in by `@oclif/command` at render time, so a test that inspects `bin/run`'s `require()` calls catches one and misses the other. The guard has to invoke the binary.

So the fix ships with a CI step in `.github/actions/build-and-test/action.yml`, right after the existing `pnpm build`:

```yaml
pnpm install --frozen-lockfile --prod                   # prune the dev tree first
node bin/run --help                                    # exercises the @oclif/help path
node bin/run validate test-fixtures/contract/api.ts     # exercises a real command end to end
pnpm install --frozen-lockfile                          # restore for the jest step
```

**The install has to be `--prod`.** Under pnpm's isolated layout both dependency sets are symlinked into `node_modules`, so invoking the CLI after a full install shows only that a package is declared *somewhere* — which is true of a devDependency, and is exactly the case that breaks. Moving these two back to `devDependencies` leaves a full-install smoke test green while `--prod` dies with `MODULE_NOT_FOUND` on every invocation. Pruning first is what makes the step fail on a misdeclared runtime dependency.

`CI` is set explicitly on the step: `--prod` removes the modules directory to prune, and refuses to do that unattended without it. Actions sets it already, but the step should not depend on that.

It runs once per Node version rather than once per shard — the failure depends on the dependency set and the runtime, not on which quarter of the suite is running, and the prune is a full reinstall. Fails loudly via `set -uex`.

## How this was verified

- **Before:** on this branch's base, a clean install then `node bin/run --help` and `node bin/run validate …` both die with `Cannot find module '@oclif/errors/handle'`. So the failure is pre-existing, not introduced by anything in this stack.
- **The guard discriminates.** Under `--prod`: `jest` is pruned, both oclif packages remain, and `--help` and `validate` succeed. With their root links removed, the entrypoint dies with `MODULE_NOT_FOUND`.
- **After:** every command the docker image is planned to expose was run against a real contract:

  | Command | Result |
  |---|---|
  | `--help` | prints the full command table |
  | `validate` | `Contract is valid` |
  | `lint` | `Found 0 errors and 0 warnings` |
  | `checksum` | `ef62006f8d00d5899d7f0f03f3c1ca66d2b52be4` |
  | `generate -g openapi3 -l yaml -o …` | writes `api.yml` |
  | `validation-server … -p 5677` | see below |

- **`validation-server` specifically**, because two of the image's interface promises depend on it: with stdout redirected to a file — so **no TTY** — it printed `Validation server running on port 5677`, and `GET /health` answered `200`. That is the string bff-client's Gradle `ExecFork` `waitForOutput` waits on, and the check that it is line-buffered without a terminal.
- `pnpm lint:check` clean.

## Scope note

This was found while building `ts-lint` (next PR in the stack), which could not be exercised through the CLI at all until it was fixed. It is a separate concern from `ts-lint`, so it is a separate PR.

### DIFF
--- a/.github/actions/build-and-test/action.yml
+++ b/.github/actions/build-and-test/action.yml
@@ -19,16 +19,35 @@ runs:
     run: pnpm build
     shell: bash
 
-  - name: Smoke-test the built CLI
+  - name: Smoke-test the built CLI against a production-only install
     # Every jest suite imports the library directly, so none of them loads
-    # bin/run. A runtime dependency that resolves only because a package
-    # manager hoisted it therefore passes the whole suite and then kills every
-    # command on the first real invocation — which is what pnpm's isolated
-    # layout, and the docker image built from it, actually do.
+    # bin/run: a package the entrypoint needs can be missing from
+    # `dependencies` and the whole suite still passes.
+    #
+    # The install has to be `--prod` for this step to observe that. Under
+    # pnpm's isolated layout both dependency sets are symlinked into
+    # node_modules, so invoking the CLI after a full install shows only that a
+    # package is declared *somewhere* — which is true of a devDependency, and
+    # is exactly the case that breaks. Pruning first is what makes the step
+    # fail on a misdeclared runtime dependency.
+    #
+    # Restores the full install afterwards, because the jest step below needs
+    # it. `build/` predates the prune and is unaffected by both installs.
+    #
+    # Once per Node version: the failure depends on the dependency set and the
+    # runtime, not on which test shard this job happens to be running.
+    if: inputs.shard == '1'
+    env:
+      # `--prod` removes the modules directory to prune the dev tree, and
+      # refuses to do that unattended unless CI is set. Actions sets it
+      # already; naming it here keeps the step from depending on that.
+      CI: "true"
     run: |-
       set -uex
+      pnpm install --frozen-lockfile --prod
       node bin/run --help
       node bin/run validate test-fixtures/contract/api.ts
+      pnpm install --frozen-lockfile
     shell: bash
 
   - name: Create test reports directory
--- a/.github/actions/build-and-test/action.yml
+++ b/.github/actions/build-and-test/action.yml
@@ -19,6 +19,18 @@ runs:
     run: pnpm build
     shell: bash
 
+  - name: Smoke-test the built CLI
+    # Every jest suite imports the library directly, so none of them loads
+    # bin/run. A runtime dependency that resolves only because a package
+    # manager hoisted it therefore passes the whole suite and then kills every
+    # command on the first real invocation — which is what pnpm's isolated
+    # layout, and the docker image built from it, actually do.
+    run: |-
+      set -uex
+      node bin/run --help
+      node bin/run validate test-fixtures/contract/api.ts
+    shell: bash
+
   - name: Create test reports directory
     run: mkdir -p ./test-reports/jest
     shell: bash
--- a/.github/actions/build-and-test/action.yml
+++ b/.github/actions/build-and-test/action.yml
@@ -24,27 +24,21 @@ runs:
     # bin/run: a package the entrypoint needs can be missing from
     # `dependencies` and the whole suite still passes.
     #
-    # The install has to be `--prod` for this step to observe that. Under
-    # pnpm's isolated layout both dependency sets are symlinked into
-    # node_modules, so invoking the CLI after a full install shows only that a
-    # package is declared *somewhere* — which is true of a devDependency, and
-    # is exactly the case that breaks. Pruning first is what makes the step
-    # fail on a misdeclared runtime dependency.
-    #
-    # Restores the full install afterwards, because the jest step below needs
-    # it. `build/` predates the prune and is unaffected by both installs.
-    #
-    # Once per Node version: the failure depends on the dependency set and the
-    # runtime, not on which test shard this job happens to be running.
+    # The install has to be `--prod` for this step to observe that. Under pnpm's
+    # isolated layout both dependency sets are symlinked into node_modules, so
+    # invoking the CLI after a full install shows only that a package is
+    # declared somewhere — true of a devDependency, and exactly the case that
+    # breaks. Pruning first is what makes the step fail on a misdeclared runtime
+    # dependency, so the prune is asserted rather than assumed.
     if: inputs.shard == '1'
     env:
-      # `--prod` removes the modules directory to prune the dev tree, and
-      # refuses to do that unattended unless CI is set. Actions sets it
-      # already; naming it here keeps the step from depending on that.
+      # `--prod` removes the modules directory, and refuses to do that
+      # unattended without this. Actions sets it; not depending on that.
       CI: "true"
     run: |-
       set -uex
       pnpm install --frozen-lockfile --prod
+      test ! -e node_modules/jest
       node bin/run --help
       node bin/run validate test-fixtures/contract/api.ts
       pnpm install --frozen-lockfile
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
   "dependencies": {
     "@oclif/command": "^1.8.0",
     "@oclif/config": "^1.17.0",
+    "@oclif/errors": "^1.3.6",
+    "@oclif/help": "^1.0.15",
     "@oclif/plugin-help": "^3.2.3",
     "ajv": "^8.18.0",
     "ajv-formats": "^2.1.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,6 +18,12 @@ importers:
       '@oclif/config':
         specifier: ^1.17.0
         version: 1.18.17
+      '@oclif/errors':
+        specifier: ^1.3.6
+        version: 1.3.6
+      '@oclif/help':
+        specifier: ^1.0.15
+        version: 1.0.15
       '@oclif/plugin-help':
         specifier: ^3.2.3
         version: 3.3.1


timdawborn:
# Review — #2710 `fix(COMPASS-17): declare the oclif packages bin/run needs at runtime`

Automated multi-agent review (`/pr-review-toolkit:review-pr`), scoped to this PR's diff against its base (#2709). Findings marked **[verified]** were reproduced by extracting the PR tree, installing, building, and running the CLI under both a full and a `--prod` install.

---

## Important

### 1. The regression guard does not guard the property this PR fixes
**[verified]** · `.github/actions/build-and-test/action.yml:22-32`

The new smoke test runs after `prepare-repository`'s `pnpm install --frozen-lockfile` — a **full** install. Under pnpm's isolated layout, `dependencies` and `devDependencies` of the root importer are *both* symlinked into `node_modules/`, so the step cannot tell them apart. It proves "declared *somewhere*", not "declared as a runtime dependency" — and the second is the whole point of the PR (title: "…needs at **runtime**"; motivation: the image runs `pnpm install --frozen-lockfile --prod`).

Reproduced on the PR tree, with `@oclif/errors` and `@oclif/help` moved verbatim from `dependencies` into `devDependencies`:

```
# full install — exactly what CI does
$ pnpm install --frozen-lockfile && pnpm build && node bin/run --help
VERSION
  @airtasker/spot/2.0.3 …            # step is GREEN

# what the docker image does
$ CI=true pnpm install --frozen-lockfile --prod && node bin/run --help
node:internal/modules/cjs/loader:1522
  throw err;                          # dead on every invocation
```

**Failure scenario:** the next PR in the stack adds a command needing a package and the author reaches for `pnpm add -D`, or a future dependency sweep demotes one of these two. Build passes, `node bin/run --help` passes, the full jest suite passes, CI is green. The image then builds green and `spot <cmd>` fails at runtime — the exact class of bug this PR exists to eliminate, re-entering at the moment the ticket claims it closed.

**Fix** — make the guard install the way the image does. `typescript` is already a prod dependency, so build first, then prune:

```yaml
- name: Smoke-test the built CLI against a production-only install
  run: |-
    set -uex
    pnpm install --frozen-lockfile --prod
    node bin/run --help
    node bin/run validate test-fixtures/contract/api.ts
    pnpm install --frozen-lockfile   # restore devDeps for the jest step
  shell: bash
```

Verified working on the as-shipped tree: under `--prod`, `--help`, `validate`, `lint` and `checksum` all succeed, and `checksum` reproduces the `ef62006f8d00d5899d7f0f03f3c1ca66d2b52be4` digest from the PR body.

If a `--prod` step is out of scope here, the alternative that actually holds is a `depcheck`/`publint`-style assertion that `bin/run`'s resolution closure is a subset of `dependencies` — but the `--prod` install is cheaper and directly models the failure.

### 2. The step's comment claims more than the check enforces
`.github/actions/build-and-test/action.yml:23-27`

> "A runtime dependency that resolves only because a package manager hoisted it therefore passes the whole suite and then kills every command on the first real invocation — which is what pnpm's isolated layout, **and the docker image built from it**, actually do."

This reads as "this step reproduces the docker image's failure mode". It does not — per finding 1, it reproduces only the isolated-layout half. Per the org rule on defensive routines, state what the check observes: that `bin/run` resolves its closure under *this job's* install. As written, the next reviewer trusts the comment and skips adding the `--prod` guard.

Secondary: "and the docker image built from it" is a forward reference to something that does not exist in the repo at this commit — it lands two PRs later. A reader who arrives after merge has no navigable referent. Either drop the clause or point at a real ticket URL.

---

## Suggestion

- **The step runs 16 times per PR.** `action.yml:22` — the composite is called from a 4-node × 4-shard matrix. Only the Node-version axis carries signal; the shard axis is pure duplication. Gate it with `if: inputs.shard == '1'` (the `shard` input is already available). Cheap either way at ~2s, but the step is sharding-agnostic.

---

## Verified — no action needed

I was skeptical of `@oclif/help`, since it is a plain `dependencies` entry of `@oclif/command@1.8.36` in the lockfile and pnpm's isolated layout *should* satisfy it from `@oclif/command`'s own tree. It does not. Removing only the root `node_modules/@oclif/help` symlink from an otherwise-complete install gives:

```
Error: Could not load a help class, consider installing the @oclif/help package,
failed with message: Cannot find module '@oclif/help'
Require stack:
  - …/.pnpm/@oclif+help@1.0.15/node_modules/@oclif/help/lib/util.js
```

`@oclif/help`'s own `lib/util.js` re-resolves `@oclif/help` with `paths` rooted at the consuming CLI's package directory, so the root declaration is genuinely load-bearing. **Both additions are necessary; neither is over-declared.**

Version ranges are sound: `^1.3.6` → 1.3.6 and `^1.0.15` → 1.0.15, byte-identical to what `@oclif/command@1.8.36` already resolves in the lockfile, so no duplicate instance is introduced, and a caret on a `1.x` range cannot admit a major. The lockfile diff touches only the `importers` block — both packages were already in the tree, so there is no new resolution and no added supply-chain surface.

No `npm`/`yarn`/`bun` anywhere in `.github/**` or `package.json`. No bare ticket key in any committed file — `COMPASS-17` appears only in the commit subject and PR body, which is where the rule wants it. No `TODO`/`FIXME` in the diff.

**Out of scope but noted:** `.github/actions/prepare-repository/action.yml` pins `pnpm/action-setup@v4` while `.github/workflows/build-and-test.yml` pins `@v6.0.9` — same action, two majors apart, neither SHA-pinned. Both hunks predate this PR.

## Note on CI

The body says `build-and-test` will not report here. **It has** — `test-node-{18,20,22,24}` and `lint-check` are green on this PR, which means the new smoke-test step *did* execute on all four Node versions. That is useful evidence in this PR's favour, and worth correcting in the body.

## Strengths

- The diagnosis is correct and non-obvious, and the body is honest about *why* 529 green tests were never evidence — plus why a unit test inspecting `bin/run`'s `require()` calls would catch `@oclif/errors` and miss `@oclif/help`. That reasoning holds up under testing.
- `node bin/run --help` is a better guard than it looks: with no `oclif.manifest.json` present, `@oclif/config` scans and `require`s every command module to read its statics, so one invocation exercises the full static import closure of all nine commands, not just the oclif bootstrap.
- Every verification claim in the body that could be re-run reproduced exactly, including the `checksum` digest.
- Correctly split out from #2711 as its own atomic PR rather than bundled into the `ts-lint` work that surfaced it.

timdawborn:
# Second-pass review — #2710

Multi-agent re-review of both commits, with the follow-up `a54bbe5` reviewed adversarially. Verified against pnpm 10.28.1's source, the real CI logs for run `32103875464`, and live worktree reproductions.

**Verdict: the fix is sound. All five falsification targets came back falsified**, and the whole step costs 6.894s. The findings are about the guard's own preconditions being unasserted — the same class of gap the PR exists to close.

---

## Falsification results

| Target | Verdict | Evidence |
|---|---|---|
| Restore not guaranteed under `set -e` | **Falsified** | A failing step fails the composite action and every remaining step is skipped; no `if: always()` or `continue-on-error` anywhere. Jest never runs against a pruned tree, so there is no misleading second failure. `--prod` removes nothing from the pnpm store that `setup-node` caches. |
| `if: inputs.shard == '1'` never true | **Falsified — it works** | Log evidence, not inference: of the 16 test jobs, exactly 4 executed the body (the `--shard=1/4` job per Node version) and 12 registered-and-skipped. Shard 1 shows `Packages: -757` with `jest`, `ts-jest`, `jest-junit`, `@types/jest`, `eslint-plugin-jest` removed, both invocations, then `Packages: +773`. |
| Prune/restore breaks downstream | **Falsified** | Shard 1 went on to pass jest (80 tests, 70.1s) and `pnpm build-docs` (16.0s). Neither root nor `docs/package.json` has `prepare` or `postinstall`, so there is no fresh-install-only step to lose. |
| `CI: "true"` unnecessary | **Confirmed required** | `env -u CI -u GITHUB_ACTIONS pnpm install --frozen-lockfile --prod < /dev/null` → `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY`. Path in pnpm's source: `validateModules` → `INCLUDED_DEPS_CONFLICT` for the `docs` workspace project → `purgeModulesDirsOfImporters`, gated by `confirmModulesPurge && !opts.ci`. Inert under Actions, genuinely load-bearing on any non-Actions runner. |
| Cost | **Falsified — gating is more than sufficient** | Step total **6.894s**: prune 1.1s, `--help` 0.8s, `validate` 2.2s, restore 2.7s, against a 125s job. The store is warm, so both installs are hardlink relinks. |

Both declarations verified load-bearing **in isolation**, not just together: removing only the root `@oclif/errors` link kills `bin/run` with `MODULE_NOT_FOUND`; removing only the root `@oclif/help` link produces `Error: Could not load a help class…` on `--help`, `help <cmd>` and bare invocation. Neither is noise. Versions dedupe to the exact instances `@oclif/command` already resolves (`errors 1.3.6`, `help 1.0.15`), so no second module instance and no `instanceof` hazard.

---

## Important

### 1. Nothing asserts the prune happened, so the step can silently revert to the no-op it was written to fix
`.github/actions/build-and-test/action.yml:44-49`

The step's entire value rests on `--prod` having removed the dev tree. If it ever doesn't — a pnpm version treating `--prod` as advisory on a complete tree, `CI` becoming unset so pnpm warns instead of removing, a `node-linker` or workspace change, a `dependenciesMeta` entry retaining a devDependency's root link — both invocations run against a full tree and pass green while proving nothing. That is verbatim the original bug this commit names: *"the smoke test ran after a full install, which cannot observe the property this change is about."* Green looks identical in both worlds.

The inconsistency is inside the same stack: #2713's script asserts every one of its own preconditions (`if [ ${#DECLARED[@]} -eq 0 ]`, `if [ -z "$(ls -A "${CLI_DIR}")" ]`, and self-containment "assert that first"). This step asserts none.

```bash
pnpm install --frozen-lockfile --prod
test ! -e node_modules/jest || {
  echo "--prod left the dev tree in place; this step cannot observe a misdeclared runtime dependency" >&2
  exit 1
}
```

### 2. The guard's *execution* is keyed to a test-sharding literal, and nothing fails if it never matches
`.github/actions/build-and-test/action.yml:39` — `if: inputs.shard == '1'`

This PR's thesis is that green CI was never evidence the binary worked. The gating reintroduces that shape one level up: whether the smoke test runs now depends on a string literal matching a matrix value in a **different file**, and no step anywhere fails if it stops matching.

**Failure scenario:** a later PR drops sharding (`jest -w 4` already parallelises; `--shard` is the newer half) and removes the `shard` input, or renames it, or a second caller of this composite omits it. GitHub does **not** enforce `required: true` on composite-action inputs — it warns at most — so `inputs.shard` becomes `''`, the step silently never runs on any job, CI stays green, and the docker image later in this ticket ships with the exact `MODULE_NOT_FOUND` this PR prevents.

Given the step is only 7s, the gating buys ~21s of runner time for that fragility. **The better shape dissolves findings 1, 2 and the restore concern together:** lift it into its own top-level job in `build-and-test.yml`. It needs no matrix, no sharding, and no prune/restore — a job that installs `--prod` from scratch, builds, and invokes the binary. It is also then independently visible in the checks list and separately promotable to a required status check.

### 3. `dependency-review` is red on the one PR in the stack that adds dependencies, and the body lists only the green checks
Run `32103875540`: `Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled`. It fails identically on #2708 and #2709, so it is pre-existing and environmental — but this is the PR that adds runtime dependencies, and the check that would have vetted them is inoperative. Both new packages are marked `deprecated: Package no longer supported` in the lockfile, which is exactly what that check exists to surface.

Per `pull-requests.md`, either note it explicitly with the reason it is not actionable here, or link a follow-up ticket to enable the dependency graph on `airtasker/spot`.

---

## Suggestions

- **No assertion that there is no `node_modules` above the checkout.** A `node_modules` on any parent path satisfies resolution and makes the step an unconditional pass. This was hit during the audit: with a stray `node_modules` one directory up, deleting **both** `@oclif/errors` and `@oclif/help` still gave `--help` and `validate` exit 0 — Node resolved them from the parent's `.pnpm` store, confirmed via `require.resolve`. Latent on `ubuntu-latest` (`/home/runner/work/spot/spot` has no such parent); live the moment the job runs in a container image shipping a global `node_modules`, or on a warm self-hosted runner. #2713 asserts exactly this for its equivalent check.

- **`node bin/run --help` is silently load-bearing.** It is the **only** line covering `@oclif/help` (`validate` exits 0 without it, verified). The comment explains why the install must be `--prod` but not why there are two invocations, so a later trim of `--help` as redundant drops coverage for one of the two packages this PR adds, with nothing going red. One clause on that line.

- **The comment states an unconditional restore that `set -e` prevents.** `:33` — "Restores the full install afterwards, because the jest step below needs it." If either invocation fails, the restore never runs. Harmless in practice (finding above), but reword to "on success", or make it an `if: always()` step.

- **Comment ratio on this file went 10% → 26%** — 20 comment lines above a 5-line script. The `build/` sentence defends a non-risk (nothing in the script touches `build/`, and there is no `prepare`/`postinstall`), the "Once per Node version" paragraph restates the `if:` line, and the three-line `CI:` justification is longer than the thing it justifies. Keep the first paragraph and "Pruning first is what makes the step fail on a misdeclared runtime dependency."

- **Body inaccuracy, carried from the original commit:** *"`@oclif/command` reaches for `@oclif/help` when it renders help **or an error**"* — the second half does not hold. With the root `@oclif/help` link removed, `node bin/run definitely-not-a-command` and `validate --bogus-flag` both render their errors fine; only `--help`, `help <cmd>` and bare invocation fail. The design conclusion is unaffected.

- Promoting two deprecated packages to direct `dependencies` makes the eventual `@oclif/core` migration a four-package change. Correct and necessary here — worth a sentence rather than a change.

- One real (pre-existing) consequence of `fail-fast: true`: a smoke-test failure on one Node version now cancels the other 15 matrix jobs, so a genuine test failure elsewhere is discarded rather than reported. A new trigger for existing behaviour, not a new bug.

---

## Strengths

- **The `--prod` insight is the correct fix and the non-obvious half.** A full-install smoke test genuinely cannot observe this under pnpm's isolated layout, and the commit message records the differential that proves it — deps moved back to `devDependencies` leaves the step green while `--prod` dies.
- **Every verification claim in the body reproduced**, including the `checksum` digest and the pnpm source path for the `CI` requirement.
- The diagnosis of *why* 529 green tests were never evidence — and why a unit test on `bin/run`'s `require()` calls would catch one package and miss the other — holds up under testing.

vlj91 (approved):
LGTM

PR #2711, opened 2026-08-18
feat(COMPASS-17): add ts-lint command with bundled canonical config

timdawborn:
## Ticket

> **Stacked on #2710 → #2709 → #2708.** Review those first. `build-and-test` does report on this PR — `test-node-{18,20,22,24}` and `lint-check` are green.

## What

`spot ts-lint [directory] [--fix]` — checks a tree of Spot contracts for formatting, lint and type errors, under configuration Spot carries itself.

```
$ spot ts-lint spots
Formatting (2):
  /repo/spots/api.ts
  /repo/spots/models.ts
  Run with --fix to reformat.
Lint (1):
  /repo/spots/models.ts:27:7 error Variable name `DEFAULT_price` must match one of the following formats: camelCase, PascalCase (@typescript-eslint/naming-convention)
Types (1):
  /repo/spots/models.ts:33 - Type 'string' is not assignable to type 'number'.
```

## Why

This is what lets the three consumer repos delete their contract Node toolchains. Today each one installs prettier, eslint, typescript-eslint and a tsconfig purely to check contracts, and carries a near-identical `eslint.config.mjs` and `.prettierrc.yml` to configure them. Bundling the tools *and* the configuration into Spot means a consumer needs none of it — just the container.

## Design

**Three steps, all of which run even when an earlier one has already found something.** A contributor who runs this once should see everything wrong with the tree, not the first category of thing.

1. **Prettier**, via `prettier/standalone` with the TypeScript and estree plugins passed explicitly. Neither config discovery nor plugin resolution is used, so a `.prettierrc` left behind in a consumer repo cannot change how contracts are formatted. (This also avoids the dynamic `import()` in prettier's default CJS entrypoint, which does not work under jest's VM without `--experimental-vm-modules`.)
2. **ESLint**, via the programmatic API with `overrideConfigFile: true` and the config as an **in-code value**, not a path. Its plugins are the ones the config module imported, so they resolve from Spot's own `node_modules`; there is no `--config` path semantics and no `resolvePluginsRelativeTo`, and a leftover `eslint.config.mjs` beside the contract is neither read nor needed.
3. **A type check through ts-morph**, under the same compiler options `parse` uses — including the `@airtasker/spot` mapping fixed in #2708. There is no emit by construction; the diagnostics are the whole product. This is what replaces rails-monolith's `pnpm tsc --outDir output` gate, whose `output/` directory nothing ever consumed.

**Exit codes.** 0 when clean or fully fixed; 1 for any remaining formatting difference, lint error, or type error, and 1 when the directory holds no `.ts` files at all — a renamed tree or an empty bind mount must not read as a pass. Set through `process.exitCode`, not `process.exit`: `this.log` writes asynchronously when stdout is a pipe, so exiting outright truncated the report (three identical runs of a 1601-line report printed 770, 1601 and 577 lines). Warnings are printed and never fail, and "No problems found" is withheld whenever there is a report to show.

**File discovery is a walk of the tree, not of the import graph** — every `.ts` under the directory, skipping `node_modules` and dot-directories. A contract file that nothing imports yet is exactly the file most likely to be unformatted, and it must not pass by being invisible. There is a test for this.

**Under `--fix`, ESLint's fixes are applied before Prettier writes.** An ESLint fix produces code, not layout, so Prettier has to be the one to lay out whatever it produced. Doing it the other way round leaves a tree that `--fix` claims to have fixed and that fails on the next run.

## The canonical config

Reconciled from the three consumers' real configs, not designed fresh.

**Formatting.** All three consumers already share an identical `.prettierrc.yml`: `trailingComma: none`, `tabWidth: 2`, `semi: true`, `singleQuote: true`, `printWidth: 80`. That is the canonical config. It is worth being explicit that this is **not** Prettier's default — the plan for this ticket called for defaults plus an accepted one-off reformat, but adopting what the consumers already use means no reformat at all in two of the three repos, and removes the "reformat races contract churn" risk from the plan's risk table.

**Rules.** The union of the three configs' rule-offs, plus the `naming-convention` rule that bff-client and enrichment-engine carry and rails-monolith does not. `@typescript-eslint/ban-types` was split in typescript-eslint 8, so its single `off` becomes `no-empty-object-type`, `no-wrapper-object-types` and `no-unsafe-function-type`.

On `no-wrapper-object-types` specifically, the reasoning in the first version of this PR was wrong and the comment has been corrected. A contract imports `String` and `Boolean` from `@airtasker/spot`, and that import shadows the global — so the rule has nothing to say about a well-formed contract either way. Turned on against the clean fixture, which uses `String` in path params, headers and query params, it reports nothing. What the exemption actually suppresses is the opposite case: a file naming the global because the import is missing. That is a real mistake, and it is caught where it has consequences, by the parser rejecting the type.

**`eslint-plugin-prettier` is dropped.** Formatting failures surface from the Prettier step instead of a second time as a lint rule. Same gate, one report each.

**Prettier moves 3.3.2 → 3.9.6**, the version the consumers are on. 3.9 collapses short unions onto one line, so this reformats five of Spot's own files.

## How this was verified

**Against the three real consumer contract trees.** This is the reconciliation the plan requires before any consumer migration PR is written, so the canonical config is fixed here:

| Repo | Files | Formatting | Lint | Types |
|---|---|---|---|---|
| `rails-monolith/spots` | 230 | clean | clean | clean |
| `bff-client/…/spots` | 345 | **29 reformat** | clean | clean |
| `enrichment-engine/spots` | 4 | clean | clean | clean |

- **No consumer gains a lint error** — including rails-monolith, which does not have `naming-convention` today and passes it anyway.
- **No consumer gains a type error**, so the type step can replace rails' `tsc` gate without introducing failures. (Expected: `validate`/`generate` already parse with `strict: true`.)
- bff-client's 29 files are the same union-wrapping change and nothing else; it is pinned to prettier 3.8.3. That is the whole reformat cost of this migration, across all three repos.
- The plan predicted enrichment-engine would take "the largest style delta — 4-space→canonical". It takes none; its contracts are already 2-space. The 4-space file is its `eslint.config.mjs`, which is being deleted anyway.

**Other checks:**

- `pnpm test` — 541 tests across 50 suites pass, including 12 new ones.
- **The Prettier bump changes no behaviour**: compiled `build/` output is byte-identical to the parent branch apart from the new `ts-lint` files.
- `node bin/run ts-lint test-fixtures/ts-lint/clean` exits 0. `dirty` exits 1 reporting all three categories from one run. `--fix` then clears the formatting and leaves the naming-convention error and the type error standing — neither is something a fixer may decide on the author's behalf — and the command still exits 1.
- A test writes a `.prettierrc` with `singleQuote: false` and an `eslint.config.mjs` into the fixture tree and asserts the outcome is unchanged. That is the load-bearing claim of the in-code-config design, so it has a test rather than a comment. **The ESLint half of that test needed a different rule to mean anything:** it originally used `no-undef`, which `tseslint.configs.recommended` switches off again from inside `overrideConfig`, so it reported nothing whether or not the leftover config was read. It now uses `no-restricted-syntax` on `TSInterfaceDeclaration`, a rule the bundled config never mentions — verified to report 5 findings on the clean fixture when applied, and 0 as shipped.
- A test asserts a non-`--fix` run does not write to the tree.

## Notes

- `test-fixtures/ts-lint/` is excluded from Spot's own prettier and eslint runs. Those trees are held to the *canonical contract* config, which differs from this repo's own (single vs double quotes), and `dirty/` is deliberately broken.
- **No TypeScript bump.** The plan budgeted `typescript` 4.9 → 5 as this PR's riskiest commit, on the assumption typescript-eslint 8 would force it. Its peer range is `>=4.8.4 <6.1.0`, so it does not. Likewise the eslint `^9` fallback the plan designed for is not needed — `^10` works.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,5 @@
 build/
+
+# Held to the canonical contract formatting that ts-lint bundles, which is not
+# this repo's own; dirty/ is deliberately unformatted.
+test-fixtures/ts-lint/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,7 @@ Spot is a TypeScript-based API contract definition tool that generates OpenAPI, 
 ### CLI Usage (after build)
 - `npx @airtasker/spot generate --contract api.ts --generator openapi3 --out output/` - Generate OpenAPI3 from contract
 - `npx @airtasker/spot lint api.ts` - Lint a Spot contract
+- `npx @airtasker/spot ts-lint spots` - Check the TypeScript in a contract tree for formatting, lint and type errors, under configuration Spot bundles itself (`--fix` rewrites in place)
 - `npx @airtasker/spot mock api.ts` - Run mock server from contract
 - `npx @airtasker/spot validate api.ts` - Validate contract syntax
 
--- a/cli/src/commands/ts-lint.spec.ts
+++ b/cli/src/commands/ts-lint.spec.ts
@@ -30,8 +30,10 @@ describe("ts-lint command", () => {
   });
 
   afterEach(() => {
-    writeSpy.mockRestore();
+    // Restored first: a throw from mockRestore would otherwise leak a non-zero
+    // exit code, which reports as a wholly green suite that exits 1.
     process.exitCode = restoreExitCode;
+    writeSpy.mockRestore();
   });
 
   test("exits zero and says so on a clean tree", async () => {
@@ -53,9 +55,6 @@ describe("ts-lint command", () => {
   });
 
   test("exits one when the directory holds no contracts", async () => {
-    // A tree that has been renamed or arrived empty reaches this path. If it
-    // exited zero the gate would pass while checking nothing, which is the one
-    // failure a consumer's green pipeline could never show them.
     const empty = fs.mkdtempSync(path.join(os.tmpdir(), "spot-ts-lint-empty-"));
 
     await TsLint.run([empty]);
--- a/cli/src/commands/ts-lint.spec.ts
+++ b/cli/src/commands/ts-lint.spec.ts
@@ -1,0 +1,66 @@
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import TsLint from "./ts-lint";
+
+const FIXTURES = path.join(__dirname, "../../../test-fixtures/ts-lint");
+
+/**
+ * The exit code is the whole of this command's contract with CI: a consumer's
+ * pipeline reads nothing else. It is also the part a unit test on `tsLint`
+ * cannot reach, because the mapping from an outcome to an exit lives here.
+ */
+describe("ts-lint command", () => {
+  jest.setTimeout(60000);
+
+  let restoreExitCode: number | undefined;
+  let out: string;
+  let writeSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    restoreExitCode = process.exitCode;
+    process.exitCode = undefined;
+    out = "";
+    writeSpy = jest
+      .spyOn(process.stdout, "write")
+      .mockImplementation((chunk: unknown) => {
+        out += String(chunk);
+        return true;
+      });
+  });
+
+  afterEach(() => {
+    writeSpy.mockRestore();
+    process.exitCode = restoreExitCode;
+  });
+
+  test("exits zero and says so on a clean tree", async () => {
+    await TsLint.run([path.join(FIXTURES, "clean")]);
+
+    expect(process.exitCode).toBeUndefined();
+    expect(out).toContain("No problems found");
+  });
+
+  test("exits one on a tree with problems, reporting every category", async () => {
+    await TsLint.run([path.join(FIXTURES, "dirty")]);
+
+    expect(process.exitCode).toBe(1);
+    expect(out).toContain("Formatting (");
+    expect(out).toContain("Lint (");
+    expect(out).toContain("Types (");
+    // A run that found something must not also claim the tree is clean.
+    expect(out).not.toContain("No problems found");
+  });
+
+  test("exits one when the directory holds no contracts", async () => {
+    // A tree that has been renamed or arrived empty reaches this path. If it
+    // exited zero the gate would pass while checking nothing, which is the one
+    // failure a consumer's green pipeline could never show them.
+    const empty = fs.mkdtempSync(path.join(os.tmpdir(), "spot-ts-lint-empty-"));
+
+    await TsLint.run([empty]);
+
+    expect(process.exitCode).toBe(1);
+    expect(out).toContain("Nothing was checked");
+  });
+});
--- a/cli/src/commands/ts-lint.ts
+++ b/cli/src/commands/ts-lint.ts
@@ -41,10 +41,19 @@ export default class TsLint extends Command {
     fixed.forEach(file => this.log(`Fixed ${file}`));
     report.forEach(line => this.log(line));
 
-    if (ok) {
+    // Only claim the tree is clean when there is nothing to show. A report can
+    // be non-empty on a passing run — warnings do not fail the command — and
+    // printing both the findings and "No problems found" contradicts itself.
+    if (ok && report.length === 0) {
       this.log("No problems found");
-    } else {
-      process.exit(1);
+    }
+
+    if (!ok) {
+      // Not process.exit: this.log writes to stdout, which is asynchronous
+      // when stdout is a pipe — every CI runner — and exiting outright
+      // truncates the report mid-stream. Setting the code lets Node flush and
+      // exit on its own.
+      process.exitCode = 1;
     }
   }
 }
--- a/cli/src/commands/ts-lint.ts
+++ b/cli/src/commands/ts-lint.ts
@@ -1,0 +1,50 @@
+import { Command, flags } from "@oclif/command";
+import { tsLint } from "../../../lib/src/ts-lint/ts-lint";
+
+const ARG_DIR = "directory";
+
+/**
+ * oclif command to check a tree of Spot contracts for formatting, lint and
+ * type errors, under configuration bundled with Spot itself.
+ */
+export default class TsLint extends Command {
+  static description =
+    "Check the TypeScript in a Spot contract tree for formatting, lint and type errors";
+
+  static examples = ["$ spot ts-lint", "$ spot ts-lint spots --fix"];
+
+  static args = [
+    {
+      name: ARG_DIR,
+      required: false,
+      default: ".",
+      description: "directory to check",
+      hidden: false
+    }
+  ];
+
+  static flags: flags.Input<flags.Output> = {
+    help: flags.help({ char: "h" }),
+    fix: flags.boolean({
+      description: "Reformat and apply lint fixes in place",
+      default: false
+    })
+  };
+
+  async run(): Promise<void> {
+    const { args, flags } = this.parse(TsLint);
+
+    const { report, fixed, ok } = await tsLint(args[ARG_DIR], {
+      fix: flags.fix
+    });
+
+    fixed.forEach(file => this.log(`Fixed ${file}`));
+    report.forEach(line => this.log(line));
+
+    if (ok) {
+      this.log("No problems found");
+    } else {
+      process.exit(1);
+    }
+  }
+}
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -19,7 +19,10 @@ module.exports = tseslint.config(
       "**/build/",
       "docs/webpack.config.js",
       "jest.config.js",
-      "jest.ci.config.js"
+      "jest.ci.config.js",
+      // Held to the canonical contract config that `ts-lint` bundles, which
+      // differs from this repo's own. `dirty/` is deliberately broken.
+      "test-fixtures/ts-lint/"
     ]
   },
   {
--- a/lib/src/generators/openapi2/openapi2-specification.ts
+++ b/lib/src/generators/openapi2/openapi2-specification.ts
@@ -101,24 +101,19 @@ export type HeaderObject =
   | ArrayHeaderObject;
 
 export interface StringHeaderObject
-  extends HeaderObjectBase,
-    StringParameterObject {}
+  extends HeaderObjectBase, StringParameterObject {}
 
 export interface NumberHeaderObject
-  extends HeaderObjectBase,
-    NumberParameterObjectType {}
+  extends HeaderObjectBase, NumberParameterObjectType {}
 
 export interface IntegerHeaderObject
-  extends HeaderObjectBase,
-    IntegerParameterObjectType {}
+  extends HeaderObjectBase, IntegerParameterObjectType {}
 
 export interface BooleanHeaderObject
-  extends HeaderObjectBase,
-    BooleanParameterObjectType {}
+  extends HeaderObjectBase, BooleanParameterObjectType {}
 
 export interface ArrayHeaderObject
-  extends HeaderObjectBase,
-    ArrayParameterObjectType {}
+  extends HeaderObjectBase, ArrayParameterObjectType {}
 
 interface HeaderObjectBase {
   description?: string;
@@ -172,16 +167,13 @@ interface ParameterObjectBase {
 }
 
 type QueryParameterObjectType =
-  | BaseParameterObjectTypes
-  | ArrayMultiParameterObjectType;
+  BaseParameterObjectTypes | ArrayMultiParameterObjectType;
 
 type HeaderParameterObjectType =
-  | BaseParameterObjectTypes
-  | ArrayParameterObjectType;
+  BaseParameterObjectTypes | ArrayParameterObjectType;
 
 type PathParameterObjectType =
-  | BaseParameterObjectTypes
-  | ArrayParameterObjectType;
+  BaseParameterObjectTypes | ArrayParameterObjectType;
 
 type FormDataParameterObjectType =
   | BaseParameterObjectTypes
@@ -204,15 +196,13 @@ export interface StringParameterObject {
   enum?: string[];
 }
 
-export interface NumberParameterObjectType
-  extends NumberParameterObjectTypeBase {
+export interface NumberParameterObjectType extends NumberParameterObjectTypeBase {
   type: "number";
   format?: "float" | "double";
   default?: number;
 }
 
-export interface IntegerParameterObjectType
-  extends NumberParameterObjectTypeBase {
+export interface IntegerParameterObjectType extends NumberParameterObjectTypeBase {
   type: "integer";
   format?: "int32" | "int64";
   default?: number;
@@ -232,8 +222,7 @@ export interface ArrayParameterObjectType extends ArrayParameterObjectTypeBase {
   collectionFormat?: "csv" | "ssv" | "tsv" | "pipes";
 }
 
-export interface ArrayMultiParameterObjectType
-  extends ArrayParameterObjectTypeBase {
+export interface ArrayMultiParameterObjectType extends ArrayParameterObjectTypeBase {
   collectionFormat?: "csv" | "ssv" | "tsv" | "pipes" | "multi";
 }
 
@@ -306,29 +295,25 @@ export type OAuth2SecuritySchemeObject =
   | AccessCodeOAuth2SecuritySchemeObject;
 
 export interface ImplicitOAuth2SecuritySchemeObject
-  extends OAuth2SecuritySchemeObjectBase,
-    SecuritySchemeObjectBase {
+  extends OAuth2SecuritySchemeObjectBase, SecuritySchemeObjectBase {
   flow: "implicit";
   authorizationUrl: string;
 }
 
 export interface PasswordOAuth2SecuritySchemeObject
-  extends OAuth2SecuritySchemeObjectBase,
-    SecuritySchemeObjectBase {
+  extends OAuth2SecuritySchemeObjectBase, SecuritySchemeObjectBase {
   flow: "password";
   tokenUrl: string;
 }
 
 export interface ApplicationOAuth2SecuritySchemeObject
-  extends OAuth2SecuritySchemeObjectBase,
-    SecuritySchemeObjectBase {
+  extends OAuth2SecuritySchemeObjectBase, SecuritySchemeObjectBase {
   flow: "application";
   tokenUrl: string;
 }
 
 export interface AccessCodeOAuth2SecuritySchemeObject
-  extends OAuth2SecuritySchemeObjectBase,
-    SecuritySchemeObjectBase {
+  extends OAuth2SecuritySchemeObjectBase, SecuritySchemeObjectBase {
   flow: "accessCode";
   authorizationUrl: string;
   tokenUrl: string;
@@ -398,15 +383,13 @@ interface SchemaObjectBase {
 }
 
 export interface NumberSchemaObject
-  extends SchemaObjectBase,
-    NumberSchemaObjectBase {
+  extends SchemaObjectBase, NumberSchemaObjectBase {
   type: "number";
   format?: "float" | "double";
 }
 
 export interface IntegerSchemaObject
-  extends SchemaObjectBase,
-    NumberSchemaObjectBase {
+  extends SchemaObjectBase, NumberSchemaObjectBase {
   type: "integer";
   format?: "int32" | "int64";
 }
--- a/lib/src/generators/openapi3/openapi3-specification.ts
+++ b/lib/src/generators/openapi3/openapi3-specification.ts
@@ -158,15 +158,13 @@ interface SchemaObjectBase {
 }
 
 export interface NumberSchemaObject
-  extends SchemaObjectBase,
-    NumberSchemaObjectBase {
+  extends SchemaObjectBase, NumberSchemaObjectBase {
   type: "number";
   format?: "float" | "double";
 }
 
 export interface IntegerSchemaObject
-  extends SchemaObjectBase,
-    NumberSchemaObjectBase {
+  extends SchemaObjectBase, NumberSchemaObjectBase {
   type: "integer";
   format?: "int32" | "int64";
 }
@@ -222,8 +220,7 @@ export interface ObjectSchemaObject extends SchemaObjectBase {
 
 export interface ObjectPropertiesSchemaObject {
   [name: string]:
-    | (SchemaObject & ObjectPropertySchemaObjectBase)
-    | ReferenceObject;
+    (SchemaObject & ObjectPropertySchemaObjectBase) | ReferenceObject;
 }
 
 interface ObjectPropertySchemaObjectBase {
@@ -285,8 +282,7 @@ export interface OAuth2SecuritySchemeObject extends SecuritySchemeObjectBase {
   flows: OAuthFlowsObject;
 }
 
-export interface OpenIdConnectSecuritySchemeObject
-  extends SecuritySchemeObjectBase {
+export interface OpenIdConnectSecuritySchemeObject extends SecuritySchemeObjectBase {
   type: "openIdConnect";
   openIdConnectUrl: string;
 }
--- a/lib/src/parsers/__spec-examples__/types.ts
+++ b/lib/src/parsers/__spec-examples__/types.ts
@@ -42,9 +42,7 @@ interface TypeInterface {
   union: boolean | Date | null;
   unionDiscriminated: DiscriminatedUnionElementA | DiscriminatedUnionElementB;
   unionDiscriminatedNullable:
-    | DiscriminatedUnionElementA
-    | DiscriminatedUnionElementB
-    | null;
+    DiscriminatedUnionElementA | DiscriminatedUnionElementB | null;
   aliasString: AliasString;
   aliasArray: AliasArray;
   aliasWithDescription: AliasWithDescription;
--- a/lib/src/ts-lint/eslint-config.ts
+++ b/lib/src/ts-lint/eslint-config.ts
@@ -1,0 +1,46 @@
+import js from "@eslint/js";
+import { Linter } from "eslint";
+import prettierConfig from "eslint-config-prettier/flat";
+import tseslint from "typescript-eslint";
+
+/**
+ * The lint rules Spot contracts are held to.
+ *
+ * This is a value, not a config file path. `ts-lint` hands it to ESLint with
+ * `overrideConfigFile: true`, so the plugins are the ones this module
+ * imported — resolved from Spot's own `node_modules` — and an
+ * `eslint.config.*` sitting next to a contract is neither read nor needed.
+ *
+ * Contract syntax is why most of the rules below are off: an endpoint is a
+ * class nobody references, holding decorated methods with empty bodies, whose
+ * parameters exist to be read by the parser rather than by the code.
+ */
+export const eslintConfig: Linter.Config[] = tseslint.config(
+  js.configs.recommended,
+  tseslint.configs.recommended,
+  // Formatting is the Prettier step's job. Without this, a rule that also has
+  // an opinion about layout reports a second, differently-worded complaint
+  // about a line Prettier has already flagged.
+  prettierConfig,
+  {
+    rules: {
+      "@typescript-eslint/naming-convention": [
+        "error",
+        { selector: "default", format: ["camelCase", "PascalCase"] },
+        // Header and query-parameter names are wire names — `x-auth-token`,
+        // `sample-query` — and are not the contract author's to choose.
+        { selector: "property", format: null }
+      ],
+      "@typescript-eslint/no-inferrable-types": "off",
+      "@typescript-eslint/no-use-before-define": "off",
+      "@typescript-eslint/no-empty-function": "off",
+      "@typescript-eslint/explicit-function-return-type": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/no-empty-object-type": "off",
+      // `String`, `Number` and `Boolean` are Spot's own contract types, not the
+      // JavaScript wrapper objects these rules exist to catch.
+      "@typescript-eslint/no-wrapper-object-types": "off",
+      "@typescript-eslint/no-unsafe-function-type": "off"
+    }
+  }
+) as Linter.Config[];
--- a/lib/src/ts-lint/eslint-config.ts
+++ b/lib/src/ts-lint/eslint-config.ts
@@ -37,8 +37,13 @@ export const eslintConfig: Linter.Config[] = tseslint.config(
       "@typescript-eslint/explicit-function-return-type": "off",
       "@typescript-eslint/no-unused-vars": "off",
       "@typescript-eslint/no-empty-object-type": "off",
-      // `String`, `Number` and `Boolean` are Spot's own contract types, not the
-      // JavaScript wrapper objects these rules exist to catch.
+      // A contract imports `String`, `Number` and `Boolean` from
+      // `@airtasker/spot`, and that import shadows the global — so on a
+      // well-formed contract this rule has nothing to say either way. What
+      // turning it off suppresses is the opposite case: a file that names the
+      // global because the import is missing. That is a real mistake, and it is
+      // caught where it has consequences, by the parser rejecting the type,
+      // rather than as a wrapper-object complaint here.
       "@typescript-eslint/no-wrapper-object-types": "off",
       "@typescript-eslint/no-unsafe-function-type": "off"
     }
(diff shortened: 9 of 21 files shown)


timdawborn:
# Review — #2711 `feat(COMPASS-17): add ts-lint command with bundled canonical config`

Automated multi-agent review (`/pr-review-toolkit:review-pr`): general code review, test-coverage analysis, error-handling analysis, type-design analysis and comment-accuracy analysis. Scoped to this PR's diff against its base (#2710). Findings marked **[verified]** were reproduced by building the branch and running the code — mutation-testing the config, piping stdout, and probing fixtures.

This is the largest PR in the stack and the one that adds new public surface, so it drew the most agents. The design is sound; the findings cluster on two themes — **a gate that can pass without checking anything**, and **a canonical config that almost nothing tests**.

---

## Critical

### 1. A directory with no `.ts` files exits 0
**[verified — four agents independently]** · `lib/src/ts-lint/ts-lint.ts:37-42`

```ts
const files = collectContractFiles(root);
if (files.length === 0) {
  return { report: [`No TypeScript files found under ${root}`], fixed: [], ok: true };
}
```

End to end:

```
$ node bin/run ts-lint /tmp/emptydir
No TypeScript files found under /tmp/emptydir
No problems found
EXIT=0
```

The output is self-contradictory — it announces that it checked nothing, then that nothing is wrong — and `ok: true` makes it a pass.

This command is about to become the contract gate for three consumer repos via a container invocation with a hardcoded path. The day someone renames `spots/` → `contracts/`, or moves the tree one level deeper, or a bind-mount lands empty, or a sparse checkout omits it, every run prints those two lines and exits 0. The job is permanently green and no contract is ever linted again. Nobody reads a green job's log.

Note the asymmetry, which is what makes this a bug rather than a choice: a **nonexistent** directory makes `fs.readdirSync` throw ENOENT and correctly exits non-zero. The code already fails hard on a typo'd path and goes green on the more likely real-world variant. Untested either way, so there is not even a recorded decision.

**Fix:** return `ok: false` with a message saying nothing was checked. If some caller genuinely needs an empty tree to pass, make that an explicit `--allow-empty`, not the default.

### 2. `process.exit(1)` truncates the report whenever stdout is a pipe
**[verified — three agents]** · `cli/src/commands/ts-lint.ts:47`

`this.log()` writes to `process.stdout`, which is **asynchronous when stdout is a pipe** on POSIX. `process.exit(1)` kills the process before the buffer drains. Three identical runs on a 40-file tree producing a 1601-line report:

```
$ node bin/run ts-lint $BIG | cat > out
run1 lines=770    run2 lines=1601    run3 lines=577
```

Redirecting straight to a file is fine (sync); piping is not. Every consumer invocation in this stack's target environment — a container in CI — has stdout as a pipe.

The exit code is correct so CI still goes red, but the contributor sees an arbitrary prefix of the problems, fixes those, and the next run surfaces "new" errors that were always there. The header lies too: `Lint (1600):` prints while only 700 of the lines follow.

`lint.ts` uses the same idiom, but it logs one line before exiting; this command's entire product is a multi-hundred-line report, so the shared pattern is not safe here.

**Fix** — verified in place, 1601/1601/1601 across three piped runs with the exit code preserved:

```ts
} else {
  process.exitCode = 1;
}
```

---

## Important

### 3. `engines.node: ">=18.12.0"` is false now that the lint stack is a runtime dependency
**[verified — two agents]** · `package.json`

This PR moves `eslint`, `prettier`, `typescript-eslint`, `@eslint/js` and `eslint-config-prettier` into **`dependencies`**. From the lockfile:

As devDependencies (#2709) this never reached consumers. As runtime dependencies it does. A consumer on Node 18 — a version this package still advertises — gets `EBADENGINE` on install, or a hard failure under `engine-strict`. Meanwhile `test-node-18` still runs in the matrix.

Either bump `engines` to `>=20.19.0` and drop 18 from the matrix (itself a breaking change, so it wants a major), or keep the lint stack out of `dependencies`. **This is the one finding in the stack that breaks installs for existing external npm users**, so it wants a decision before merge rather than a follow-up.

### 4. The test guarding the design's central claim is vacuous on its ESLint half
**[verified]** · `lib/src/ts-lint/ts-lint.spec.ts:68-84`, `ts-lint.ts:130`

The body says of `overrideConfigFile: true`:

> That is the load-bearing claim of the in-code-config design, so it has a test rather than a comment.

The Prettier half holds. The ESLint half does not — running the clean fixture plus the test's own `eslint.config.mjs` with `overrideConfigFile: true` **removed** gives an identical result:

```
overrideConfigFile=true:  0 messages
overrideConfigFile=false: 0 messages
```

The leftover config sets `no-undef: "error"`, and `tseslint.configs.recommended` — inside `overrideConfig`, which wins over any discovered file — turns `no-undef` off again. So deleting `overrideConfigFile: true` leaves the suite green, which is exactly the regression the test exists to catch.

Discriminating fixture, verified 0 vs 5 messages:

```js
export default [{ rules: { 'no-restricted-syntax': ['error',
  { selector: 'TSInterfaceDeclaration', message: 'leftover config was read' }] } }];
```

### 5. Five of the eight rule disablements are dead to the fixtures
**[verified]** · `lib/src/ts-lint/eslint-config.ts:32-45`

Flipping each `"off"` to `"error"` in turn and running both fixture trees:

| rule | fixture trips it? |
|---|---|
| `no-empty-function` | yes (10 sites) |
| `explicit-function-return-type` | yes (10 sites) |
| `no-unused-vars` | yes (22 sites) |
| `no-inferrable-types` | **no** |
| `no-use-before-define` | **no** |
| `no-empty-object-type` | **no** |
| `no-wrapper-object-types` | **no** |
| `no-unsafe-function-type` | **no** |

`no-wrapper-object-types` is the sharp one. The body singles it out as the disablement that "matters, because `String` and `Boolean` are Spot's own contract vocabulary", and the code comment says the same — but turned on against the clean fixture, which does `import { String } from '@airtasker/spot'` and uses `String` in path params, headers and query params, it reports **nothing**. The import shadows the global, so the rule may never see a wrapper type at all.

Per the repo's own rule on invariants named in comments: this needs either a fixture that isolates it, or a corrected comment.

Separately, deleting `js.configs.recommended` and `eslint-config-prettier` outright also leaves all six tests green. The entire ESLint surface the suite exercises is **one rule at one line** (`naming-convention` at `dirty/models.ts:27`), against a config shipping ~85 rules. `eslint-config-prettier` is the more interesting of the two — under ESLint 10 + typescript-eslint 8 the formatting rules it disables are largely already gone, so it may be a genuine no-op worth confirming rather than shipping on faith.

### 6. A syntax error in one file discards every other finding, and under `--fix` leaves a partially rewritten tree
**[verified]** · `lib/src/ts-lint/ts-lint.ts:95`

`format()` from `prettier/standalone` **throws** on unparseable input. It is called after `runEslint` has completed and before `runTypeCheck` runs, with no catch. So a single contract with a syntax error means: the fully-computed ESLint report for the whole tree is thrown away, the type check never runs, and `cli/src/commands/ts-lint.ts` has no catch either, so oclif prints a raw stack.

The error does not name the file — no `filepath` is passed in `prettierConfig` — so on a 345-file tree you get `SyntaxError: ',' expected. (12:7)` and bisect by hand.

Under `--fix` it is worse: files already written stay written, and `fixed` never reaches the caller, so there is no record of what was modified. Verified with a tree of `a-ok.ts` (valid, unformatted) and `z-broken.ts` (syntax error) — `a-ok.ts` was rewritten on disk and the call threw.

This directly defeats the module docstring's headline invariant ("every step runs even when an earlier one has already found something"). ESLint handles the same input gracefully — a parse error becomes a `fatal` message — so only the Prettier step has this hole. Catch per file, attribute it, keep going.

### 7. The type step is not bounded by the walked file list
**[verified]** · `ts-lint.ts:61,75` vs `:171`

`collectContractFiles` documents skipping `node_modules`; `runTypeCheck` re-imports it via `resolveSourceFileDependencies()`, and `contractCompilerOptions` (from #2708) does not set `skipLibCheck`. A `node_modules/junk/index.d.ts` with a bad declaration produces:

```
Types (2): …/node_modules/junk/index.d.ts:2 - Type 'number' is not assignable to type 'string'
```

…and `ok: false`. Same for any file imported from outside the walked directory. A consumer fails `spot ts-lint` on a file they did not ask to check, cannot edit, and that `--fix` cannot touch. The `files` array does not bound the type step, and the comment claims it does.

### 8. The documented ESLint-before-Prettier ordering is not load-bearing in any test
**[verified]** · `ts-lint.ts:43-46`

The comment states the invariant explicitly — "an ESLint fix produces code, not layout … doing it the other way round leaves a tree that `--fix` claims to have fixed and that fails on the next run". Swapping the two calls leaves **all six tests passing**.

The reason is measurable: neither fixture contains a single auto-fixable ESLint error. `ESLint({fix:true}).lintFiles()` returns `output === undefined` for every file in both trees, so `ESLint.outputFixes` and the `fixed.add(...)` branch never execute against a non-empty result — `fixed` is populated only by Prettier.

Fix: put one auto-fixable lint error into `dirty/` whose fix produces code needing re-layout (a `prefer-const` or `no-extra-boolean-cast` site on a line long enough that Prettier must rewrap). Assert that one `--fix` run leaves the tree clean on the *next* check.

### 9. Under `--fix`, lint `line:col` point at pre-Prettier text
**[verified]** · `ts-lint.ts:49-51`

ESLint runs first and its messages are computed against the ESLint-fixed source; Prettier then rewrites the same files and shifts the lines; the stale positions are what get printed.

```
  .../m.ts:5:14 error Variable name `BAD_name` must match ... (naming-convention)
```

Reported at line 5, actually at line 3. Editor and CI error parsers jump to the wrong place, and offsets compound across a large tree. The type step is unaffected (it reads from disk after Prettier writes). Cheapest fix: when `fix` is true, re-run ESLint report-only after `runPrettier` has written.

### 10. No test of the CLI exit code
`cli/src/commands/ts-lint.ts:37-49`

There is no spec file for `ts-lint` under `cli/`. `process.exit(1)` on findings and implicit 0 on clean is the entire contract CI consumes, and it is verified only by hand in the PR description. Confirmed working today by hand — which is the problem: a regression is invisible to `pnpm test` and surfaces only as a consumer repo's CI going permanently green.

Spawn `bin/run ts-lint` against both fixtures as a child process; assert `status === 0` for `clean`, `1` for `dirty`, and the three category headers in stdout.

---

## Suggestions

- **The warning path is unreachable and self-contradictory if reached.** `ts-lint.ts:146,155-159`. No rule in the bundled config is set to `"warn"` (`js.configs.recommended` and `tseslint.configs.recommended` are all errors), so the `severity === 1` branch never runs. If it ever did, the CLI would print `Lint (2): … warning …` immediately followed by `No problems found` and exit 0, because `ok` counts errors only while the header counts both. Either add a `warn` rule with a test, or drop the branch.

- **ESLint-ignored files report as clean.** With `overrideConfigFile: true`, a file passed explicitly that matches a default ignore yields a `File ignored…` warning with `errorCount: 0`. The file was *not linted*, and the command says no problems were found — the same "not examined rendered as clean" class as finding 1.

- **`--fix` in CI is unconditionally green.** Under `fix: true`, `runPrettier` writes and always returns `ok: true`. A CI job that copy-pastes the `--fix` example from the command's own `static examples` rewrites files in an ephemeral workspace, discards them, and exits 0 forever. Worth documenting `--fix` as local-only, or exiting non-zero when it changed anything.

- **`fixed: Set<string>` is a mutable out-parameter** threaded into two helpers that both write to it and neither returns it. `TsLintOutcome.fixed` can be non-empty with `fix: false` — an invalid state only convention prevents. Returning `fixed` from each step and unioning at the top makes it unrepresentable.

- **Dot-directories are skipped silently.** `ts-lint.ts:72` — `entry.name.startsWith(".")` drops a contract in `.generated/`, contradicting the stated design principle that a file "must not pass by being invisible". Symlinked subtrees are skipped too (`entry.isDirectory()` is false for a symlink), which is routine in a pnpm workspace.

- **`test-fixtures/ts-lint/clean/` duplicates `test-fixtures/contract/` byte-for-byte** modulo quote style. The copy is defensible (the canonical config is single-quote, this repo's is not) but nothing records the relationship, so extending `test-fixtures/contract/` with a new decorator shape will leave the ts-lint copy stale and `clean/` will quietly stop representing a current contract. One comment naming its twin is enough.

- **`ts-lint.ts:163-164` docstring claims more than the code checks.** "a tree that passes here is a tree `generate` and `validate` can read" — `parse` runs `validateProject` *and then* `parseContract`, which returns its own errors (a missing `@api`, an unsupported type, a malformed decorator). A tree can be type-clean and still unreadable. Narrow it to the type-check guarantee.

- **`CLAUDE.md` "CLI Usage" gains no `ts-lint` entry**, though this is the most consequential new public command in the stack. (The README's command list is regenerated by `oclif-dev readme` at `prepack`, so that one is fine.) The list is already partial, so housekeeping rather than a rule breach.

- **Configs are exported mutable module singletons.** `prettierConfig.singleQuote = false` and `eslintConfig.push(...)` both succeed, shared across every `tsLint()` call in the process and reachable by deep import (`files` ships `build/`, no `exports` map). `Object.freeze` / `as const satisfies Options` would express what the docstrings claim.

- **`as Linter.Config[]` is an unchecked assertion over a real incompatibility.** Removing it gives `TS2322: Type 'ConfigArray' is not assignable…`. And it buys nothing where it counts: `Linter.Config["rules"]` is an open index signature, so a typo'd rule name (`no-consoul`, `@typescript-eslint/no-inferrable-typos`) compiles *and* runs clean at severity `"off"`. When typescript-eslint v9 renames `no-wrapper-object-types`, the `off` silently becomes a no-op — the exact outcome the comment exists to prevent.

---

## Verified clean — the biggest risks in this PR are not present

- **The collateral reformat is provably behaviour-preserving.** `diff -rq build/` against the parent branch is **empty except for the new ts-lint files**. Every hunk in `openapi2-specification.ts`, `openapi3-specification.ts`, `lib/src/types.ts` and `validate.ts` is an `interface X extends A, B {}` heritage-clause or type-alias re-wrap; both generator files contain zero top-level runtime declarations. This is the strongest available evidence for a 200-line sweep across public API.
- **`lib/src/types.ts` is not a breaking change.** The single hunk re-wraps `LiteralType`'s union onto one line — same four members, same order, no widening or narrowing. Leading-pipe and inline union syntax are the identical type; the emitted `.d.ts` differs only in whitespace. **No major version is warranted by type changes** (finding 3 is a packaging break, not a type break).
- **The #2710 trap does not recur.** Every runtime `require` in the built `ts-lint` (`@eslint/js`, `eslint`, `eslint-config-prettier/flat`, `prettier/standalone`, `prettier/plugins/*`, `typescript-eslint`, `@oclif/command`) is a declared `dependency` with satisfied peer ranges.
- **The reformat is genuine Prettier 3.9.6 output.** `prettier --list-different` over the whole repo returns nothing, so the 82-character heritage clauses are legitimate (Prettier may exceed `printWidth` where there is no valid break point).
- **`.prettierignore` / `eslint.config.js` only add `test-fixtures/ts-lint/`** — new source under `lib/src/ts-lint/` and `cli/src/commands/` is still formatted and linted. `tsconfig.json` `include` does not cover `test-fixtures/`, so the deliberately-broken `dirty/` tree cannot break `pnpm build`.
- **No exit-code swallowing.** `ok: false` does reach `process.exit(1)`. There is no `catch {}`, no `catch (e) { /* ignore */ }`, and no console-warn-only catch anywhere in the diff. The permanently-green risk comes in through findings 1 and 6, not a swallowed exception.
- No `COMPASS-17`, `TODO`, `FIXME`, `PARITY GAP`, `deferred` or `not yet ported` anywhere in the diff. No provenance or change-commentary comments — the consumer-reconciliation story stayed in the PR description where it belongs.

## Note on CI

The body says `build-and-test` will not report here. **It has** — `test-node-{18,20,22,24}` and `lint-check` are all green on this PR.

## Strengths

- **Three claims most likely to be asserted-and-never-tested each have an isolating test instead**: config-beside-the-contract is ignored (`ts-lint.spec.ts:70`), an unimported file is still checked (`:86`), and a non-`--fix` run does not write (`:40`). Finding 4 is a gap in one half of the first of those, not an absence of the instinct.
- **The Prettier config is genuinely load-bearing.** Mutating each option against the clean fixture: `singleQuote`, `tabWidth`, `semi`, `trailingComma`, `printWidth` (both 120 and 40) and `bracketSpacing` are all detected. Only `arrowParens` is not, and the config does not set it. The `.prettierrc`-beside-the-contract test writes `singleQuote: false, tabWidth: 8`, both of which would change the clean fixture's output if they took effect — so that test is real.
- **The `naming-convention` property exemption is load-bearing and tested**: removing `{ selector: "property", format: null }` produces three findings on the clean tree (`'x-auth-token'` and the hyphenated query names).
- Running all three steps unconditionally, and ordering ESLint's fixes before Prettier's write, are both correct decisions with the reasoning recorded at the call site.
- `copyFixture` correctly isolates the mutating tests into a tmpdir, so the `--fix` test cannot corrupt the checked-in fixture — a common failure in this shape of suite.
- The type step is genuinely guarded end-to-end: the clean fixture imports from `@airtasker/spot`, so a regression in #2708's path mapping surfaces here as `Cannot find module` and fails the clean test.
- `eslint-config.ts` / `prettier-config.ts` docstrings are the model the rest of the codebase should follow — they state exactly what the mechanism does (no config discovery, no plugin resolution) and the spec isolates it.

timdawborn:
# Second-pass review — #2711

Multi-agent re-review of both commits (general review, tests, error handling, type design, comments), with the follow-up `4a6111b` reviewed adversarially. Four agents worked this PR; every finding was reproduced by building the branch and running the CLI against constructed trees.

**Verdict: the follow-up fixed four real bugs, and introduced two Criticals of the same class it set out to close.** The empty-tree fix, `process.exitCode`, and the per-file Prettier catch are all correct and verified. The `node_modules` filter and the untested rule exemptions are the problems.

---

## Critical

### 1. The `node_modules` diagnostic filter disables the entire type step for any tree whose path contains a `node_modules` segment
`lib/src/ts-lint/ts-lint.ts:216-220` — *three agents independently*

```ts
return !file.getFilePath().split(path.posix.sep).includes("node_modules");
```

The predicate tests the **whole absolute path**, not the portion below the checked directory. Demonstrated with byte-identical files:

| tree location | output | exit |
|---|---|---|
| `…/mono/node_modules/@co/contracts` | `No problems found` | **0** |
| `…/plain` (same files) | `Types (1): Type 'string' is not assignable to type 'number'.` | 1 |

Formatting and ESLint still run, so the command looks alive and the loss is invisible. **Scenario:** a consumer publishes contracts as a package and gates the installed copy — `spot ts-lint node_modules/@company/contracts` — or a monorepo mounts the tree under such a path. The type step is a permanent no-op and the pipeline is green: the same "gate nothing can notice has stopped working" failure this commit was written to close, one layer down.

**And it breaks the invariant the step is documented by.** `ts-lint.ts:194-197` promises "under the same compiler options `parse` uses, so a tree that clears this step also clears the type-check `parse` performs". `parse()` → `validateProject` throws on **all** pre-emit diagnostics, unfiltered, from the same options object:

```
$ node bin/run ts-lint $S/valjunk   → No problems found                       (exit 0)
$ node bin/run validate .../api.ts  → Error: Type 'string' is not assignable…  (exit 2)
```

So the gate that exists to pre-empt `validate`/`generate` failures now passes trees those commands reject, and `validate` still hits the unfixable-`node_modules`-diagnostic problem the commit says it eliminated — because the fix landed in one of the **two** callers of the shared options object.

**Fix at the shared layer instead:** `skipLibCheck: true` in `contractCompilerOptions` (`lib/src/ts-project.ts`). That is precisely "don't check declaration files", it needs no path heuristic, it cannot produce the false negative above, it fixes `validate`/`generate` too, and it is what this repo's own `tsconfig.json` already does. If a ts-lint-local filter is kept for another reason, scope it to `path.relative(root, filePath)` and correct the docstring.

### 2. A typo in any of the nine bundled rule exemptions silently re-enables the rule, and nothing catches it
`lib/src/ts-lint/eslint-config.ts:34-48`

Rule names are `Record<string, …>` keys, so `tsc` accepts a typo; ESLint 10 then **silently ignores** the unknown key, leaving the real rule at its `tseslint.configs.recommended` severity. Verified differentially on one tree containing `export interface EmptyBody {}`:

- correct config → `ok: true`, `report: []`
- `no-empty-object-type` → `no-empty-object-typo` → `ok: false`, hard lint error

**Scenario:** a consumer has deleted its own `eslint.config.mjs` as part of this migration, so it has no way to suppress the rule. Its contract gate goes red on a file the canonical config believes it exempted, and the author has no local config to fix.

Measured coverage of the nine: flipping each `"off"` to `"error"` against the clean-tree test, **five survive unexercised** — `no-inferrable-types`, `no-use-before-define`, `no-empty-object-type`, `no-wrapper-object-types`, `no-unsafe-function-type`. Three are covered (`no-empty-function`, `explicit-function-return-type`, `no-unused-vars`).

This is the one place where correcting the `no-wrapper-object-types` comment instead of adding coverage was the wrong trade: the corrected comment now makes a precise, testable claim, and the PR body records the manual verification as prose. Convert it into the test it already is — one fixture with an inferrable annotation, a forward type reference, an empty `{}` headers object, an imported `String`/`Boolean` param and a bare `Function` type makes all five load-bearing under the existing clean-tree test. TypeScript cannot help here; a fixture is the only available guard.

---

## Important

### 3. The new catch block throws, discarding everything it was added to preserve
`lib/src/ts-lint/ts-lint.ts:117`

`catch (e)` is `unknown`; `(e as Error).message` is an unchecked assertion. Probed by stubbing `format()`:

- `throw "boom"` → `TypeError: Cannot read properties of undefined (reading 'split')`
- `throw null` → `TypeError: Cannot read properties of null (reading 'message')`

The run dies **inside the handler**, naming no file at all, and every ESLint finding already computed plus the type check are discarded — verbatim the failure this catch prevents. Prettier plugins are third-party code; a string throw is one plugin away.

```ts
const detail = e instanceof Error ? e.message : String(e);
```

### 4. "Prettier is the only step that throws" is false — three sibling throw sites are uncovered
`ts-lint.ts:113-114`, and the sites at `:105`, `:123`, `:164`

1. **`fs.readFileSync` (`:105`, immediately above the `try`)** — an unreadable file kills the whole run: `EACCES: permission denied, open …/secret.ts`, exit 1, zero findings for the rest of the tree.
2. **`ESLint.lintFiles` (`:164`, before Prettier runs)** — a dangling `*.ts` symlink: `Error: No files matching '…/dangling.ts' were found.` The run dies before any step reports.
3. **`fs.writeFileSync` (`:123`, under `--fix`)** — verified: with `models.ts` mode 444, `api.ts` **was rewritten** (md5 changed), the run died naming only `models.ts`, and **no `Fixed …` line printed**. That is word-for-word the failure the commit message claims to have fixed — "leaving files rewritten with no record of which" — surviving because the fix covered `format()` only.

Site 3 is the realistic one: a read-only bind mount, or `docker run --user` against files owned by another uid, which is the exact ownership hazard #2713 discusses. Move the read/write inside the per-file `try` with their own categories, and print `fixed` before rethrowing anything fatal.

### 5. The `overrideConfigFile` test cannot exercise its own mechanism under jest
`lib/src/ts-lint/ts-lint.spec.ts:66-88`

Removing `overrideConfigFile: true` does fail the test — but with:

```
TypeError: A dynamic import callback was invoked without --experimental-vm-modules
  at dynamicImportConfig (eslint/lib/config/config-loader.js:185:38)
```

ESLint 10 loads **every** config file via `import()`; the same error occurs with `eslint.config.cjs`. So under jest the planted config can never be read at all, the planted rule never fires, and its content is inert. The `no-undef` → `no-restricted-syntax` swap bought nothing **in this environment** — both produce the identical TypeError. (The rule swap is genuinely better outside jest: measured 0 findings for `no-undef` vs 5 for `no-restricted-syntax` against the built config. It just isn't what makes the test pass.) The `.prettierrc` half is inert by construction too — `prettier/standalone` performs no config discovery.

Either move it to a child-process test against the built CLI, where ESLint's `import()` works, or drop the rule-content pretence and assert that `ESLint` is constructed with `overrideConfigFile: true`. As written it passes for a reason unrelated to its comment.

### 6. `continue` → `break` in the new catch survives all 12 tests
`ts-lint.ts:118`

The planted file is named `zz-broken.ts` and sorts **last**, so there are no later files for `continue` to protect. The docstring's "the remaining files are still checked" is unverified. Renaming it `aa-broken.ts` and asserting `report` contains `Formatting (2):` kills the mutation — confirmed.

### 7. The `process.exitCode` restore is load-bearing and unguarded, and its failure reports all-green
`cli/src/commands/ts-lint.spec.ts:32-35`

Deleting **just** `process.exitCode = restoreExitCode`:

```
Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
exit status: 1
```

A wholly green report that exits non-zero with no failing test named anywhere. `afterEach` also restores it **after** `writeSpy.mockRestore()`, so a throw from the restore skips it. Confirmed a leaked `exitCode` makes `npx jest <single suite>` exit 1 with every test passing — masked under `-w 4`, which is why `pnpm test` would not show it. Replace the manual save/restore with a helper that returns the code rather than leaving it on the global:

```ts
async function runCommand(args: string[]): Promise<{ exitCode?: number; out: string }>
```

### 8. A passing run with findings prints no outcome line at all
`cli/src/commands/ts-lint.ts:44-49`

ESLint 10 defaults `reportUnusedDisableDirectives` to `warn`, and the bundled config never overrides `linterOptions`, so a stale `// eslint-disable-next-line` in a consumer contract reaches this:

```
Lint (1):
  …/api.ts:8:3 warning Unused eslint-disable directive (…)
exit=0
```

Stdout carries a `Lint (N):` header — **the same shape a failing run prints** — and nothing else. A script grepping for `No problems found` now reports failure on a passing run, which is a regression from the contradictory-but-present behaviour it replaced. Print a terminating outcome line unconditionally (`Passed with 1 warning` / `Failed: 2 errors`).

Related: **the guard has no test.** Replacing `if (ok && report.length === 0)` with `if (ok)` leaves all three new CLI tests green — the identical mutation-coverage defect #2712's follow-up fixed for the TTY guard, one PR earlier in the same stack. Measured: **0 rules at severity 1, 65 at severity 2**, so the warning path is currently unreachable through the config and reachable only through that ESLint default.

### 9. A symlinked contract subdirectory is silently skipped
`ts-lint.ts:84-85`

`entry.isDirectory()` is false for a symlink to a directory, and the `else` branch only picks up `.ts` files, so the subtree is neither walked nor mentioned:

```
$S/mix/good.ts          (clean)
$S/mix/nested -> …/real (holds a type error)
$ node bin/run ts-lint $S/mix  →  No problems found   exit=0
```

Same failure class as finding 1, and `spots -> ../shared/spots` is an ordinary monorepo shape for the three consumer repos. Use `fs.statSync(full).isDirectory()` with a visited-realpath set, or `readdirSync(root, { recursive: true })`; at minimum name skipped symlinks.

### 10. The ESLint-before-Prettier ordering invariant still has no test
`ts-lint.ts:56-58`

Inverting the two calls leaves the whole suite green (12/12). `dirty`'s only lint error is `naming-convention`, which is not auto-fixable, so no test exercises an ESLint fix at all — `ESLint({fix:true}).lintFiles()` returns `output === undefined` for every file in both fixtures. Add a fixture line with an auto-fixable violation whose fix produces bad layout, then assert a following non-`--fix` run reports no `Formatting` section.

---

## Suggestions

- **One syntax error is reported three times** — `Unparseable (1)`, ESLint `Parsing error`, and `Types`, all naming the same file. This PR drops `eslint-plugin-prettier` explicitly for "same gate, one report each"; the new section breaks that rule. Keep the catch, drop the section, and let the two steps that already name it do the work.
- **`{ report: [], ok: false }` is representable** and would print nothing while exiting 1. Unreachable today only because four independent return sites each happen to push a line first. The CLI guards `ok && empty` and leaves `!ok && empty` unguarded.
- **`--fix` reports stale `line:col`** — reproduced: a `BAD_name` violation reported at `m.ts:2:7` while post-fix it is on line 12. The existing fixture hides it because Prettier happens not to shift `models.ts:27`.
- **`fix: true` with an unparseable file is untested** — precisely the bug the commit message claims to fix.
- **`expect(outcome.fixed).toEqual([])`** is missing from the no-fix test; `fix: false ⟹ fixed === []` is enforced at two sites and asserted at neither.
- **`report` elements can contain embedded newlines** (the `no-empty-object-type` message arrives as one element with three `\n`s), so `report.length` is neither a line nor a finding count — the doc comment overclaims.
- **`prettier` floats on `^3.9.6`** while the whole premise is one canonical formatting shared by three repos, and the body documents that a *minor* bump reformatted files. `ts-morph` is pinned exactly in the same block for the same reason. Pin it.
- **`collectContractFiles` on a nonexistent directory** throws a raw `ENOENT … scandir`. The comments lean on this ("already fails, from readdir") but nothing tests it and the message is a bare stack.
- **Comment duplication:** the empty-tree rationale appears three times (`ts-lint.ts:39-42`, `ts-lint.spec.ts:91-94`, `cli/…/ts-lint.spec.ts:57-60`) — keep the one in the code. `ts-lint.ts:185-187` describes code in another file. The `node_modules` rationale is stated twice in code and again in the commit body. `ts-lint.spec.ts:77` says the fixture "has four of" — it has **five**, and the commit message says 5.
- **README's oclif-generated `# Commands` section** documents USAGE for every command and still omits `ts-lint` at the top of the stack. Regenerate it here.
- **`as Linter.Config[]`** is defensible and narrower than it looks — `tseslint.config()` still type-checks its arguments (a `rules` → `rulez` typo is caught), and the assertion only papers over a real nominal incompatibility between `@typescript-eslint/utils`' `FlatConfig.LanguageOptions` and `@eslint/core`'s. Keep it, add a line naming what it hides so nobody "fixes" it.

---

## Verified sound

- **`process.exitCode` reaches the shell and nothing hangs.** `> file` gives exit 1; three piped runs of a 702-line report printed 702 lines each; piping into `head -3` and `grep -q` still exits 1 with no EPIPE stack. 1.7s for `dirty`. No ts-morph/ESLint/Prettier handle keeps the loop alive.
- **Empty tree → `ok: false` breaks no caller.** #2713's parity script runs only the two committed fixtures, both non-empty. No in-repo or downstream caller points at a possibly-empty tree.
- **The `node_modules` type test is not vacuous** — with the filter stubbed to `return true` it fails with the two planted `junk/index.d.ts` diagnostics, and deleting the `uses-junk.ts` write makes it pass, confirming the earlier vacuity was real and is fixed.
- **Unparseable under `--fix` is safe** — the broken file is byte-identical after the run, the other two are fixed, exit 1, and a second run reports identically.
- **`path.posix.sep` is the right separator** — ts-morph normalises `getFilePath()` to `/` on Windows too. The defect in that line is the segment-anywhere match, not the separator.
- **`lib/src/types.ts` is non-breaking** — one hunk, in the first commit only: `LiteralType` reflowed from a leading-pipe union to one line. Same members, same order, purely Prettier 3.9's short-union collapse.
- **`file === undefined` is still reported** — project-level diagnostics survive the filter.
- No bare ticket keys, `TODO`, or `FIXME` anywhere in the diff.

## Type design — `TsLintOutcome`, as it now stands

| | first pass | now |
|---|---|---|
| Encapsulation | 6/10 | 6/10 |
| Invariant expression | 3/10 | 3/10 |
| Invariant usefulness | 4/10 | **8/10** |
| Enforcement | 3/10 | **4/10** |

**Not restructuring the type was the right call** — one consumer, internal, in the same PR, wanting three facts. A discriminated union would be the textbook answer and the wrong trade. Usefulness rose sharply because "a gate that passes while checking nothing" is the most valuable invariant this command can hold and it now holds it.

Enforcement barely moved, because every invariant that was convention-enforced still is: the type still permits the exact bug this commit fixed (`{report: […], ok: true}` for an empty tree), and the nine exemptions have neither type nor test protection. The cheapest structural improvement, if you want one: replace `ok: boolean` with `errors` / `warnings` / `checked` counts and derive `ok` at the call site. That removes the CLI's string inspection, makes the warning path explicit and testable, and enables exit 2 for "nothing checked" — which matters operationally, because "your contracts have problems" and "I checked nothing" go to different people.

## Strengths

- The four bugs the follow-up names were all real, and three were found by reasoning about failure modes a green suite cannot show: pipe-buffered truncation (770/1601/577 lines from three identical runs), a gate passing on an empty mount, and unfixable `node_modules` diagnostics.
- **Two claims were corrected rather than defended** — including admitting the original `no-wrapper-object-types` reasoning was backwards, and that the leftover-config test had been passing tautologically. Self-reporting a test bug is the rarest class of finding.
- The `no-wrapper-object-types` correction is the best-verified comment in the stack: the rule gates on `isReferenceToGlobalFunction`, so an imported `String` is invisible to it, and driving the parser on contracts naming global `String` and `Boolean` rejects both. Accurate — only too long at 7 lines.
- Per-file Prettier catch keeps every other step's findings, verified end-to-end under `--fix` including that the unparseable file is left untouched.
- `createContractProject` reuse means the type step cannot drift from what `parse` accepts — which is exactly why finding 1 matters.

vlj91 (approved):

PR #2712, opened 2026-08-18
fix(COMPASS-17): fail fast when generate runs without a terminal

timdawborn:
## Ticket

> **Stacked on #2711 → #2710 → #2709 → #2708.** Review those first. `build-and-test` is `on: pull_request: branches: [master]`, so it will not report here until the stack merges and GitHub retargets this PR.

## What

Two changes to `generate`, and one comment on `validation-server`.

## 1. Fail fast without a terminal

`generate` prompts for `--generator`, `--language` and `--out` when they are missing. That is fine at a shell. With no terminal — a CI step, a Gradle `exec`, `docker run` without `-it` — there is nothing for inquirer to read from, so the command either hangs until the job times out or accepts an empty answer and carries on with it.

Every call site in this ticket is about to become one of those. So:

```
$ node bin/run generate -c api.ts < /dev/null
 ›   Error: Cannot prompt for --generator (-g), --language (-l), --out (-o)
 ›   without a terminal. Pass every flag explicitly when running
 ›   non-interactively.
$ echo $?
2
```

**Every missing flag is named at once.** Reporting them one prompt at a time sends the caller round the loop three times, which is exactly what a CI turnaround is worst at. Interactive use is unchanged — the guard only fires when `process.stdin.isTTY` is falsy.

Exit code 2 distinguishes "your invocation is wrong" from the exit 1 that `generate` already uses for "this generator/language does not exist".

## 2. Report where the output actually went

`generate` now logs the absolute path it wrote:

```
Generated /repo/spots/doc/output/api.yml
```

Under a container the destination is the thing callers get wrong, and both failure modes are silent:

- An `-o` pointing outside the bind mount is written **inside the container** and the command exits 0. The file simply never appears on the host.
- `-o ~/x` tilde-expands against the **container's** home directory, not the caller's.

Neither is detectable from the exit code. Printing the resolved path is what makes them visible in the log.

The resolution moves out of `outputFile` into an exported `resolveOutputPath`, which `outputFile` then calls — so the path reported and the path written cannot drift apart. Note this changes `path.join` to `path.resolve`; for the arguments these call sites pass (`relativePath` is always a bare filename) the write target is identical, and `resolveOutputPath` is covered for the relative, absolute and `~` cases.

## 3. Mark the validation-server readiness line as interface

bff-client's Gradle `ExecFork` task blocks on the `Validation server running` prefix before it runs the contract tests. Rewording that line hangs that build until it times out, with no other signal. It now says so at the line.

> Plan note: the plan places this string in `lib/src/validation-server/server.ts`. It is actually in `cli/src/commands/validation-server.ts`.

## How this was verified

- `pnpm test` — 542 tests across 51 suites pass, 7 of them new.
- **Non-interactive guard**, through the built CLI with stdin redirected from `/dev/null`, not only through unit tests: all three flags missing names all three and exits 2; with `-g` and `-l` supplied it names only `--out`.
- **The full round trip from a directory with no `node_modules` on any parent path** — which is how the image will be invoked. `generate -c api.ts -g openapi3 -l yaml -o out` wrote `out/api.yml` and logged its absolute path; `ts-lint .` walked the tree and reported on it. Both exited as expected.
- `resolveOutputPath` has unit coverage for a relative directory, an absolute one, and `~`.

## Aside, for the docker PR

`parse()`'s type resolution is sensitive to the process working directory when the `@airtasker/spot` mapping resolves to **source** (`lib/src/lib.ts`, i.e. under ts-jest): the project then pulls in Spot's own sources, which need `@types/node` reachable from the cwd. It is not sensitive when the mapping resolves to the **built** `build/lib/src/lib.d.ts`, which is what the image ships — confirmed by the no-`node_modules` run above. Worth re-checking inside the container in the image PR rather than assuming.

### DIFF
--- a/cli/src/commands/generate.spec.ts
+++ b/cli/src/commands/generate.spec.ts
@@ -1,7 +1,12 @@
 import * as fs from "fs";
+import { prompt } from "inquirer";
 import * as os from "os";
 import * as path from "path";
 import Generate from "./generate";
+
+jest.mock("inquirer", () => ({ prompt: jest.fn() }));
+
+const promptMock = prompt as unknown as jest.Mock;
 
 const CONTRACT = path.join(__dirname, "../../../test-fixtures/contract/api.ts");
 
@@ -12,11 +17,16 @@ describe("generate", () => {
 
   afterEach(() => {
     process.stdin.isTTY = realIsTTY;
+    promptMock.mockReset();
     jest.restoreAllMocks();
   });
 
   function withoutTerminal(): void {
     process.stdin.isTTY = false;
+  }
+
+  function withTerminal(): void {
+    process.stdin.isTTY = true;
   }
 
   async function failureFrom(argv: string[]): Promise<Error> {
@@ -90,5 +100,24 @@ describe("generate", () => {
     expect(log.mock.calls.map(call => String(call[0])).join("")).toContain(
       `Generated ${written}`
     );
+  });
+  test("prompts for the missing flags when there is a terminal", async () => {
+    // The other cases all sit on the no-terminal side of the guard, so without
+    // this one the condition itself is unconstrained: making it unconditional
+    // keeps them all green while destroying every interactive invocation.
+    withTerminal();
+    const outDir = fs.realpathSync(
+      fs.mkdtempSync(path.join(os.tmpdir(), "spot-generate-tty-"))
+    );
+    promptMock
+      .mockResolvedValueOnce({ Generator: "openapi3" })
+      .mockResolvedValueOnce({ Language: "yaml" })
+      .mockResolvedValueOnce({ "Output destination": outDir });
+    jest.spyOn(process.stdout, "write").mockReturnValue(true);
+
+    await Generate.run(["-c", CONTRACT]);
+
+    expect(promptMock).toHaveBeenCalledTimes(3);
+    expect(fs.existsSync(path.join(outDir, "api.yml"))).toBe(true);
   });
 });
--- a/cli/src/commands/generate.spec.ts
+++ b/cli/src/commands/generate.spec.ts
@@ -1,0 +1,94 @@
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import Generate from "./generate";
+
+const CONTRACT = path.join(__dirname, "../../../test-fixtures/contract/api.ts");
+
+describe("generate", () => {
+  jest.setTimeout(60000);
+
+  const realIsTTY = process.stdin.isTTY;
+
+  afterEach(() => {
+    process.stdin.isTTY = realIsTTY;
+    jest.restoreAllMocks();
+  });
+
+  function withoutTerminal(): void {
+    process.stdin.isTTY = false;
+  }
+
+  async function failureFrom(argv: string[]): Promise<Error> {
+    try {
+      await Generate.run(argv);
+    } catch (e) {
+      return e as Error;
+    }
+    throw new Error(`Expected generate ${argv.join(" ")} to fail`);
+  }
+
+  test("exits 2 rather than prompting when there is no terminal", async () => {
+    withoutTerminal();
+
+    await expect(Generate.run(["-c", CONTRACT])).rejects.toMatchObject({
+      oclif: { exit: 2 }
+    });
+  });
+
+  test("names every missing flag, not just the first one", async () => {
+    withoutTerminal();
+
+    // One run should be enough for a caller to fix its invocation. Reporting
+    // only `--generator` sends them round the loop three times.
+    const error = await failureFrom(["-c", CONTRACT]);
+
+    expect(error.message).toContain("--generator (-g)");
+    expect(error.message).toContain("--language (-l)");
+    expect(error.message).toContain("--out (-o)");
+  });
+
+  test("names only the flags that are actually missing", async () => {
+    withoutTerminal();
+
+    const error = await failureFrom([
+      "-c",
+      CONTRACT,
+      "-g",
+      "openapi3",
+      "-l",
+      "yaml"
+    ]);
+
+    expect(error.message).toContain("--out (-o)");
+    expect(error.message).not.toContain("--generator");
+    expect(error.message).not.toContain("--language");
+  });
+
+  test("reports the file it wrote, not the directory it was given", async () => {
+    withoutTerminal();
+    // realpath because macOS reaches the temp directory through a symlink,
+    // and the command reports the resolved path.
+    const outDir = fs.realpathSync(
+      fs.mkdtempSync(path.join(os.tmpdir(), "spot-generate-"))
+    );
+    const log = jest.spyOn(process.stdout, "write").mockReturnValue(true);
+
+    await Generate.run([
+      "-c",
+      CONTRACT,
+      "-g",
+      "openapi3",
+      "-l",
+      "yaml",
+      "-o",
+      outDir
+    ]);
+
+    const written = path.join(outDir, "api.yml");
+    expect(fs.existsSync(written)).toBe(true);
+    expect(log.mock.calls.map(call => String(call[0])).join("")).toContain(
+      `Generated ${written}`
+    );
+  });
+});
--- a/cli/src/commands/generate.ts
+++ b/cli/src/commands/generate.ts
@@ -6,7 +6,7 @@ import { Contract } from "../../../lib/src/definitions";
 import { generateJsonSchema } from "../../../lib/src/generators/json-schema/json-schema";
 import { generateOpenAPI2 } from "../../../lib/src/generators/openapi2/openapi2";
 import { generateOpenAPI3 } from "../../../lib/src/generators/openapi3/openapi3";
-import { outputFile } from "../../../lib/src/io/output";
+import { outputFile, resolveOutputPath } from "../../../lib/src/io/output";
 import { parse } from "../../../lib/src/parser";
 
 export default class Generate extends Command {
@@ -43,6 +43,28 @@ export default class Generate extends Command {
     const { contract: contractPath } = flags;
     let { language, generator, out: outDir } = flags;
     const contractFilename = path.basename(contractPath, ".ts");
+
+    // Each missing flag below falls back to an interactive prompt. With no
+    // terminal — a CI step, a Gradle exec, a docker run without `-it` — that
+    // prompt has nothing to read from, so the command hangs or exits on a
+    // silently empty answer. Name every missing flag at once, so a caller
+    // fixes its invocation in one pass rather than one flag per run.
+    if (!process.stdin.isTTY) {
+      const missing = [
+        generator ? null : "--generator (-g)",
+        language ? null : "--language (-l)",
+        outDir ? null : "--out (-o)"
+      ].filter((flag): flag is string => flag !== null);
+
+      if (missing.length > 0) {
+        this.error(
+          `Cannot prompt for ${missing.join(
+            ", "
+          )} without a terminal. Pass every flag explicitly when running non-interactively.`,
+          { exit: 2 }
+        );
+      }
+    }
 
     if (!generator) {
       generator = (
@@ -108,11 +130,10 @@ export default class Generate extends Command {
     const transformedContract = generatorTransformer(parse(contractPath));
     const formattedContract = formatTransformer(transformedContract);
 
-    outputFile(
-      outDir,
-      `${contractFilename}.${formatExtension}`,
-      formattedContract
-    );
+    const outputName = `${contractFilename}.${formatExtension}`;
+    outputFile(outDir, outputName, formattedContract);
+
+    this.log(`Generated ${resolveOutputPath(outDir, outputName)}`);
   }
 }
 
--- a/cli/src/commands/generate.ts
+++ b/cli/src/commands/generate.ts
@@ -44,11 +44,12 @@ export default class Generate extends Command {
     let { language, generator, out: outDir } = flags;
     const contractFilename = path.basename(contractPath, ".ts");
 
-    // Each missing flag below falls back to an interactive prompt. With no
-    // terminal — a CI step, a Gradle exec, a docker run without `-it` — that
-    // prompt has nothing to read from, so the command hangs or exits on a
-    // silently empty answer. Name every missing flag at once, so a caller
-    // fixes its invocation in one pass rather than one flag per run.
+    // Each missing flag below falls back to an interactive prompt, reading
+    // stdin. With no terminal — a CI step, a Gradle exec, a docker run without
+    // `-it` — inquirer still renders and waits on a line event that never
+    // arrives, so the command hangs until whatever is running it gives up.
+    // Name every missing flag at once, so a caller fixes its invocation in one
+    // pass rather than one flag per run.
     if (!process.stdin.isTTY) {
       const missing = [
         generator ? null : "--generator (-g)",
--- a/cli/src/commands/generate.ts
+++ b/cli/src/commands/generate.ts
@@ -45,9 +45,11 @@ export default class Generate extends Command {
     const contractFilename = path.basename(contractPath, ".ts");
 
     // Each missing flag below falls back to an interactive prompt, reading
-    // stdin. With no terminal — a CI step, a Gradle exec, a docker run without
-    // `-it` — inquirer still renders and waits on a line event that never
-    // arrives, so the command hangs until whatever is running it gives up.
+    // stdin. With no terminal the answer never arrives, and how that presents
+    // depends on the stdin given: at EOF (docker run without `-i`) the prompt
+    // never settles and the process exits having generated nothing; an open but
+    // silent pipe (a CI step, a Gradle exec) hangs; a pipe carrying newlines
+    // takes the first choice and generates the wrong artifact with exit 0.
     // Name every missing flag at once, so a caller fixes its invocation in one
     // pass rather than one flag per run.
     if (!process.stdin.isTTY) {
--- a/cli/src/commands/validation-server.spec.ts
+++ b/cli/src/commands/validation-server.spec.ts
@@ -1,0 +1,11 @@
+import { READY_LINE_PREFIX } from "./validation-server";
+
+describe("validation-server", () => {
+  test("the readiness prefix downstream builds block on is unchanged", () => {
+    // Not a tautology: it is the only thing in this repository that fails when
+    // the string is reworded. Without it a rename passes review and CI here,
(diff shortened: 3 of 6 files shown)


timdawborn:
# Review — #2712 `fix(COMPASS-17): fail fast when generate runs without a terminal`

Automated multi-agent review (`/pr-review-toolkit:review-pr`): general code review, test-coverage analysis with mutation testing, and error-handling analysis. Scoped to this PR's diff against its base (#2711). Findings marked **[verified]** were reproduced by running the code.

---

## Important

### 1. The TTY predicate can be deleted and the suite stays green
**[verified by mutation — two agents independently]** · `cli/src/commands/generate.ts:52`, `cli/src/commands/generate.spec.ts`

All four `generate` tests call `withoutTerminal()` as their first line. Nothing ever sets `process.stdin.isTTY` truthy. Mutating the guard to `if (true)` — i.e. making the flag check unconditional so `generate` *always* demands `-g -l -o`:

```
✓ exits 2 rather than prompting when there is no terminal (1709 ms)
✓ names every missing flag, not just the first one (10 ms)
✓ names only the flags that are actually missing (11 ms)
✓ reports the file it wrote, not the directory it was given (2416 ms)
```

All four pass, 6.2s, no signal. That mutation destroys every interactive invocation of `generate` — the prompting the body explicitly promises is "unchanged" — and ships green. The suite as written cannot distinguish "fail fast when there is no terminal" from "fail always".

Inversion and outright deletion *are* caught, but only by a 60-second timeout: the suite goes 6s → **184s** and reports `thrown: "Exceeded timeout of 60000 ms"`. That diagnostic names no invariant, so a future engineer reads it as a flaky hang rather than a broken guard. (`jest.setTimeout(60000)` at `generate.spec.ts:9` makes it worse.)

Per the org rule — an invariant stated in a comment needs a fixture that isolates it — one test on the other side of the condition collapses this to a ~1s assertion. Verified against the real code (passes, 1.9s) and against the mutation (fails, 1.4s, meaningful message):

```ts
jest.mock("inquirer", () => ({ prompt: jest.fn() }));

test("with a terminal, missing flags are prompted for rather than rejected", async () => {
  const outDir = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "spot-tty-")));
  process.stdin.isTTY = true;
  (prompt as unknown as jest.Mock)
    .mockResolvedValueOnce({ Generator: "openapi3" })
    .mockResolvedValueOnce({ Language: "yaml" })
    .mockResolvedValueOnce({ "Output destination": outDir });

  await Generate.run(["-c", CONTRACT]);

  expect(prompt).toHaveBeenCalledTimes(3);
  expect(fs.existsSync(path.join(outDir, "api.yml"))).toBe(true);
});
```

### 2. `validation-server.ts` declares a cross-repo string contract with nothing enforcing it
`cli/src/commands/validation-server.ts:44-47`

The new comment says rewording `Validation server running` hangs bff-client's Gradle `ExecFork` until it times out, "with no other signal". Nothing in this repo pins that string. There is no spec file for `validation-server` at all — `generate.spec.ts` is the only spec under `cli/src/commands/`.

The only thing that currently pins it is `grep -q "Validation server running"` in **#2713's** `scripts/check-image-parity`. If this PR lands and #2713 slips or is reworked, a style-motivated reword passes review, passes CI here, and breaks a downstream build silently in another repo.

A sentence of the form "change X and Z breaks" is a spec for a test. Cheapest real guard:

```ts
// cli/src/commands/validation-server.ts
export const READY_LINE_PREFIX = "Validation server running";
```
```ts
test("the readiness prefix bff-client's ExecFork blocks on is unchanged", () => {
  expect(READY_LINE_PREFIX).toBe("Validation server running");
});
```

Weak in isolation, but it makes a reword impossible without editing a test that says why — which is the entire point.

---

## Suggestions

- **The README still shows the now-failing invocation.** `README.md:67` and `:216` show `npx @airtasker/spot generate --contract api.ts` with no other flags. That remains correct for interactive npm use (it prompts), but #2714 reframes the README around Docker, where it exits 2 — and #2714 does not touch those lines. `--help` (whose text generates the README's commands section) also does not say the three flags are required non-interactively.

- **The guard comment names a failure mode that was not the one observed.** `generate.ts:47-50` says "so the command hangs or exits on a silently empty answer". Verified against `inquirer@8.2.6`: it sets `skipTTYChecks: true` by default (`lib/ui/baseUI.js`), so at EOF it renders and waits on a `line` event that never arrives — a hang. The "silently empty answer" half only happens with *piped* stdin, which the guard rejects anyway. State the one verified mode.

- **The guard over-fires for piped-but-answerable stdin.** `generate.ts:52`. The true precondition is "stdin cannot supply an answer"; `!isTTY` is a superset. Verified: `inquirer@8.2.6`'s `lib/prompts/input.js` `filterInput()` returns the default on an empty line and readline emits `line` for piped input, so `printf '\n' | spot generate -c api.ts -g openapi3 -l yaml` works today and now exits 2. Niche, arguably a desirable tightening — but it is a behaviour change shipping under a `fix(...)` patch. If deliberate, say so in the flag help.

- **Consumer name as rationale.** `validation-server.ts:44-45` names bff-client's ExecFork task. That is provenance shape — it rots when that build changes. State the invariant ("machine-parsed readiness signal; downstream builds block on this exact prefix") and keep the consumer as a parenthetical example.

- **`resolveOutputPath` is now exported public API with no pinned semantic.** `lib/src/io/output.ts:16`. The change is `path.join` → `path.resolve`, which diverge only when `relativePath` is absolute or escapes via `..`. All four call sites (`init.ts:26,63,80`, `generate.ts:111`) pass bare filenames, so the body's claim holds today — but one line decides it for future callers: `expect(resolveOutputPath("/srv/contracts", "/etc/passwd")).toBe("/etc/passwd")`.

- **The exit-1 arm is unpinned.** Exit 2 is only meaningful by contrast with exit 1, and only the 2 is asserted. Probed: `-g bogus` correctly gives `{exit: 1}` / `No such generator bogus`. A scripted caller branching on the code has no regression protection on that arm.

- **`outputFile`'s `override = false` skip branch has no test**, before or after this PR — it is what stops `spot init` clobbering an existing `api.ts` (three call sites in `init.ts`). Not introduced here, but the file just gained its first spec, so this is the cheapest moment it will ever be.

- Nit: `generate.spec.ts:68` `mkdtemp`s into `os.tmpdir()` and never cleans up.

---

## Verified — the predicate is right, and that was the main risk

- **Correct stream.** `inquirer` defaults `input` to `process.stdin`, and inquirer 8 sets `skipTTYChecks: true` so it silently waits rather than throwing. Guarding on `process.stdin.isTTY` therefore tests the actual precondition, not the `process.stdout.isTTY` correlate — the classic proxy-signal trap, avoided.
- **No fallback-to-default.** The guard errors rather than falling through to the `-o` prompt at `:117-124`, which carries `default: "."`. Falling through there is the hang-becomes-silently-wrong-output shape: a CI job would have written the generated contract into the working directory and exited 0.
- **The exit really is non-zero, end to end.** `this.error(msg, { exit: 2 })` throws a `CLIError` carrying `oclif.exit = 2`; `bin/run` terminates via `.catch(require('@oclif/errors/handle'))`, which exits with that code.
- **No unguarded sibling caller.** `inquirer.prompt` is imported and called only in `generate.ts` (three sites, all downstream of the guard). `lib/src/io/output.ts` contains no prompting helper — only `outputFile` and the new pure `resolveOutputPath`. Nothing else under `cli/src/commands/` blocks on stdin. The shared-code sweep holds.
- **No false positive on fully-flagged runs.** The error fires only when `missing.length > 0`.
- **Tests do not depend on ambient state.** `isTTY` is set explicitly via `withoutTerminal()` and restored from a captured `realIsTTY` in `afterEach` — not inherited from how Jest was invoked.

## Note on CI

The body says `build-and-test` will not report here. **It has** — `test-node-{18,20,22,24}` and `lint-check` are all green on this PR.

## Strengths

- The message names every missing flag in one pass, long and short form (`--generator (-g)`), and only the ones actually missing — a caller fixes its invocation in one round trip rather than three.
- Exit 2 is distinct from the existing exit-1 validation errors, so a scripted caller can tell "you invoked me wrongly" from "your contract is invalid".
- Coupling the reported path to the written path through a single `resolveOutputPath` call, then asserting both in one test, is the correct structural fix for the drift described.
- `generate.spec.ts:68` is the only thing stopping an over-aggressive guard — it supplies all three flags *and* calls `withoutTerminal()`, asserting success plus the written file plus the log line. It is load-bearing well beyond what its name ("reports the file it wrote") suggests; worth renaming or splitting so nobody deletes the `withoutTerminal()` call as irrelevant scaffolding.
- `failureFrom` is a good DAMP helper: it throws a descriptive error when the command unexpectedly succeeds, rather than silently passing.

timdawborn:
# Second-pass review — #2712

Multi-agent re-review of both commits, with the follow-up `7bf6f85` reviewed adversarially. Three agents worked this PR; the inquirer behaviour below was measured by driving real `inquirer@8.2.6` against three stdin shapes, and the guard was mutated six ways.

**Verdict: the follow-up's central claim is exactly right, and the contamination risk it was checked for does not exist.** The mock strengthens the suite rather than weakening it. The findings are that one comment is now less accurate than the version it replaced, and one new test pins the wrong half of its contract.

---

## Important

### 1. The rewritten guard comment is falsified by the very example it leads with
`cli/src/commands/generate.ts:47-51`

> "inquirer still renders and waits on a line event that never arrives, so the command hangs until whatever is running it gives up"

The premise is right — `skipTTYChecks` defaults to `true` (`node_modules/inquirer/lib/ui/baseUI.js:64`, upstream, no pnpm patch), so inquirer does not throw `isTtyError`; it builds a readline with `terminal: true`. But "hangs" is one of three behaviours, and **not** the one that happens for `docker run` without `-i`, or for `< /dev/null` — which is the shape the PR body says was used to verify:

| stdin shape | observed (inquirer 8.2.6) |
|---|---|
| EOF — `< /dev/null`, `docker run` without `-i` | **no hang.** readline closes immediately, the prompt promise never settles, the loop empties, process gone in ~160ms having generated nothing. Exit 1 via `ERR_USE_AFTER_CLOSE` from inquirer's own `process.on("exit")` handler; on older Node with no such error, a silent exit 0. |
| open pipe, no data — Gradle `exec`, some CI runners | **hangs.** Killed at 6s, nothing written. |
| pipe carrying newlines | **resolves silently with the first choice** — `{"Generator":"a"}`, exit 0, wrong artifact generated. |

So the comment names the mode that does not apply to its own example, and the follow-up **deleted** the third mode ("or exits on a silently empty answer") — the only genuinely silent one, and the strongest justification for the guard. The commit message's reason for dropping it ("accepting a silently empty answer needs piped stdin, which the guard rejects anyway") is true but circular: the comment exists to tell the next reader what the guard prevents, and one of those things is a wrong file generated with exit 0.

The pre-follow-up wording was the more accurate of the two. Suggested replacement:

```
// With no terminal the answer never arrives, and how that presents depends on
// the stdin it was given: at EOF (docker run without `-i`) the prompt never
// settles and the process exits having generated nothing; an open, silent pipe
// (a CI step, a Gradle exec) hangs; a pipe carrying newlines takes the first
// choice and generates the wrong artifact with exit 0.
```

### 2. The readiness spec pins the constant, not the line — and deleting the emitting call leaves the suite green
`cli/src/commands/validation-server.spec.ts:9`

Arguing both sides, since the brief asked.

**For:** not a tautology. A reword fails it (verified: `"Validation server listening"` → red), and `git grep "Validation server running"` confirms the spec's own claim at the time it was written.

**Against, and stronger:** what a downstream matches is the *composed line*, and the export is named `READY_LINE_PREFIX` — the "prefix" property, that it starts the line, is exactly what the spec cannot see. **Deleting `validation-server.ts:52` outright leaves all 551 tests / 53 suites green.** So nothing verifies that the command emits the prefix, that it lands at the start of a line, or that it is emitted only after the server is accepting connections. That last property is what the new docstring asserts ("downstream builds block on this exact prefix *before they run against the server*"); it currently holds — `defer: () => new Promise(resolve => app.listen(port, resolve))` — and nothing observes it. There is no spec for the command at all.

Cheap upgrade that keeps the tripwire and adds both properties: export `readyLine(port)` returning `` `${READY_LINE_PREFIX} on port ${port}` ``, call it from `this.log`, and assert `readyLine(3001) === "Validation server running on port 3001"`. That fails on a reword, on a reorder, and on losing the prefix position — same three lines of test. Better still, mock the server module so `defer()` resolves without a socket and assert on stdout.

---

## Suggestions

- **The absolute-`relativePath` semantic is "decided" only inside a test comment.** `lib/src/io/output.spec.ts:25-33` against `output.ts:5-17`. The parameter is named `relativePath`; the docstring promises only "where `outputFile` writes, as an absolute path". Neither says an absolute second argument discards `outDir`. The spec asserts exactly that and calls it "the decided behaviour" — but no decision was made; `path.resolve` was already there.

  It also locks in that a parameter named `relativePath` may escape `outDir`, which cuts against this PR's own motivation ("an `-o` outside the bind mount is written inside the container and exits 0"). Current call sites are safe — verified rather than taken on trust: `generate.ts` passes `path.basename(contractPath, ".ts")` (cannot be absolute) and all three `init.ts` calls pass literals. Failure scenario for a future caller: any `outputFile(outDir, nameDerivedFromInput, …)` writes outside `outDir` on an absolute value, and the new `Generated <path>` log reports the escape as a success.

  Pick one and put it at the site: rename to `outputPath` and state the precedence in the docstring, or make the function honour the name (`path.resolve(dir, path.join("/", relativePath).slice(1))`) and assert containment instead of escape. Minor: the literals are POSIX-only.

- **"The only thing in this repository" is no longer true.** `validation-server.spec.ts:5-8`. At the stack tip, `scripts/check-image-parity` matches the literal in its own `case` pattern and `README.md` prints it in a fenced block — three copies, and the comment claims one. Delete the sentence (the test name carries it), and better, have the parity script assert against `READY_LINE_PREFIX` so the claim becomes true again.

- **The guard runs before generator validation, so a bad `-g` still costs two runs.** `generate.ts:53`, ahead of the `availableGenerators().includes` check. `spot generate -c api.ts -g openapi4 -o out` in CI reports only `--language (-l)`; the caller adds it, reruns, and *then* learns the generator does not exist. Two round trips — the exact cost the guard and the "One run should be enough" comment exist to avoid. Validating a *supplied* `-g` before the guard fixes it without weakening the name-them-all property.

- **A scripted piped-answer invocation now exits 2**, shipping as a `fix:` patch. `printf 'openapi3\nyaml\nout\n' | spot generate -c api.ts` works today and returns 2 after this. Honest caveat: piped answers silently take the *first* choice regardless of what you pipe, so nobody should depend on it — but a patch release of a public package is where a consumer will not look for it. One line in the body or release note.

- **`generate.spec.ts:41` is named "exits 2 rather than prompting" and never checks the prompting half.** Now that inquirer is mocked that is free: `expect(promptMock).not.toHaveBeenCalled()`.

- **Two temp-dir tests leak directories** in `os.tmpdir()`, and `jest.setTimeout(60000)` is now sized for a hang the mock has made impossible — though `parse()` alone takes ~1.4s, so some headroom over the 5s default is still warranted.

- **Comment deletions worth making:** `lib/src/io/output.spec.ts:26-29` is four lines explaining why a test exists; the test name already says it.

---

## Corrections to the commit message

- **"fails in 3ms with a real assertion"** — right about the timing and the diagnostic, wrong that an assertion evaluates. Under `if (true)` the new test dies on an uncaught rejection from `await Generate.run` carrying the guard's own message; it never reaches either `expect`. Its two `expect`s are also near-tautological: with exactly three `mockResolvedValueOnce` queued, a 2nd or 4th prompt throws a `TypeError` before the count is checked. The test earns its place regardless.
- The remaining `{ exit: 2 }` on `this.error` is `@oclif/errors`' own default, so that option is redundant-but-documentary — dropping it fails no test. Worth knowing only if you thought that assertion pinned the deliberate choice of 2 over 1.

---

## Verified sound — the contamination risk does not exist

The module-wide `jest.mock("inquirer")` at `generate.spec.ts:7` does **not** weaken the four pre-existing tests. It strengthens them:

| Mutation of `generate.ts:53` | Result | Wall clock |
|---|---|---|
| baseline | 5 pass | 4.4s |
| `if (true)` | only the new test fails | **4ms** |
| `if (process.stdin.isTTY)` (inverted) | tests 1, 2, 3, 5 fail | **3.6s** (was a 60s timeout) |
| guard deleted | tests 1, 2, 3 fail | 5.7s |
| drop `generator ? null :` | 2 fail | — |
| drop the `--out` entry | 2 fail | — |

`mockReset()` in `afterEach` is sufficient and the suite is order-independent: after a reset the mock returns `undefined`, so any test reaching a prompt dies on `(undefined).Generator` — a `TypeError` carrying none of the flag names and no `oclif.exit`. The previously-hanging failure became a *faster, better-diagnosed* failure, not a passing one. One caveat: tests 2 and 3 now fail via that downstream `TypeError` rather than their own assertion; only the new test 5 fails with a message naming the guard.

Also verified:

- **Exit path is non-zero end to end**, through the built CLI with `< /dev/null`, and survives a pipe.
- **`READY_LINE_PREFIX` renders byte-identically** — hexdumped the real server's stdout: `Validation server running on port 5699\n`, identical to the pre-refactor literal.
- **Nothing in the follow-up catches, defaults, or discards** — it is one test, one exported constant, two rewritten comments.
- **`path.join` → `path.resolve` is safe at every current call site**, checked rather than assumed.
- **The provenance sweep is clean**: `bff`, `gradle`, `execfork`, `first consumer`, `ported`, `legacy` return nothing across `cli/src` and `lib/src`. No bare ticket keys, no `TODO`/`FIXME`.

## Pre-existing, worth a line while the file is open

`lib/src/validation-server/server.ts:65` — `new Promise(resolve => app.listen(port, resolve))`. The listen callback fires only on success; `EADDRINUSE` arrives as an `'error'` event with no listener, so the `try`/`catch` in the command never sees it. Verified: a raw `Emitted 'error' event on Server instance` stack dump, exit 1 — loud and non-zero, so not a silent failure, but it never reaches `this.error` and the readiness line simply never appears. `.on("error", reject)` fixes it.

## Strengths

- **The central claim is true and reproducible**: `if (true)` leaves the four original tests green and fails only the new TTY case. That is a real mutation-killing test, not a coverage line — and it converts a 60s-timeout diagnosis into a 3.6s one.
- Extracting `resolveOutputPath` so the reported path and the written path come from one call is the right shape for the failure mode the PR is about; the two cannot drift.
- The `path.join` → `path.resolve` risk assessment in the body is accurate and holds against the tree.

vlj91 (approved):

PR #2713, opened 2026-08-18
build(COMPASS-17): self-contained docker image and parity CI gate

timdawborn:
## Ticket

> **Stacked on #2712 → #2711 → #2710 → #2709 → #2708.** Review those first. `build-and-test` does report on this PR — `docker-parity`, `test-node-{18,20,22,24}` and `lint-check` are all green, so the gate this PR adds has run.

## What

A `Dockerfile`, an allowlist `.dockerignore`, `scripts/check-image-parity`, and a `docker-parity` CI job wired into `publish`.

```sh
docker run --rm --user "$(id -u):$(id -g)" \
  --volume "$PWD:$PWD" --workdir "$PWD" \
  ghcr.io/airtasker/spot:<version> \
  generate -c api.ts -g openapi3 -l yaml -o doc/output
```

## The image

Four stages, following the shape `openapi-generators` already uses for the downstream half of the same pipelines: a base carrying the node and pnpm pins, a **prod-deps stage resolved on its own** so the compiler and test toolchain never reach the runtime, a builder, and a runtime.

- **`pnpm exec tsc`, not the pack lifecycle.** `pnpm build` *is* `tsc`, so it would do the same thing — the script to stay off is `prepack`, which also runs `oclif-dev manifest` (the image must ship none) and `pnpm build-docs`, a webpack/redoc bundle needing `--openssl-legacy-provider` that this image never serves.
- **`docs/package.json` is copied into the base stage.** It is a workspace member, and `--frozen-lockfile` refuses to install without it. Its own dependencies are all `devDependencies`, so no webpack or redoc reaches prod-deps.
- **`package.json`, `bin/` and `build/` keep their relative positions** in the runtime stage. oclif discovers commands by scanning `oclif.commands` relative to the nearest `package.json`.
- **98 MB.** The plan estimated 250–350 MB.
- **`mock`, `docs` and `init` are removed.** Two serve long-running local dev servers and one scaffolds a project on the host; none belongs in a container invoked per-command. Deleting their compiled files removes the commands outright, because no `oclif.manifest.json` is shipped that could still advertise them. `docker run <image> --help` lists exactly the six supported commands.

`.dockerignore` is an allowlist — exclude everything, then re-include what the Dockerfile `COPY`s — so a new top-level directory cannot silently inflate the build context.

## Two things the gate could not catch, now fixed

**The expected status of each comparison.** `compare_run` asserted only that the CLI and the image agreed. It does catch the `HOME` regression — that makes the image exit 0 while the CLI exits 2 — but it could not catch the underlying failure ceasing to be a failure. If `contract-with-type-error.ts` were edited, or `validate` stopped type-checking, both sides would exit 0, the comparison would pass, and the case built specifically to exercise oclif's error path would become a seventh always-succeeding check. That fixture is shared with `parser.spec.ts` in #2708, so the edit is plausible. Each case now names the status it must produce; inverting one fails the gate with a message naming the case.

**arm64 had never been built.** `docker-parity` builds single-arch, because the script has to *run* the image and a multi-arch result cannot be loaded. The publish job in the next PR builds `linux/amd64,linux/arm64` under QEMU, so a release would have been the first time arm64 was built at all — and an emulated `pnpm install --prod` that fails where amd64 resolved cleanly would surface on Apple Silicon and bff-client's arm64 runner, not here. The job now builds every published architecture without loading it: that covers the build, not the behaviour. Executing arm64 is worth a follow-up.

## Image defaults contradicted the documented invocation

`WORKDIR /workspace` was created root-owned `0755` and no `USER` was set, so both forms failed in opposite directions: `docker run --user …` with the default workdir could not write to it, and a bare `docker run` with no `--user` produced exactly the root-owned output the header calls out as breaking downstream builds. Now `chmod 1777` and `USER node`, both still overridden by `--user`. Verified: `id` in the image reports uid 1000, and a `--user` run writes into `/workspace`.

## The bug the parity gate found

**`ENV HOME=/tmp` is load-bearing, and this is the finding I'd most like a second pair of eyes on.**

The invocation convention requires `--user "$(id -u):$(id -g)"` — not for tidiness, but because rails-monolith's RSpec boot caches on the mtime of Spot's output and bff-client's Gradle codegen treats it as a task input, so root-owned output breaks both. That uid has no `/etc/passwd` entry, so `os.homedir()` returns `/`. oclif writes an error log under the home directory before it exits, that `mkdir` fails with `EACCES`, and **the throw inside the error handler pre-empts the exit code**.

The result:

```
$ docker run ... spot:parity validate <a contract with a type error>
    Error: Type 'string' is not assignable to type 'number'.
[Error: EACCES: permission denied, mkdir '/.cache']
$ echo $?
0
```

Every failing command exited 0. `spot validate` would have reported an invalid contract and still passed rails-monolith's `spot-contract-validation` required gate, and every `validate-contract` step in every consumer. Silently, with the error text right there in the log.

With `HOME=/tmp` it exits 2, matching the host CLI exactly.

**Why the first version of the parity script missed it, and what changed:** the script compared successful commands plus `ts-lint dirty`. `ts-lint` fails through `process.exit(1)` directly, so it was immune; `validate`, `lint` and `generate` fail through oclif's error handler, which was broken. A gate built only from those cases passes against an image where nothing can fail. There is now an explicit comparison of a command that fails **through oclif's own error path**, and it is commented as to why.

## `scripts/check-image-parity`

| Check | What it covers |
|---|---|
| `generate` × 7 generator/language combinations | output compared **byte for byte** via `diff --recursive` |
| `validate`, `lint`, `lint --no-trailing-forward-slash=error`, `ts-lint clean`, `ts-lint dirty`, `validate` on an invalid contract | **the exit status each case must produce**, then exit code and stdout identical between the two |
| self-containment | runs from a `mktemp -d` with **no `node_modules` on any parent path**, asserting that first so the check cannot pass vacuously |
| `validation-server` | `docker run -d --publish`, poll `docker logs` for the readiness line (so: no TTY), `curl /health`, idempotent teardown |
| `mock`, `docs`, `init` | non-zero exit **and** a "not found" message — not merely non-zero, which any failure would satisfy |

The generator list is parsed out of `cli/src/commands/generate.ts` and checked against the script's matrix. A generator added there fails this gate with a message naming it, rather than going quietly uncompared — and if the parse yields nothing, the script exits rather than comparing zero cases.

`lint --no-trailing-forward-slash=error` is in there specifically because bff-client passes it.

## How this was verified

- **`./scripts/check-image-parity` passes end to end** against a locally built image — 7 generator/language combinations identical, 6 exit-code-and-output comparisons identical, self-containment, validation-server readiness + port + teardown, and the three absent commands.
- **`docker buildx build --platform linux/amd64,linux/arm64` succeeds.** arm64 is not optional: engineers run this on Apple Silicon, and bff-client's devcontainer CI runner is arm64.
- `pnpm test` — 542 tests across 51 suites pass. `pnpm lint:check` clean.

## Other changes

- **`.tool-versions`**: `node >=18.12.0` → `nodejs 22.22.1`. The old value was not a valid version spec and used a key (`node`) that no other repo in the org uses — `openapi-generators`, `bff-client` and `enrichment-engine` all use `nodejs`. It now matches the image base tag, and `AGENTS.md` records that the base tag, this line and the corepack pin move together.
- **`AGENTS.md`** (`CLAUDE.md` symlinks to it): a Docker section, `ts-lint` in the CLI list, `ts-lint/` and `ts-project.ts` in the structure list, and the oclif command-scan note that explains why deleting a file removes a command. Also corrects `parser/` to `parsers/`, which has never existed under that name.

## Not in this PR

Publishing. The `docker-publish` job, the ghcr private-flip checklist and the docker-first README are the next PR, gated on the same GitHub Release event that already publishes to npm.


timdawborn:
# Review — #2713 `build(COMPASS-17): self-contained docker image and parity CI gate`

Automated multi-agent review (`/pr-review-toolkit:review-pr`): general code review, error-handling/shell-safety analysis, and comment-accuracy analysis. Scoped to this PR's diff against its base (#2712). Findings marked **[verified]** were checked against the tree or by running the code.

The `HOME=/tmp` discovery, and the honesty about why the *first* version of the parity gate missed it, is the best thing in this stack. Most of what follows is about making the gate durable and the docs match the tree.

---

## Important

### 1. `compare_run` never asserts an expected exit status
**[verified — two agents]** · `scripts/check-image-parity:117-139`, used at `:148-153`

`compare_run` compares `cli_status` to `image_status` and the two outputs, then prints `OK: <label> (exit ${cli_status}, identical output)`. It never asserts that the cases chosen *because they must fail* actually failed.

To be precise about what this does and does not cover: it **does** catch the `HOME` regression today — that makes the image exit 0 while the CLI exits 2, so the statuses differ and the gate fails. The weakness is durability. If the failure stops being a failure on *both* sides — someone edits `lib/src/__spec-examples__/contract-with-type-error.ts`, which is **shared with #2708's `parser.spec.ts`**, or `validate` stops type-checking the project — then both sides exit 0, the comparison passes, and the case silently degrades into a seventh always-succeeding comparison while its comment still claims it covers oclif's error path.

The fixture's own trailing comment already anticipates this ("a parse that succeeds here has stopped type-checking the project altogether") — that is a spec for an assertion the script does not make.

**Fix:** take an expected status, e.g. `compare_run "validate an invalid contract" 1 validate …`, and inside:

```bash
if [ -n "${expected}" ] && [ "${cli_status}" -ne "${expected}" ]; then
  fail "${label}: expected exit ${expected}, both sides exited ${cli_status}"
  return
fi
```

### 2. `AGENTS.md:38` states a behaviour the tree does not have
**[verified]**

> Both distributions are published from one GitHub Release event, so a version means the same thing on npm and on ghcr.

At this commit there is no ghcr publish. `publish` publishes to npmjs.org and npm.pkg.github.com only; it gains `needs: docker-parity` but no image push, and `ghcr` appears **nowhere** in the workflow (confirmed: `grep -c ghcr` on this branch's `build-and-test.yml` returns 0). The publish arrives in #2714.

This is a future-tense claim written in the present tense in a committed context doc. The next reader — human or agent — concludes that pulling `ghcr.io/airtasker/spot:2.0.3` is possible and version-aligned. Two more sites in the same PR have the same shape:

- the `docker-parity` build step comment: *"Multi-arch cannot be loaded, and is covered on publish."* Nothing covers it on publish at this commit — and see finding 3.
- `Dockerfile:8-10` runs `ghcr.io/airtasker/spot:<version>` in its usage example with no note that the image is unpublished and (per #2714) will be private.

Move these into #2714 alongside the code that makes them true, or restate them as forward-looking with the mechanism named.

### 3. arm64 is never parity-checked, and "covered on publish" overstates what publish does
**[verified]**

`docker-parity` builds with `load: true` and no `platforms:`, so it builds and runs **amd64 only**. #2714's `docker-publish` builds `linux/amd64,linux/arm64` under QEMU. So the arm64 manifest's first ever build is the release run, and nothing has ever *executed* it.

"Covered on publish" means only that it *builds*, not that it *behaves identically* — which is what this script exists to establish. The two reasons this PR's own comments give for arm64 mattering (Apple Silicon engineers, bff-client's arm64 devcontainer CI runner) are exactly the population that receives an unverified image.

Minimum: build multi-arch in `docker-parity` with `push: false` so the emulated build is exercised per-PR; better, add an arm64 `load` + run leg.

### 4. `docker-parity` is added but never promoted to a required status check
**[verified]** · `.github/workflows/build-and-test.yml`

The only required status check on `master` is `test-summary-publish`. Adding `docker-parity` to `publish.needs` gates the *release*, not merges.

Per `pull-requests.md` → "Introducing a new required CI check": this stack merges, a later PR breaks the Dockerfile, `docker-parity` goes red on `master`, nothing blocks that merge, every subsequent PR inherits a red result, red becomes expected, and more breakage lands unnoticed — until a release fails. An unpromoted check is worse than no check.

Ship the branch-protection change in this PR, or add a "Steps to merge" item naming the required-status-checks list, the owner of the setting change, and a ticket URL, and state that the check is advisory until promoted. (No job triggers on `merge_group`, so confirm there is no merge queue rather than assuming — if there is, promoting without `merge_group` coverage becomes the incident.)

### 5. The `Dockerfile:47-49` comment's stated mechanism is wrong
**[verified]**

```dockerfile
# `tsc` directly rather than `pnpm build`, to stay off the `prepack` script: that one
# also builds the redoc documentation bundle through webpack, which needs
# --openssl-legacy-provider and ships nothing this image serves.
RUN pnpm exec tsc
```

`package.json` on this branch has `"build": "tsc"`, so `pnpm build` and `pnpm exec tsc` are the same command. `prepack` is a pack/publish lifecycle script — `pnpm run build` never triggers it. The comment documents a hazard that does not exist and names a distinction that does not exist.

The underlying *fact* is right (you do not want `prepack`'s webpack/redoc bundle, and the image must ship no `oclif.manifest.json`). Only the mechanism is wrong. A maintainer who "simplifies" this line to `pnpm build` will believe they introduced a webpack dependency they did not. Rewrite against `prepack`, not against `pnpm build`.

### 6. `AGENTS.md:61` scopes a manifest claim repo-wide when it is true only of the image
**[verified]**

> oclif finds them by scanning this directory relative to the nearest `package.json`, with no manifest shipped. That is what lets the image drop a command by deleting its compiled file.

The npm package **does** ship one: `prepack` runs `oclif-dev manifest`, and `files` includes `"oclif.manifest.json"`. As written, a reader concludes that deleting a compiled command file removes the command in every distribution; in the npm package the manifest would still advertise it. (The equivalent comments in the `Dockerfile` and the parity script are correctly scoped to the image — no change needed there.)

---

## Suggestions

- **No `USER` instruction; the default `WORKDIR` is unwritable under the documented invocation.** `Dockerfile:52-81`. The comment calls `/workspace` "a sensible default for a bare `docker run`", but `WORKDIR` creates it `root:root 0755`. Under the `--user "$(id -u):$(id -g)"` this file itself calls load-bearing, `/workspace` is read-only, so a bare `docker run --user … image generate … -o .` fails on write. Conversely a genuinely bare `docker run` (no `--user`) runs as root and produces exactly the root-owned output the header says breaks rails-monolith's RSpec boot and bff-client's Gradle codegen. Add `USER node` (uid 1000 exists in the base image) and/or `RUN mkdir -p /workspace && chmod 1777 /workspace`. The parity script always passes both flags, so this gap is invisible to the gate.

- **Base image pinned by mutable tag, not digest.** `Dockerfile:28,52`. `node:22.22.1-bookworm-slim` is rebuilt for CVE patches under the same tag, so two builds of the same Spot commit can produce different images and the parity result is not reproducible. Append `@sha256:…`.

- **The generator-coverage `sed` is narrower than its comment promises.** `:57-66`. The extraction uses `[a-z0-9-]`, which covers all four current keys (`raw`, `json-schema`, `openapi2`, `openapi3` — verified). But a future key containing a dot, underscore or capital — `"openapi3.1": {` is the obvious one — never enters `DECLARED`; `COVERED` is also missing it, the two strings still compare equal, and the new generator ships uncompared with a green gate. The empty-parse guard at `:68` does not catch this because four other generators still parse. Widen the class, or state the restriction in the comment.

- **The detached container is not covered by the `EXIT` trap, and port 5600 is hardcoded.** `:192-212`. The trap removes `SCRATCH` only. If `set -e` fires between the `docker run --detach` and the `docker rm -f`, the container survives holding 5600. The `docker rm -f … || true` at `:193` recovers on the *next* run, but an unrelated local process on 5600 makes the `curl` at `:208` hit the wrong server and report a false OK. Extend the trap: `trap 'rm -rf "${SCRATCH}"; docker rm -f "${SERVER_NAME}" >/dev/null 2>&1 || true' EXIT`.

- **`pipefail` + `grep -q` can score a found readiness line as not-found.** `:204`. `grep -q` exits on first match and closes the pipe; `docker logs` takes SIGPIPE (141), and `-o pipefail` returns 141 for the pipeline. Direction is a false FAIL, not a false pass — so not a silent failure, but it makes the gate flaky, which is how gates get ignored. Capture then match.

- **The `node_modules`-on-any-parent walk stops one level short of its own claim.** `:167` — `while [ "${parent}" != "/" ]` never tests `/node_modules`. Either check `/`, or narrow the comment to "between the scratch dir and the filesystem root".

- **The `generate` matrix can pass on two empty directories.** `:106`. `diff --recursive` on two empty trees exits 0. `set -e` covers a *failing* `generate`, but not one that exits 0 having written nothing. The script guards this class explicitly everywhere else (`:68`, `:171`, `:184`) — this is the one exception. Add `[ -n "$(ls -A "${CLI_DIR}")" ]` before the diff.

- **`rm -f` cannot fail, including when it matches nothing.** `Dockerfile:66-68`. If the emit layout moves (an `outDir`/`rootDir` change, or `include` widening so the inferred root shifts), the build succeeds and the image silently ships `mock`, `docs` and `init` — long-running dev servers in an image documented as not carrying them. Assert each file exists before removing it.

- **The absent-command check scores on a message substring, not the failure route.** `:237` — `grep -qi "not found"` against merged stdout+stderr. Docker-level failures carry the same substring (`manifest … not found`, `Error response from daemon: … not found`), so a run that never reached the oclif dispatcher scores `OK`. Blast radius is limited (a wholly broken image aborts earlier under `set -e`), but the assertion is on the wrong axis.

- **The script cannot observe the provenance its header asserts.** `:2-7`, `:246`. It takes an arbitrary `<image-ref>` as `$1` and never verifies it was built from this checkout, yet the failure message says "at this commit". Running it against `ghcr.io/airtasker/spot:2.0.1` with a locally built CLI prints "All checks passed". Only the CI job establishes same-commit; `AGENTS.md:34` documents this as something a human runs by hand. State what is observed.

- **Provenance narration.** `:157-160` — "Spot used to need the consumer's node_modules…" narrates a fix made earlier in this stack; once merged, a reader cannot tell whether the caveat is live. ("The regression this whole image depends on" also reads backwards — you depend on the *fix*.) State the invariant instead.

- **Consumer names as rationale.** `Dockerfile:18-20`, `:144`, `:188` justify behaviour by naming rails-monolith and bff-client. Genuinely informative, but they rot when either consumer changes. Prefer decision-first, example-second.

- **`AGENTS.md:29`** documents `ts-lint spots --fix` under the description "Check contract TypeScript for…" — `--fix` rewrites in place.

- **`.dockerignore:4`** justifies a committed allowlist by `.claude/worktrees/`, which is untracked and developer-local, so unverifiable for the next reader. Generalise to "per-developer tooling directories".

- **`AGENTS.md:1-3`** still reads `# CLAUDE.md` / "This file provides guidance to Claude Code" — pre-existing, not in this diff, but worth fixing while the file is being edited.

---

## Verified clean

**Shell safety:** `set -ue -o pipefail` present at `:11` (all three options). No `grep` under `-e` outside an `if`/`elif` condition. No interpolated variable can resolve empty and trivially satisfy a comparison — `IMAGE` is arg-count-guarded, `REPO_ROOT` comes from `pwd` after a `set -e`-guarded `cd`. `local` declarations are separated from command substitution, so exit statuses are not masked. Both `|| true`s discard genuinely signal-free exit codes (leftover-container cleanup at `:193`, diff rendering after `fail` was already recorded at `:135`). The script exits 1 on any recorded failure, and the workflow step has no `continue-on-error` and no `|| true`, so the exit code becomes a job failure. The Dockerfile has no `RUN … || true` and no `&&` chain masking a middle command.

**The empty-comparison trap is explicitly closed** at `:68-73` — an empty `DECLARED` aborts rather than letting `"" != ""` pass a vacuous coverage check, with the reasoning recorded in the message. This is exactly the right instinct.

**Self-containment** (`:158-188`) asserts its own precondition (no `node_modules` on any parent path) *before* running so it cannot pass vacuously, and checks **both** the exit code and the produced artifact (`[ -f "${SCRATCH}/api.yml" ]`). `curl -fsS` at `:213` makes an HTTP 5xx non-zero, so a listening-but-broken server is caught. The absent-command check asserts a message and not merely non-zero.

**Image structure:** `.dockerignore` is an allowlist and all nine `!` patterns (`package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `docs/package.json`, `tsconfig.json`, `index.ts`, `bin`, `lib`, `cli`) have a corresponding `COPY`, and nothing `COPY`ed is missing. Every runtime dependency the CLI needs is in `dependencies`, so `pnpm install --prod` in the `prod-deps` stage is safe — the #2710 class of bug does not recur. `tsconfig.json` excludes `**/*.spec.ts` and `**/__spec-examples__/*`, so no test files compile into `build/` and oclif cannot register a spurious command. No `.npmrc` needed (zero `@airtasker`-scoped entries in the lockfile).

**Version alignment is exact and documented as coupled:** `nodejs 22.22.1` ↔ `node:22.22.1-bookworm-slim` (both stages) ↔ `corepack prepare pnpm@10.28.1` ↔ `"packageManager": "pnpm@10.28.1"`.

**Every other AGENTS.md claim verified against the tree:** `lib/src/parsers/`, `lib/src/ts-project.ts`, `lib/src/ts-lint/`, `lib/src/linting/` all exist; the six-command list matches `cli/src/commands/` minus the three deleted; `validation-server` does print "Validation server running on port …" and `/health` exists. The `parser/` → `parsers/` correction fixes a path that never existed.

No bare ticket keys, `TODO` or `FIXME` anywhere in the diff. `pnpm` throughout.

## Note on CI

The body says nothing will report here, "including the `docker-parity` job this PR adds". **It has** — `docker-parity` is green on this PR (3m26s), alongside `test-node-{18,20,22,24}` and `lint-check`. That is real evidence for the gate and worth correcting in the body.

## Strengths

- The `prod-deps` stage resolved independently rather than by pruning the builder is the right shape.
- 98 MB against a 250–350 MB estimate, and multi-arch buildx verified.
- The `HOME=/tmp` finding is excellent, and the write-up of *why the first gate missed it* — `ts-lint` fails via `process.exit` and was immune, while `validate`/`lint`/`generate` fail through oclif's handler which was broken — is exactly the reasoning that makes a gate trustworthy. Finding 1 is a refinement of that same instinct, not a contradiction of it.

timdawborn:
# Second-pass review — #2713

Multi-agent re-review of both commits, with the follow-up `56d3871` reviewed adversarially. Three agents worked this PR; every finding was verified by building the image, running the full parity script end to end, running the compiled CLI for every asserted status, and reading the real CI log for the `docker-parity` job.

**Verdict: the substance is right and the multi-arch step earns its keep.** But the two shell guards the follow-up added both have the same failure mode as the thing they were added to prevent, and the two Dockerfile behaviour changes are the one part of the follow-up the gate shipped alongside never exercises.

---

## Important

### 1. The new expected-status guard fails **open** when `expected` is not a number
`scripts/check-image-parity:139`

`[ "${cli_status}" -ne "${expected}" ]` — if a future call site omits the status (`compare_run "lint" lint "${CONTRACT}"`), `expected` becomes `"lint"`, `shift 2` eats the first real argument, and `[` errors with `integer expected` and returns 2. Sitting in an `if` condition it is exempt from `set -e`, and non-zero reads as **false** — so the guard is skipped, the mangled invocation runs on both sides, fails identically, and prints:

```
bash: line 7: [: lint: integer expected
OK: lint (exit 2, identical output)
still alive, failures=0
```

The only signal is one unread stderr line, in a script whose new comment at `:122-125` states that a case must not be able to degrade into an always-succeeding check. **The guard added to prevent that has the same failure mode one level up.**

```bash
case "${expected}" in
'' | *[!0-9]*)
  echo "ERROR: ${label}: expected status '${expected}' is not a number" >&2
  exit 1
  ;;
esac
```

### 2. The EXIT trap runs under `set -e`, so a failing `rm -rf` skips the container reap **and** rewrites the exit status
`scripts/check-image-parity:211`

`set -e` applies inside a trap body and aborts a `;`-list there. Verified:

```
$ bash -c 'set -ue -o pipefail; trap "false; echo SECOND-CLAUSE-RAN" EXIT; echo body; exit 0'
body
outer exit=1
```

`SECOND-CLAUSE-RAN` never printed, and a deliberate `exit 0` came out as 1. The recoverable command is ordered ahead of the consequential one, giving two consequences:

- **the container reap is skipped** — the exact leak this trap was added to prevent — leaving `spot-parity-validation-server` holding `--publish 5600:5600`. Invisible on a GitHub-hosted runner; on a dev machine or self-hosted runner a *concurrent* run then hits a port conflict or reads a stale server.
- **a fully passing run prints `All checks passed.` and exits 1** — CI red with no `FAIL:` line anywhere in the log. A misdiagnosis with no evidence attached.

```bash
trap 'docker rm -f "${SERVER_NAME}" >/dev/null 2>&1 || true; rm -rf "${SCRATCH}" || true' EXIT
```

Related, untouched by the follow-up: `:246`'s `docker rm -f "${SERVER_NAME}" >/dev/null` has no `|| true`, so a daemon hiccup aborts before the three absent-command checks and the `failures` summary. Already-printed `FAIL:` lines survive, so it is a wrong-reason abort rather than lost evidence — but `|| true` costs nothing.

### 3. `USER node` and `chmod 1777 /workspace` are entirely unexercised by the gate
`Dockerfile:90-91`; `scripts/check-image-parity:47,190,214` (`--user`) and `:49,192,216` (`--workdir`) — *two agents independently*

All three `docker run` invocations pass `--user "$(id -u):$(id -g)"` **and** pin `--workdir`, so neither `USER node` nor the default `WORKDIR /workspace` is ever the thing in effect. The follow-up's own opening argument is that a case which cannot observe the failure it exists for is not coverage; these two lines are in exactly that state, added in the same commit.

**Failure scenario:** someone drops `USER node`, reorders `WORKDIR` above the `chmod`, or a future base image stops shipping a `node` user. The gate stays green, the release ships, and the first consumer running the bare `docker run` form gets the ownership problem the Dockerfile header names. Both of these pass against the built image today:

```sh
[ "$(docker run --rm --entrypoint id "${IMAGE}" -u)" = 1000 ] \
  || fail "image does not default to uid 1000"
docker run --rm --entrypoint sh "${IMAGE}" -c 'touch /workspace/.probe' \
  || fail "default WORKDIR is not writable by the image's own user"
```

### 4. The new build-before-test instruction does not reproduce, and its remedy cannot fix the cause it names
`AGENTS.md:79-81` — *two agents independently*

> "The specs under `cli/src/commands/` call oclif's `Command.run`, which loads the config by scanning `oclif.commands` — `./build/cli/src/commands`. Run `pnpm build` before `pnpm test`, or a stale `build/` fails them with an unrelated oclif error."

The mechanism half is true — confirmed with `DEBUG='@oclif*'`, the spec's `Command.run` logs `loading IDs from …/build/cli/src/commands`. The consequence half is false, tested both ways:

- `build/cli/src/commands` renamed away, then all of `build/` removed → **all 9 tests / 3 suites pass, both times.** `globby.sync` returns `[]`, `Plugin._manifest` catches and warns, `Command.run` instantiates the class directly.
- `build/` present with a stale orphan command file that throws on require → `generate.spec.ts` still passes, with a `module: @oclif/config / task: toCached` warning to stderr.

And the prescribed remedy is wrong either way: `pnpm build` is bare `tsc`, which never deletes stale output — confirmed the orphan survives it. Only `prepack` does `rm -rf build`. Per the repo's own rule on verifying runtime-behaviour claims before encoding them in docs, drop the sentence; if the scan is worth documenting, keep only the first half.

### 5. `checksum` ships in the image and nothing compares it
`scripts/check-image-parity` (zero occurrences); `AGENTS.md:35`

`docker run <image> --help` lists six commands; the gate compares five. The script has a mechanism specifically to stop a new **generator** slipping through uncompared (with a hard `exit 1` if the parse yields nothing) but no equivalent for a **command** — so the gap that mechanism exists to prevent already exists one level up.

`checksum` is the costliest omission: consumers key cache invalidation on its output, so a divergence either misses a contract change or invalidates on every run. Trivially comparable — both sides return `ef62006f8d00d5899d7f0f03f3c1ca66d2b52be4`. Add `compare_run "checksum" 0 checksum "${CONTRACT}"`, and derive the command list from `cli/src/commands/*.ts` with an explicit exemption list for `mock`/`docs`/`init`, mirroring the `DECLARED`/`CASES` check.

Related: the self-containment block covers only `generate -g openapi3 -l yaml`. `validate`, `lint`, `ts-lint` and `checksum` always run with the repo's `node_modules` on a parent path, so the image's own resolution is unproven for them. All four work from a `mktemp -d` (tested), so this is coverage-only — but `ts-lint` has the largest runtime dependency surface and the mount would mask a missing prod dep.

### 6. `AGENTS.md:38` is correct here and stale at the stack tip
> "The image is built and parity-checked on every pull request; publishing it is not wired up yet."

Correct for this PR. #2714 lands `docker-publish` and touches only the workflow and `README.md`, so the top branch ships a working ghcr publish alongside a doc denying it exists. **Nothing needs to change in this PR** — flagging so it is fixed in #2714, where the note there recommends matching the README's Releases wording rather than restoring the old "cannot diverge" claim.

Also: "not wired up **yet**" is the temporal phrasing the org rules name explicitly.

### 7. The multi-arch comment now contradicts the step added six lines below it
`.github/workflows/build-and-test.yml:106-107`

> "Multi-arch cannot be loaded, and is covered on publish."

True before the follow-up. The follow-up then added, **in the same job**, a step whose own comment says the opposite — that building it here is what stops a release being the first time arm64 is built. Multi-arch *build* is now covered right here; only multi-arch *runtime behaviour* is uncovered anywhere. A reader trusting line 107 concludes arm64 is first built at release time, which is precisely the gap the follow-up closed. Rewrite to: "Multi-arch cannot be loaded, so the step below builds it without running it."

---

## Suggestions

- **The "no output to compare" guard is unreachable.** `:109-110`. `set -ue -o pipefail` is in force and lines `:105-106` are unguarded, so any failing `generate` aborts before line 109, and a `generate` that exits 0 has written its directory. The vacuous case is genuine (`diff --recursive` on two empty dirs exits 0) but its *reachable* form is partial output, not zero output. Counting closes it: assert `${#CASES[@]}` directories exist under both trees. As written it reads as coverage while being dead — the shape the commit set out to remove.

- **A crashed validation-server is diagnosed as a readiness-line failure, after 60 wasted seconds.** `:220-244`. Verified: with a bad contract path the container is `Running=false ExitCode=1` within 3s, logs reading `Error: File not found`. The loop's only exit is the readiness string, so it polls 60 times and then reports "never printed its readiness line" — pointing the reader at stdout buffering and the TTY contract rather than the crash. `:243`'s log dump does carry the truth, so this is diagnosis quality plus a minute of CI.

  ```bash
  [ "$(docker inspect -f '{{.State.Running}}' "${SERVER_NAME}")" = true ] \
    || { fail "validation-server exited before becoming ready"; break; }
  ```

- **Both build steps export to the same GHA cache scope.** `:111` and `:125`, both `cache-to: type=gha,mode=max` with the default scope, so each run's second export rewrites the index the first just wrote. Not a correctness problem (buildx keys records per platform) but it churns the cache. `scope: parity` / `scope: multiarch`.

- **`lint --no-trailing-forward-slash=error` produces byte-identical output to plain `lint`** on this fixture (`Found 0 errors and 0 warnings`), because no path in `test-fixtures/contract/` has a trailing slash. With `expected=0` asserted it now catches the flag being renamed or removed (oclif exits 2), which is most of its value — but it never checks that the severity override changes anything.

- **`AGENTS.md:48`** correctly lists `parsers/` (verified — `parser/` never existed) but omits `lib/src/parser.ts`, the actual `parse()` entry point every command imports. A gap rather than a scoping choice, since the neighbouring line already names a file.

- **`AGENTS.md:79`** — "which is why it sits outside `tsconfig.json`'s `include`" is not the distinguishing reason; `__spec-examples__` is kept out too, by `exclude`.

- **`AGENTS.md:29`** reads `npx @airtasker/spot ts-lint spots` while #2714 converts the README to `pnpm` throughout and the org rule is pnpm-only. `pnpm exec spot ts-lint <dir>`.

- **`AGENTS.md:1-3`** still opens `# CLAUDE.md` / "This file provides guidance to Claude Code" while `AGENTS.md` is the real file. Pre-existing on master, but this PR is already editing it.

- **Comment cuts.** `Dockerfile:69-70` restates the error string on line 73. `Dockerfile:66-68` restates the no-manifest fact from `:60-63`, ten lines apart. `scripts/check-image-parity:122-125` spends four lines on one point. `.github/workflows/build-and-test.yml:113-116` repeats the sibling step's "cannot be loaded" ten lines above. The `Dockerfile` sits at 60% comment lines.

- **`Dockerfile:86-89` overclaims twice in one sentence.** (a) The problem at the top of the file is that output must be owned by the *invoking* user; uid 1000 is not the invoking user either, so `USER node` changes *which* wrong owner it is — #2714's README bullet states this correctly, this comment does not. (b) The line opens with "Overridden by `--workdir` in normal use", so the *documented* invocation never lands in `/workspace`; the 1777 justification is about a **bare** run.

---

## Verified sound

- **All six expected statuses are correct** against the compiled CLI: `0,0,0,0,1,2`. Nothing spurious or vacuous. Inverting one produces `FAIL: ts-lint dirty: expected exit 0 from the CLI, got 1; this case no longer exercises what it was added for`.
- **The trap has no uncovered window** — it is installed before the `docker run --detach`, and `docker rm -f … || true` masks only "no such container", the expected state after the explicit reap.
- **The `while :;` walk terminates and is errexit-safe** — `set -e` does not fire on a false `[ … ] && break` in non-final position (verified empirically); `dirname /` is `/`, which the new `break` catches; `/tmp/a/b/c` terminates in 5 iterations.
- **`docker logs … || true` swallows less than it looks** — on an *exited* container `docker logs` still exits 0 and returns the real error; on a *removed* one the "No such container" text lands via `2>&1`. The diagnostic survives to `:243`.
- **The Dockerfile `for` loop works** — `exit 1` inside the single `RUN` does fail the build, and `rm -f "…/$c".*` removes every extension: the built image's commands dir holds exactly the six shipped commands × `{.js,.d.ts}` with no `mock*`/`docs*`/`init*` remnant. A surviving `mock.d.ts` would matter, since oclif scans the directory.
- **`USER node` introduces no read degradation** — `find /opt/spot -type d ! -perm -o=rx` and `-type f ! -perm -o=r` are both empty; a full read sweep as `node` produced zero errors. Bare `docker run` (uid 1000) and `--user 4242:4242` (no `/etc/passwd` entry) each wrote `/workspace/api.yml` and exited 0. `1777` is the only mode that keeps the documented arbitrary-`--user` invocation working — `chown node:node` + 0755 would break it.
- **The generator `sed` extracts exactly `json-schema openapi2 openapi3 raw`**, nothing spurious — nested keys sit at 4-space indent so the 2-space anchor excludes them.
- **The merged `2>&1` in `compare_run` is not an interleaving flake risk** — no command in the matrix writes to both streams.
- **`docker-parity` carries no `if:`**, so it runs on the release event and `publish`'s new `needs` is satisfiable rather than permanently skipped.
- `.dockerignore`'s nine `!` patterns each have a matching `COPY`; `.tool-versions` / base tag / `packageManager` are consistent at node 22.22.1 and pnpm 10.28.1; `AGENTS.md:35`'s command list matches `--help` exactly; the oclif-manifest rewording is correct (`prepack` generates it, `files` ships it, `postpack` removes it, the image ships none).
- **The gate passes end to end locally in 1m28s** and leaves no container behind.

## Strengths

- **The failing-command case genuinely earns its place.** `validate` on the type-error fixture is the only one of the six that exits through oclif's error handler (stderr-only, exit 2) rather than `process.exit` — with `expected` now asserted it is the single check that would catch the `HOME=/tmp` regression returning.
- **`RUN test -f … || exit 1` before the `rm -f`** turns a silent no-op into a build failure, and it runs on both architectures (visible as an arm64 step in the CI log).
- **The multi-arch step delivers what it claims, for 2m29s.** CI log shows `[linux/arm64 base 3/5] RUN corepack enable`, `[linux/arm64 prod-deps 1/1] RUN pnpm install --frozen-lockfile --prod` and `[linux/arm64 builder 5/5] RUN pnpm exec tsc` all executing — so an emulated install or compile failure fails the step despite no exporter. Emulation works without `setup-qemu-action` because the amd64 `moby/buildkit` image ships `/usr/bin/buildkit-qemu-aarch64`. That is the failure a release would otherwise have discovered first.
- `scripts/check-image-parity:10-11` — "Nothing here verifies the image was built from this checkout" states what the script does *not* observe. Exactly the right shape for a defensive script's header.

vlj91 (approved):
LGTM!

PR #2714, opened 2026-08-18
build(COMPASS-17): publish the docker image on release, docker-first README

timdawborn:
## Ticket

> **Stacked on #2713 → #2712 → #2711 → #2710 → #2709 → #2708.** Last of the spot-side PRs. `build-and-test` does report on this PR — `docker-parity`, `test-node-{18,20,22,24}` and `lint-check` are green; `publish` and `docker-publish` correctly show as skipped.

## What

A `docker-publish` job, and a docker-first rewrite of the README's usage section.

## Publishing

Gated on `github.event_name == 'release'` — **the same GitHub Release event that already triggers the npm publish**. One manual version bump and one Release publishes npm, GitHub Packages and `ghcr.io/airtasker/spot`, so a version means the same thing in every registry and there is no second thing to remember. No push, tag or branch trigger.

`needs: [test, lint-check, docker-parity]`, so an image that does not match the compiled CLI is never pushed. `GITHUB_TOKEN` for the ghcr login, because the existing npm job already pushes to GitHub Packages with it. QEMU + buildx, `linux/amd64,linux/arm64`, gha cache. **The npm publish job is untouched by this PR** — though note #2713 added `docker-parity` to its `needs`, so an npm release is now gated on docker parity too.

`docker-publish` and `publish` are siblings with no ordering between them, so they can fail independently and leave one registry with a version the other does not have. A `concurrency` group keeps two closely-spaced releases from racing on `latest`, `2` and `2.1`. The Docker Hub login is unconditional here, unlike in `docker-parity`: this job runs only on a release, where the secret always exists, so a skip could only mean the secret was renamed or emptied — and it would respond by pulling the base image anonymously across two platforms and three stages, failing the release with a rate-limit error pointing nowhere near the cause.

### `latest=auto`

A prerelease must not become what a consumer pulling `latest` gets. `latest=auto` applies `latest` only to non-prereleases, so the `v2.1.0-rc.1` rehearsal in the rollout below gets exactly one tag: `2.1.0-rc.1`.

It is worth knowing that this is `docker/metadata-action`'s **default**, not a deviation — it is named explicitly because `openapi-generators` uses `latest=true`, and copying that here would silently break the property. Withholding the `{{major}}` and `{{major}}.{{minor}}` tags from prereleases is `type=semver`'s own behaviour and applies regardless of this block.

## README

Rewritten around the container, outside the oclif-regenerated `<!-- toc -->` and `<!-- commands -->` markers so `prepack` cannot overwrite it.

The invocation is documented **argument by argument**, because every part of it prevents a failure that is otherwise silent:

- an `--out` outside the mount is written *inside* the container and the command **exits 0** — the file just never appears on the host;
- `--out ~/x` expands the tilde against the **container's** home directory;
- environment variables are not inherited from the calling shell;
- omitting `--user` leaves root-owned output that a later non-root run cannot overwrite — and which breaks rails-monolith's mtime cache and bff-client's Gradle incremental inputs.

Plus: the `validation-server` incantation with its readiness line and teardown advice, and a table of which commands the image carries and which stay npm-only.

**The npm package is explicitly stated to remain published and supported.** This adds a distribution; it does not deprecate one. spot is a public repo with external consumers.

The Releases section now records that a public repository pushes a **public** ghcr package by default, and that `ghcr.io/airtasker/spot` is deliberately private — so the visibility is worth re-checking after anything that could recreate the package.

## How this was verified

- The workflow parses, and both `docker-publish` and `publish` are gated on the release event and both require `docker-parity`.
- The image, its multi-arch build and the parity gate were verified in #2713. Nothing here changes the image.
- `pnpm lint:check` clean. README edits are outside the oclif markers, and the toc lists only h1 headings, none of which changed.

## Steps to merge — rollout, in order

This is the part that cannot be verified from the repo, and the failure modes are asymmetric: get the visibility wrong and the image is public; get the grants wrong and every consumer 403s.

1. **Merge the whole stack** (#2708 → #2713 → this).
2. **Publish a `v2.1.0-rc.1` GitHub Release, marked prerelease.** Same mechanism as a real release — no side channel — and `latest=auto` keeps it off `latest`.
3. **Flip the ghcr package `airtasker/spot` to private.** It will have been created public, because this repository is public. Do this before anything else.
4. **Grant read access** to the three consumer repositories' CI tokens, and to engineers' PATs. *(Open question from the plan: exactly which principals need `read:packages` — worth resolving here rather than guessing.)*
5. **Pull-test** from an arm64 laptop and using a consumer CI token, to prove both the architecture and the grants.
6. **Publish `v2.1.0`.**
7. Then the consumer migrations, in the planned order: enrichment-engine → bff-client → rails-monolith.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -130,6 +130,75 @@ jobs:
       - name: Compare the image against the CLI
         run: ./scripts/check-image-parity spot:parity
 
+  # Rides the same GitHub Release event as the npm publish below, so a version means
+  # the same thing on ghcr and on npm, and there is no second thing to remember.
+  docker-publish:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release'
+    needs:
+      - test
+      - lint-check
+      - docker-parity
+    permissions:
+      contents: read
+      packages: write
+    env:
+      DOCKERHUB_USERNAME: ${{ vars.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: docker/setup-qemu-action@v3
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Docker Hub
+        if: env.DOCKERHUB_TOKEN != ''
+        uses: docker/login-action@v4
+        with:
+          username: ${{ env.DOCKERHUB_USERNAME }}
+          password: ${{ env.DOCKERHUB_TOKEN }}
+
+      # GITHUB_TOKEN, because the npm publish job below already pushes to GitHub
+      # Packages with it.
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/airtasker/spot
+          # The release ref is the version tag, so these yield 2.1.0, 2.1 and 2. The
+          # commit is recorded in the org.opencontainers.image.revision label that
+          # metadata-action emits, so no separate sha tag is needed to trace an image
+          # back to its source.
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+          # auto, not true: a prerelease Release must not become `latest`. The
+          # rollout for this rehearses with a v2.1.0-rc.1 prerelease, and under
+          # `latest=true` that rc would become what every consumer pulling `latest`
+          # gets. auto also withholds the {{major}} and {{major}}.{{minor}} tags from
+          # prereleases, which is the same reasoning.
+          flavor: |
+            latest=auto
+
+      # arm64 is not optional: engineers run Spot on Apple Silicon, and bff-client's
+      # devcontainer CI runner is arm64.
+      - uses: docker/build-push-action@v6
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
   publish:
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -135,6 +135,11 @@ jobs:
   docker-publish:
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
+    # Two releases published close together would otherwise race on which one ends up
+    # as `latest`, `2` and `2.1`.
+    concurrency:
+      group: docker-publish
+      cancel-in-progress: false
     needs:
       - test
       - lint-check
@@ -151,8 +156,12 @@ jobs:
       - uses: docker/setup-qemu-action@v3
       - uses: docker/setup-buildx-action@v3
 
+      # Unconditional here, unlike in docker-parity: this job runs only on a release,
+      # where the secret is always available. A skip could then only mean the secret has
+      # been renamed or emptied, and skipping would turn that into an anonymous pull of
+      # the base image across two platforms and three stages — a rate-limit failure
+      # pointing nowhere near the missing secret.
       - name: Log in to Docker Hub
-        if: env.DOCKERHUB_TOKEN != ''
         uses: docker/login-action@v4
         with:
           username: ${{ env.DOCKERHUB_USERNAME }}
@@ -179,11 +188,12 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-          # auto, not true: a prerelease Release must not become `latest`. The
-          # rollout for this rehearses with a v2.1.0-rc.1 prerelease, and under
-          # `latest=true` that rc would become what every consumer pulling `latest`
-          # gets. auto also withholds the {{major}} and {{major}}.{{minor}} tags from
-          # prereleases, which is the same reasoning.
+          # A prerelease must not become what a consumer pulling `latest` gets.
+          # `latest=auto` is metadata-action's default and applies `latest` only to
+          # non-prereleases; it is named explicitly because copying `latest=true` from
+          # a neighbouring repository would silently break that. Withholding
+          # {{major}} and {{major}}.{{minor}} from prereleases is `type=semver`'s own
+          # behaviour, not this block's.
           flavor: |
             latest=auto
 
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -130,8 +130,8 @@ jobs:
       - name: Compare the image against the CLI
         run: ./scripts/check-image-parity spot:parity
 
-  # Rides the same GitHub Release event as the npm publish below, so a version means
-  # the same thing on ghcr and on npm, and there is no second thing to remember.
+  # Runs on the GitHub Release event, in parallel with the npm publish below and
+  # independently of it.
   docker-publish:
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
@@ -156,12 +156,12 @@ jobs:
       - uses: docker/setup-qemu-action@v3
       - uses: docker/setup-buildx-action@v3
 
-      # Unconditional here, unlike in docker-parity: this job runs only on a release,
-      # where the secret is always available. A skip could then only mean the secret has
-      # been renamed or emptied, and skipping would turn that into an anonymous pull of
-      # the base image across two platforms and three stages — a rate-limit failure
-      # pointing nowhere near the missing secret.
+      # Conditional, as in docker-parity: these credentials are not configured on
+      # this repository, and login-action throws rather than skipping when both are
+      # empty. Without them the base image is pulled anonymously, which is what
+      # docker-parity has always done.
       - name: Log in to Docker Hub
+        if: env.DOCKERHUB_TOKEN != ''
         uses: docker/login-action@v4
         with:
           username: ${{ env.DOCKERHUB_USERNAME }}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,18 +24,18 @@ Spot is a TypeScript-based API contract definition tool that generates OpenAPI, 
 - `pnpm prettier:check` / `pnpm prettier:fix` - Run Prettier only
 
 ### CLI Usage (after build)
-- `npx @airtasker/spot generate --contract api.ts --generator openapi3 --out output/` - Generate OpenAPI3 from contract
-- `npx @airtasker/spot lint api.ts` - Lint a Spot contract
-- `npx @airtasker/spot ts-lint spots` - Check the TypeScript in a contract tree for formatting, lint and type errors, under configuration Spot bundles itself (`--fix` rewrites in place)
-- `npx @airtasker/spot mock api.ts` - Run mock server from contract
-- `npx @airtasker/spot validate api.ts` - Validate contract syntax
+- `pnpm exec spot generate --contract api.ts --generator openapi3 --language yaml --out doc/output` - Generate OpenAPI3 from contract
+- `pnpm exec spot lint api.ts` - Lint a Spot contract
+- `pnpm exec spot ts-lint <dir>` - Check the TypeScript in a contract tree for formatting, lint and type errors, under configuration Spot bundles itself (`--fix` rewrites in place)
+- `pnpm exec spot mock api.ts` - Run mock server from contract
+- `pnpm exec spot validate api.ts` - Validate contract syntax
 
 ### Docker image
 - `scripts/check-image-parity <image-ref>` - Compare the image against the compiled CLI. Run `pnpm build` first. This is what the `docker-parity` CI job runs.
 - The image carries `generate`, `validate`, `lint`, `ts-lint`, `checksum` and `validation-server`. `mock`, `docs` and `init` are npm-only, and the Dockerfile deletes their compiled files.
 - `.dockerignore` is an allowlist. Every `!pattern` in it must have a matching `COPY` in the Dockerfile.
 - The `node` base image tag, the `nodejs` line in `.tool-versions`, and the corepack pin against `packageManager` in `package.json` all name the same versions. Move them together.
-- The npm package is published from the GitHub Release event. The image is built and parity-checked on every pull request; publishing it is not wired up yet.
+- Both the npm package and the image publish from the GitHub Release event. The jobs are siblings with no ordering, so one registry can end up with a version the other does not. The image is also built and parity-checked on every pull request.
 
 ## Architecture
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ pnpm add @airtasker/spot
 You can pass the definition above to a generator by simply running:
 
 ```sh
-npx @airtasker/spot generate --contract api.ts
+pnpm exec spot generate --contract api.ts --generator openapi3 --language yaml --out doc/output
 ```
 
 # Why we built Spot
@@ -111,20 +111,20 @@ We built Spot with this in mind. Instead of having to juggle various API format 
 
 Spot ships two ways, built from the same source at the same version.
 
+- **The npm package `@airtasker/spot`** is published to npm and to GitHub Packages and
+  is the distribution available to everyone. It is what you want for interactive local
+  use, for the in-memory API below, and for the commands the image does not carry.
 - **The Docker image `ghcr.io/airtasker/spot`** needs no Node toolchain and no
-  `node_modules` beside your contracts. This is the recommended way to run Spot in a
-  repository that is not otherwise a Node project.
-- **The npm package `@airtasker/spot`** is published to npm and to GitHub Packages, and
-  is fully supported. Nothing here deprecates it. It is what you want for interactive
-  local use, for the in-memory API below, and for the commands the image does not
-  carry.
+  `node_modules` beside your contracts, which suits a repository that is not otherwise
+  a Node project. The package is **private to Airtasker** — if you are outside the
+  organisation, use the npm package.
 
 ## Docker
 
 ```sh
 docker run --rm --user "$(id -u):$(id -g)" \
   --volume "$PWD:$PWD" --workdir "$PWD" \
-  ghcr.io/airtasker/spot:2.1.0 \
+  ghcr.io/airtasker/spot:<version> \
   generate --contract api.ts --generator openapi3 --language yaml --out doc/output
 ```
 
@@ -135,8 +135,9 @@ docker run --rm --user "$(id -u):$(id -g)" \
 echo "$GH_PACKAGES_READONLY_TOKEN" | docker login ghcr.io -u "$USER" --password-stdin
 ```
 
-Tags follow semver: `2.1.0`, `2.1`, `2`, and `latest`. Images are built for
-`linux/amd64` and `linux/arm64`.
+Substitute the release you want for `<version>`; the published tags are the full
+version, the major and minor prefixes, and `latest` for the most recent non-prerelease.
+Images are built for `linux/amd64` and `linux/arm64`.
 
 ### The invocation, argument by argument
 
@@ -154,15 +155,19 @@ that are not always loud:
 
 All of these produce a wrong result rather than an error:
 
-- **An `--out` outside the mount** is written *inside* the container, and the command
-  exits 0. The file simply never appears on the host. Spot logs the absolute path it
-  wrote to — check it against where you expected.
-- **`--out ~/x`** expands the tilde against the **container's** home directory, not
-  yours. Use a path under the mount.
+- **An `--out` outside the mount** is written *inside* the container, so the file never
+  appears on the host. Where the container user cannot write that path this fails
+  loudly; where it can — anywhere under `/tmp`, which is the container's home — the
+  command exits 0. Spot logs the absolute path it wrote to; check it against where you
+  expected.
+- **`--out ~/x`** is that writable case: the tilde expands against the **container's**
+  home directory, not yours, so it succeeds and lands nowhere useful. Use a path under
+  the mount.
 - **Environment variables are not inherited** from your shell. Pass them with
   `--env`.
-- **Omitting `--user`** leaves root-owned output, which a later non-root run cannot
-  overwrite.
+- **Omitting `--user`** writes output as the image's own unprivileged user rather than
+  as you, which a later run as yourself may not be able to overwrite. Downstream builds
+  that key a cache or an incremental task input on the output's mtime stall on this.
 
 ### Long-running: the validation server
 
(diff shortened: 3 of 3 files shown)


timdawborn:
# Review — #2714 `build(COMPASS-17): publish the docker image on release, docker-first README`

Automated multi-agent review (`/pr-review-toolkit:review-pr`), scoped to this PR's diff against its base (#2713), and cross-checked against #2713's Dockerfile, parity script and workflow. Findings marked **[verified]** were checked against the tree or the CLI source.

**No Critical findings.** On the security question specifically: the publish job cannot be reached from a PR or a fork (`on: pull_request` + `if: github.event_name == 'release'`), `permissions:` is scoped to `contents: read` / `packages: write` with no `id-token`, and `needs: docker-parity` means no image ships without the gate. That part is right.

---

## Important

### 1. The README pins an image tag that will not exist
`README.md:127`, `:138`, `:174`

`package.json` on this branch is `"version": "2.0.3"`, but the README's headline command and the validation-server command both say `ghcr.io/airtasker/spot:2.1.0`. The moment the stack merges, the first thing a reader copies out of the public README fails with `manifest unknown` — and it re-stales at every release after 2.1.0.

#2713's `Dockerfile` header already gets this right with `ghcr.io/airtasker/spot:<version>`. Same treatment here, or use a tag the workflow always produces.

### 2. arm64 ships having never been executed
**[verified]** · `build-and-test.yml:181` vs #2713's `docker-parity` job

`docker-parity` builds `load: true` with no `platforms:`, so it builds and runs **amd64 only** on `ubuntu-latest`. `docker-publish` builds `linux/amd64,linux/arm64` under QEMU. The arm64 manifest's first ever build is the release run, and no gate has ever executed it.

**Failure scenario:** a prod dependency that resolves per-arch, or a QEMU-emulation quirk in the `pnpm install --prod` stage, produces an arm64 image that cannot run `spot generate`. Parity is green, the release publishes, and every Apple Silicon engineer and bff-client's arm64 devcontainer runner — the two reasons the comment at `:176` gives for arm64 mattering — gets a broken `2.1.0`.

Minimum fix: build multi-arch in `docker-parity` (`platforms:` + `push: false`) so the emulated build is exercised on every PR; better, add an arm64 `load` + run leg.

### 3. npm and ghcr can diverge, and the README states an invariant the workflow does not enforce
`build-and-test.yml:121`/`:188`, `README.md:464-466`

`publish` and `docker-publish` are siblings: same `needs:`, no ordering between them. `pnpm publish` and `push: true` run in parallel. If the multi-arch build fails — per finding 2, the step most likely to — npm has `2.1.0` and ghcr does not.

`README.md:464` says the one event "publishes the npm package, the GitHub Packages copy, and the `ghcr.io/airtasker/spot` image, so a version means the same thing everywhere". One event *triggering* both is not both *succeeding* together. Either make `publish` `needs: docker-publish`, or soften the README to what the workflow guarantees and document the recovery path.

### 4. The public README's "recommended" path is a private package
`README.md:114-116`, `:131`, `:468-471`

Spot is a public repo with external consumers — the npm downloads badge is at `:99`. The README calls Docker "the recommended way to run Spot" and then tells the reader the package is private and needs a `read:packages` token. An external user gets `denied`.

Nothing in the workflow makes or keeps it private, either. `:468` correctly states that the first push from a public repository creates a **public** package and that the flip to private is a manual rollout step — but there is no automated re-check for a property the README asserts as fact, and the rollout list itself flags "which principals need `read:packages`" as an open question.

Say plainly that the image is Airtasker-internal and npm is the external distribution, or make the package public.

### 5. The Docker Hub login guard was copied from a job where it means something to one where it cannot
`build-and-test.yml:141`

In #2713's `docker-parity`, `if: env.DOCKERHUB_TOKEN != ''` correctly means "this is a fork PR with no secrets", and the comment there says exactly that.

`docker-publish` runs only on `release`, where secrets are always present. So the condition can never mean "fork" — it can only mean "the secret was renamed, deleted, or rotated to empty", and it responds by **silently skipping login**. The multi-arch build then pulls `node:22.22.1-bookworm-slim` anonymously across two platforms and three stages on a shared-egress runner, and can hit the rate limit — failing the release with an error that points nowhere near the missing secret. Make the login unconditional here.

---

## Suggestions

- **Rollout narration in a permanent workflow file.** `build-and-test.yml:168-172`: *"The rollout for this rehearses with a v2.1.0-rc.1 prerelease, and under `latest=true` that rc would become what every consumer pulling `latest` gets."* This is the change-commentary shape `rules.md` calls out — a one-off event described in a file that outlives it. Six months on a reader cannot tell whether the rehearsal is still pending. Keep the first sentence (the invariant), drop the rollout.

  Two accuracy problems in the same comment, worth fixing while it is being rewritten:
  - `latest=auto` is `docker/metadata-action`'s **default**, so the emphatic "auto, not true" reads as a deviation that is not one.
  - "auto also withholds the `{{major}}` and `{{major}}.{{minor}}` tags from prereleases" attributes to `flavor:` a behaviour belonging to `type=semver` itself, which applies regardless. Naming the wrong mechanism means a future edit removing `flavor:` looks like it would start tagging rc builds `2.1`, and someone re-adds it for the wrong reason.

- **No `concurrency:` group** on `docker-publish` (`:121`). Two Releases published in quick succession race on which one lands as `latest`, `2`, and `2.1`.

- **The out-of-mount claim is narrower than stated.** `README.md:157-159` — "An `--out` outside the mount is written *inside* the container, and the command exits 0." With the `--user "$(id -u):$(id -g)"` the same page mandates, the container uid cannot write `/`, `/opt`, or a new top-level directory, so most out-of-mount paths fail loudly with EACCES. The reliably-silent case is a *writable* out-of-mount path — i.e. `HOME=/tmp`, which is already the next bullet. Narrow the claim, or add it to `check-image-parity` so it is enforced rather than asserted.

- **Two tables in one README disagree about which commands exist.** `README.md:198` vs `:246-254` — the new "What the image carries" table lists `ts-lint`, but the oclif-generated `<!-- commands -->` block still does not document it (pre-existing from #2711, not caused here).

- **Docker-first framing vs npm-first ordering.** `README.md:49-68` — Requirements now says "Either Docker, or:" but Installation immediately below is npm-only and sits ~60 lines before the Docker section, so a docker-first reader hits `pnpm add @airtasker/spot` first.

- **`npx` in a pnpm-exclusive repo.** `README.md:67`, `:216` still use `npx`; right after a `pnpm add` line, `pnpm exec spot …` reads better under the org's pnpm rule. Separately, both of those lines show `generate --contract api.ts` with no other flags — correct for interactive npm use, but #2712 makes that exit 2 in any non-TTY context, which is the container path this README now leads with. Worth a note that the flags are required non-interactively.

- **Third-party actions are tagged, not SHA-pinned**, in the one job holding `packages: write` and pushing to a registry. Repo-wide convention is tags (cf. `test-summary/action@v2.4`), so this is not a regression — but a publish job is where pinning earns its cost. Worth a decision, not necessarily in this PR.

---

## Verified clean

- **The publish job is genuinely unreachable from a fork or PR.** `on:` has no `push`/`tag` trigger for it; `if: github.event_name == 'release'` gates it; `permissions:` is minimal and omits `id-token`.
- **`needs: [test, lint-check, docker-parity]`** means the compiled-CLI comparison from #2713 blocks the push rather than merely reporting alongside it.
- **The README's verifiable claims check out against source:** `-p 5600` is passed explicitly because the CLI default is `5907` (`cli/src/commands/validation-server.ts`); `GET /health` really exists (`lib/src/validation-server/server.ts:27`); the readiness line really is printed after `listen` resolves; `generate` really logs an absolute path (`lib/src/io/output.ts`).
- **README edits sit outside the oclif-regenerated `<!-- toc -->` and `<!-- commands -->` markers**, so `prepack` cannot overwrite them.
- **The npm publish job is untouched by this PR** — though note the net effect across the stack is that #2713 added `docker-parity` to its `needs`, so an npm release is now gated on docker parity. Worth stating, since the body's "The npm publish job is untouched" is true of this PR but reads as a stack-level claim.
- `CLAUDE.md` is a symlink to `AGENTS.md`, so there is no context-file drift between them.
- No bare ticket keys in the README or workflow.

## Note on CI

The body says nothing will report here. **It has** — `docker-parity` (3m41s), `test-node-{18,20,22,24}` and `lint-check` are all green on this PR; `docker-publish` and `publish` correctly show as skipped.

## Strengths

- **`latest=auto` over the copy-pasted `latest=true` from `openapi-generators` is the right call** and prevents a real footgun — a prerelease rc becoming what every consumer pulling `latest` gets. The reasoning is sound even though the comment stating it needs the rewrite above.
- Riding the existing GitHub Release event rather than inventing a second trigger means one version bump publishes everywhere, with nothing extra to remember.
- The README documents the invocation argument by argument, and each argument genuinely prevents a silent failure — that is the right level of detail for a container interface where the failure modes do not show up in an exit code.
- Explicitly stating that the npm package remains published and supported is the right call for a public repo with external consumers.
- The "Steps to merge" rollout list is unusually good: it sequences the ghcr private-flip *before* granting access, rehearses with a prerelease, and pull-tests both the architecture and the grants before the real release. Finding 4 is about making the README match that reality, not about the plan.

timdawborn:
# Second-pass review — #2714

Multi-agent re-review of both commits, with the follow-up `15ed72c` reviewed adversarially. Verified by building the image from #2713's Dockerfile (confirmed byte-identical between the two branches), driving every README claim against it including Linux uid semantics via a named volume, reading `docker/login-action@v4` and `docker/metadata-action@v5` sources, and querying this repository's actual Actions secrets and variables.

**Verdict: one Critical, and it is in the follow-up.** The unconditional Docker Hub login makes every release fail, and is strictly worse than the conditional version it replaced. Several README claims were fixed in one place and left standing in their sibling 17 lines away.

---

## Critical

### 1. The unconditional Docker Hub login will fail every release — the credentials do not exist in this repository
`.github/workflows/build-and-test.yml:164-168`

The follow-up's justification — "this job runs only on a release, where the secret is always available" — is false here. Measured:

```
repos/airtasker/spot/actions/variables            → {"variables":[],"total_count":0}
repos/airtasker/spot/actions/secrets              → total_count 1: NPM_TOKEN only
repos/airtasker/spot/actions/organization-secrets   → total_count 0
repos/airtasker/spot/actions/organization-variables → DOCKER_PUSH_CACHE_FROM, DOCKER_PUSH_CACHE_TO only
repos/airtasker/spot/environments                 → total_count 0
```

No `DOCKERHUB_TOKEN` secret and no `DOCKERHUB_USERNAME` variable at repo, org-visible, or environment scope. Both interpolate to `''`, and `docker/login-action@v4` does not degrade — `loginStandard` in `src/docker.ts` opens with:

```ts
if (!username && !password) {
  throw new Error('Username and password required');
}
```

**Failure scenario:** the first `v2.1.0-rc.1` Release after this merges. `docker-publish` fails at step 3 of 7, before the ghcr login and before the build. No image is pushed, at any version, ever. Meanwhile `publish` is a sibling with no ordering and publishes to npm normally — producing precisely the npm/ghcr divergence this follow-up added README prose to describe, for an entirely self-inflicted reason. Step 5 of the rollout ("pull-test from an arm64 laptop") is the first thing that would catch it.

**And the trade is strictly worse than the version it replaced.** The old `if: env.DOCKERHUB_TOKEN != ''` would have skipped, pulled `node:22.22.1-bookworm-slim` anonymously, and succeeded — which is exactly what `docker-parity`'s identical conditional login has always done, and it is green. The comment's rate-limit reasoning is a hypothesis this repo's own CI history contradicts. Worth stating plainly: because that secret has never existed, `docker/login-action` has **never once executed successfully** in this repo, so `docker-parity` being green is not evidence the login step works.

Preferred fix — keep the "fail early, name the cause" intent, but fail on a named precondition rather than on a third-party action's empty-credential error:

```yaml
      - name: Check the Docker Hub credentials are configured
        run: |
          if [ -z "$DOCKERHUB_USERNAME" ] || [ -z "$DOCKERHUB_TOKEN" ]; then
            echo "::error::vars.DOCKERHUB_USERNAME / secrets.DOCKERHUB_TOKEN are not set on this repository"
            exit 1
          fi
```

Either way, **"create `vars.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_TOKEN`" has to become step 1 of the Steps to merge list**, which currently does not mention Docker Hub credentials at all. If they are not going to be created, restore the condition.

---

## Important

### 2. `AGENTS.md:38` states the opposite of what this PR does
> "The npm package is published from the GitHub Release event. The image is built and parity-checked on every pull request; publishing it is not wired up yet."

#2713's follow-up wrote that line, and it is correct for #2713 in isolation. **This PR is what wires it up**, and it touches only the workflow and `README.md` — so the line is false the moment it merges, in the file every agent on this repo reads first. Intra-PR doc/code drift, directly.

Do **not** restore the pre-#2713 wording ("Both distributions are published from one GitHub Release event, so a version means the same thing on npm and on ghcr") — this follow-up explicitly rejected that "cannot diverge" claim for the README, and the same objection applies. Match the README's Releases section:

> Both the npm package and the image publish from the GitHub Release event. The jobs are siblings with no ordering, so one registry can end up with a version the other does not; the image is also built and parity-checked on every pull request.

### 3. The claim this follow-up retracted is still standing in the workflow, 40 lines above the job it is wrong about
`.github/workflows/build-and-test.yml:133-134`

> "Rides the same GitHub Release event as the npm publish below, so a version means the same thing on ghcr and on npm, and there is no second thing to remember."

The commit body says: *"it claimed npm and ghcr cannot diverge. One release event starting both publishes is not both succeeding."* It fixed `README.md` and left the identical claim in the workflow. Rewrite to: `# Runs on the GitHub Release event, in parallel with the npm publish below and independently of it.`

### 4. `README.md:151` still says "root" — the sibling of the claim fixed at `:165`
> table: "Output is owned by you, not by **root**. … an mtime cache or an incremental-build input on a **root-owned** file breaks it."
> bullet: "**Omitting `--user`** writes output as the image's own unprivileged user rather than as you…"

The follow-up correctly rewrote the bullet because #2713's `USER node` means output is no longer root-owned (verified: `docker run` with no `--user` gives `uid=1000(node)`), and left the table row 17 lines above asserting root twice. Same underlying inaccuracy, same section, one commit later — the exact pattern where the next review round reports it as a new finding. Grep the file for `root` before calling this fixed.

Suggested row: "Output is owned by you rather than by the image's own user (uid 1000) — load-bearing wherever a build system reads Spot's output."

### 5. The tilde claim is falsified by the command as a reader would type it
`README.md:163-165`

The follow-up strengthened this to "it succeeds and lands nowhere useful", under a heading promising "All of these produce a wrong result rather than an error". Measured against the image, from a shell:

```
$ docker run --rm --user 501:20 --volume "$PWD:$PWD" --workdir "$PWD" spot:check \
    generate --contract api.ts --generator openapi3 --language yaml --out ~/tildecheck
    Error: EACCES: permission denied, mkdir '/Users/tim/tildecheck'
exit=1
```

The **host** shell expands `~` before docker sees it, so the container gets an absolute host path it cannot create — a loud failure. The described behaviour appears only when the tilde is protected:

```
$ ... --out '~/tildecheck2'
Generated /tmp/tildecheck2/api.yml
exit=0
```

That is almost certainly how the commit message's verification was performed — quoted, which is not how the bullet is written. The bullet is not worthless: an argument list that is never shell-expanded (a Gradle `exec` args list, a compose `command:` array, a JSON argv) passes the literal `~` through and hits exactly this, and that is the build-system consumer case the whole section exists for. Name the condition rather than deleting it.

Also, the mechanism is Spot's own `expandPathWithTilde` → `os.homedir()` → `HOME=/tmp`, not shell expansion — as written a reader cannot reproduce it.

### 6. Nothing asserts the release tag matches `package.json`, so the PR's central guarantee is unenforced
`.github/workflows/build-and-test.yml:183-190` and `:236`

`docker-publish` derives the image tag from the git ref (metadata-action's `procSemver`: `vraw = this.context.ref.replace(/^refs\/tags\//g, '')`). `publish` derives the npm version from `package.json`. No step compares them, and the repo has no release automation — `v2.0.3` was a hand-written `chore(release):` PR followed by a manually created Release.

**Failure scenario:** a Release is cut from tag `v2.1.0` while `package.json` still says `2.0.3` — exactly the state this branch is in today, and exactly the mistake the follow-up says the pinned README fell into. `publish` fails on npm (`2.0.3` already exists) while `docker-publish` succeeds and pushes `:2.1.0`, `:2.1`, `:2`, `:latest`. Consumers pin `2.1.0` and get an image whose `spot --help` reports `@airtasker/spot/2.0.3` (verified against the built image) and whose `org.opencontainers.image.version` label says `2.1.0`.

```yaml
      - name: Check the tag matches the package version
        run: |
          pkg=$(node -p "require('./package.json').version")
          if [ "$pkg" != "${{ steps.meta.outputs.version }}" ]; then
            echo "::error::release tag ${{ steps.meta.outputs.version }} != package.json $pkg"
            exit 1
          fi
```

### 7. `README.md:51` contradicts the private-image framing the follow-up introduced
`### Requirements` now opens `Either Docker, or:` — offering the container as a peer alternative to Node, to every reader, 66 lines above `:119`'s "The package is **private to Airtasker** — if you are outside the organisation, use the npm package." An external consumer follows the Requirements line, installs Docker, and 403s on the pull.

Same missed-sibling shape as finding 4: the follow-up's whole purpose was "npm leads, the image says plainly that it is internal", and it revised `:114-120` without revisiting the Requirements line. Suggest `Node.js >= 18.12.0 and pnpm — or, inside Airtasker, Docker (see [Usage](#usage))`.

Minor, same area: `:115-117` says "The **package** is private to Airtasker" one line after a bullet about the npm *package*. Say "The **image** is private to Airtasker".

### 8. Omitting `--user` is a loud failure in the common case, not a quiet one
`README.md:165-167`

Tested under real Linux uid semantics (a named volume, since macOS bind mounts virtualise ownership). Three cases, no `--user`:

| Workspace | Result |
|---|---|
| owned by uid 4242, mode 755 (the normal case — your uid is not 1000) | `EACCES: permission denied, mkdir '/w/proj/doc/output'`, exit 1, nothing written |
| owned by uid 1000 (single-user Linux box) | exit 0, output owned by 1000 = you — **no problem at all** |
| owned by uid 4242, mode 777 | exit 0, output owned by 1000 ≠ you — matches the bullet |

Only the third is the described failure, and it needs a world-writable workspace. #2713's `USER node` largely *closed* this quiet failure mode; the follow-up updated the ownership detail but left the bullet in the "Quiet failure modes" list with a stall-the-downstream-build consequence that is now the less likely outcome. Either move it out of that list, or scope it to the world-writable case.

### 9. The enumeration of quietly-writable out-of-mount paths is incomplete and mis-attributed
`README.md:158-162`

"where it can — anywhere under `/tmp`, which is the container's home — the command exits 0" reads as the exhaustive set, and a reader will use it as a checklist. Measured, all exit 0 with nothing on the host:

```
--out /tmp/x        → Generated /tmp/x/api.yml        exit=0
--out /var/tmp/x    → Generated /var/tmp/x/api.yml    exit=0
--out /workspace/x  → Generated /workspace/x/api.yml  exit=0
--out /opt/nope     → EACCES, exit=1   (the loud case, as documented ✓)
```

The predicate is "world-writable", not "under `/tmp`", and "which is the container's home" is not the reason `/tmp` is writable — mode 1777 is. `/workspace` is the notable omission: #2713 deliberately `chmod 1777`s it and makes it the default `WORKDIR`, so it is the image's *own* quietly-writable directory. State the predicate, not one example of it.

---

## Suggestions

- **`cancel-in-progress: false` silently drops a release's image when a third arrives.** `:140-142`. The group name is fine — concurrency groups are repository-scoped, this is the only user of the literal, and including the ref would *defeat* the serialisation. But when a job is queued into an occupied group and a newer one arrives, GitHub **cancels the previously pending job**. Three releases inside the multi-arch build window — plausible on a rollout day publishing `rc.1`, `rc.2`, then `2.1.0` — and the middle release's image never exists while npm has all three. It reports as **cancelled**, not failed, so the README's "re-run the `docker-publish` job" advice is reached through "Re-run failed jobs", which will not include it. Either say so, or note it in the Releases section.

- **Four unpinned third-party actions in the repo's highest-privilege job.** `:156-157, 165, 173, 180, 202` — `docker/setup-qemu-action@v3`, `setup-buildx-action@v3`, `login-action@v4`, `metadata-action@v5`, `build-push-action@v6`, all mutable major tags, in the only job holding `packages: write` plus a Docker Hub credential. A moved tag exfiltrates both. The repo's recent trajectory is explicitly the other way (`use github-hosted runners for public repo security`, `pin test-summary/action to v2.4`, `pnpm/action-setup@v6.0.9`), so majors here are the outlier. Also `metadata-action` is on `v5` while `v6.2.0` is current — not a bug, `latest=auto` is the default in both.

- **The `npx` → `pnpm` sweep stopped at the README.** `AGENTS.md:27-31` still carries five `npx @airtasker/spot …` lines while this follow-up converted every README invocation. Same file as finding 2, so one edit closes both.

- **Two README snippets are now byte-identical.** `:67` (Installation) and `:221` (Usage → npm) both read `pnpm exec spot generate --contract api.ts --generator openapi3 --language yaml --out doc/output`. On master they were near-duplicates with shorter text; this follow-up expanded both to the same full command. Trim the Installation copy or link to `#usage`.

- **Comment trims.** `:159-163` spends five lines defending the *removal* of an `if:` — the step is now unconditional, so a skip is impossible; two lines suffice. `:192-197` correctly states that `latest=auto` is the default and that `type=semver` withholds prerelease major/minor, but "because copying `latest=true` from a neighbouring repository would silently break that" is speculative rationale. `:162`'s "**three stages**" is wrong — the Dockerfile has four (`base`, `prod-deps`, `builder`, `runtime`); drop the count. `:139`'s "`latest`, `2` and `2.1`" → "the floating tags".

---

## Verified and correct — recorded so the next round does not re-litigate

- **`latest=auto` is metadata-action's default.** `src/flavor.ts` at `v5` initialises `const flavor: Flavor = { latest: 'auto', … }`. The rewritten comment is accurate.
- **Prerelease tag-withholding is `type=semver`'s behaviour, correctly re-attributed.** `src/meta.ts` `procSemver`: on `semver.prerelease(vraw)` it compiles `{{version}}` regardless of the configured pattern, so all three entries collapse to `2.1.0-rc.1` and dedupe to one tag, and `latest=auto` passes `latest=false`. The rehearsal gets exactly one tag, as claimed. Getting this right without reading the source is rare.
- **`pnpm exec spot` is right** — `package.json` `"bin": { "spot": "./bin/run" }`, and `bin` is in `files`. `pnpm dlx @airtasker/spot init` resolves the same bin. All four flags verified against `cli/src/commands/generate.ts`; adding them explicitly also fixes the old snippet, which would have hit #2712's non-TTY guard.
- **The validation-server block is accurate** — ready line at 3s, `Validation server running on port 5600` on stdout with no TTY, `GET /health` → 200 through `--publish`, `app.listen(port, …)` with no host so it binds all interfaces. The image's command set exactly matches the "what the image carries" table.
- **Version-pin removal is complete** — no `2.1.0` anywhere in `README.md`, three consistent `<version>` placeholders; the one remaining `2.1.0` in the diff is an illustrative workflow comment. `<!-- toc -->` untouched and correct.
- **`docker rm -f <name>` after `--rm` self-removal exits 0**, so the teardown advice is safe in a `finally`. ghcr.io ignores the username with a PAT (verified: nonsense username + real token → "Login Succeeded"), so `-u "$USER"` works where the host account name differs from the GitHub login.
- **Trigger gating, fork safety and permissions are right** — `release: types: [published]` + `if: github.event_name == 'release'` means forks cannot reach this job; `contents: read` + `packages: write` with no `id-token` is least privilege; `github.actor` + `GITHUB_TOKEN` for ghcr is the correct pattern.

## Strengths

- **Unpinning the README's image tag was correct and correctly diagnosed** — `2.1.0` against a `2.0.3` `package.json` would have been `manifest unknown` on the first command a reader copied out of a public README.
- **Retracting the "npm and ghcr cannot diverge" claim** and replacing it with what the workflow guarantees plus a recovery path is the honest version. The two jobs really are unordered siblings — which is also why finding 3 matters.
- The README documents the invocation argument by argument, and each argument genuinely prevents a failure that does not show up in an exit code. That is the right level of detail for a container interface, and findings 5, 8 and 9 are about tightening those claims rather than removing them.
- The Steps to merge rollout list sequences the ghcr private-flip *before* granting access, rehearses with a prerelease, and pull-tests both architecture and grants before the real release. Findings 1 and 7 are about making the code and README match that plan.

vlj91 (dismissed):
🚀

vlj91 (approved):

PR #2718, opened 2026-08-19
build(COMPASS-28): require node >= 22.13 and test on 22, 24, 26

timdawborn:
## Ticket

> **PR 1 of 9, landing strictly sequentially.** This one changes the supported-runtime surface and CI only — no dependency version moves. The eight that follow are one dependency migration each, ending with `ts-morph` and `@oclif/core`.

## Why

Node 18 reached EOL on 2025-04-30 and Node 20 on 2026-04-30, so half the test matrix covered runtimes receiving no security patches.

The declared floor was also **already false**. `eslint` is a *runtime* dependency — `ts-lint` uses it — and `eslint@10.8.1` declares `engines: ^20.19.0 || ^22.13.0 || >=24`. A user on Node 18 installed successfully and then could not run `spot ts-lint` at all. That is what puts the new floor at **22.13.0** rather than a round 22.0.0.

## What

| Kind of claim | Where | Now |
|---|---|---|
| Floor (a range) | `engines.node` in `package.json` and `docs/package.json`, `README.md` | `>=22.13.0` |
| Exact pin | `.tool-versions`, both `Dockerfile` `FROM` lines | `22.23.2` |
| Tested versions | the `test` matrix | `[22, 24, 26]` |

These are deliberately not the same value, which is now written down in `AGENTS.md` — the obvious next edit is to "tidy" all of them into agreement, and that silently re-breaks the floor.

**Node 26 is a required leg.** It becomes LTS on 2026-10-28, so it is the Current line for now. `fail-fast` is off so a regression there cannot suppress the 22 and 24 results needed to place it.

## Three CI facts this audit turned up

- **`test-summary-publish` is the only required status check, and it gated almost nothing.** It had neither a checkout nor a download step, so it globbed an empty directory and reported on nothing; `lint-check` and `docker-parity` were absent from its `needs`, so neither blocked a merge. Both are now in the list. The job also asserts their results itself, because a job skipped by a failed `needs` reports the `skipped` conclusion, which a required check can count as passing — the summary action is reporting only, and the assertion is what decides the job.
- **`prepare-repository` pinned no Node version**, so `lint-check`, `docker-parity` and both publish jobs ran on whatever the runner image shipped — including the `prepack` that builds the published npm artifact. It now takes the `.tool-versions` pin via a new optional `node_version` input, with the matrix passing its own version through so the legs stay independent. All five call sites checked: `docker-publish` legitimately needs no Node.
- **The docs build carried `--openssl-legacy-provider`** behind a `node_version -gt 16` guard that every matrix value satisfied. It is a webpack-4-era MD4 workaround on a tree running webpack 5.104, and it re-enables OpenSSL algorithms Node removed deliberately. Verified `pnpm build-docs` passes without it (exit 0).

`jest.config.js` matched `**/*.spec.ts` while ignoring only `node_modules`. From the repository root that collected **512** spec files out of the worktrees under `.claude/` instead of this tree's **53**.

## How this was verified

Locally on Node 22.23.2 / pnpm 10.28.1, all exit 0:

| Check | Result |
|---|---|
| `pnpm build` | 0 |
| `pnpm test` | 0 — 53 suites, 551 tests, 44 snapshots |
| `pnpm lint:check` | 0 |
| `pnpm build-docs` (no `NODE_OPTIONS`) | 0 — the flag was not load-bearing |
| `docker build` on `node:22.23.2-bookworm-slim` | 0 |
| `./scripts/check-image-parity` | 0 — all checks passed |
| `spot validate` / `lint` / `generate -g openapi3` | 0 |
| `spot ts-lint` on the `clean` / `dirty` fixtures | 0 / 1 as expected |

Confirmed `node:22.23.2-bookworm-slim` exists on Docker Hub before pinning it, and read `actions/setup-node`'s `resolveVersionInput` to confirm `.tool-versions` is a supported `node-version-file` format and that an empty `node-version` falls through to it rather than warning.

## Note for whoever runs parity locally

`docs/webpack.config.js` writes the redoc bundle **into `build/cli/src/commands/docs/`**, which is the directory oclif scans for commands. So running `pnpm build-docs` before invoking the CLI makes oclif try to load a browser bundle as a command, and every command then emits `ReferenceError ... document is not defined` on stderr. Since `check-image-parity` compares combined output, that fails parity for reasons unrelated to the image. CI never hits it because `docker-parity` runs `pnpm build` only, and the published package ships a manifest generated *before* `build-docs` runs. Pre-existing and out of scope here — `rm -rf build && pnpm build` before running parity locally. Worth its own ticket.

## Steps to merge

No branch-protection change needed: the required check is `test-summary-publish`, whose name does not vary with the matrix.


vlj91 (approved):
LGTM!

PR #2719, opened 2026-08-19
build(COMPASS-28): declare @types/node and move to typescript 5.9

timdawborn:
## Ticket

> **PR 2 of 9. Staying in draft** until the whole chain is verified, per the request on #2718. Based on `master` with #2718 merged.

## Why

`@types/node` was **not declared anywhere**. It resolved transitively through jest at `20.8.4`, so the package type-checked against Node 20 lib definitions while `engines` claimed 22. Declaring it forces the compiler bump — every current `@types/node` major sets `typeScriptVersion: 5.6`.

Two deliberate version choices:

- **`@types/node@^22`, the floor rather than the newest (26.2.0).** Type-checking against the *oldest* supported runtime is what makes the compiler reject an API that is missing on it. Pinning to the newest would let a Node 26-only API compile and then fail at runtime on 22.
- **`typescript@^5.9.3` and no further.** `typescript-eslint` peer-depends on `>=4.8.4 <6.1.0` and has **no release above v8**, so TypeScript 6.1+ cannot be used while `ts-lint` ships `typescript-eslint`. This is why Dependabot #2685 (`typescript` → 7.0.2) can never merge as proposed.

## Three things the bump surfaced

**1. `Result` stopped narrowing — 119 errors across 14 parser files.**

`Ok.isErr` and `Err.isOk` returned `boolean`. Narrowing a union by a method call needs *every* member of that union to answer with a type predicate; one member answering `boolean` costs the call its type information, leaving the value un-narrowed in both branches. That showed up in two shapes, both visible in the build output:

```
error TS2322: Type 'Result<Type, ParserError>' is not assignable to type 'Result<Body, ParserError>'
error TS2339: Property 'unwrap' does not exist on type 'Result<Type, ParserError>'
```

The first is the `if (r.isErr()) return r;` early return; the second is the `r.unwrap()` after it. Both are fixed by two return types in `util.ts`, with **no call-site changes** — the alternative was rewriting 65 call sites onto the standalone `isErr()` guard.

Confirmed by compiling the same reduction under 4.9.5 (clean) and 5.9.3 (fails), and confirmed that the standalone guard was unaffected — so this is the method-call form specifically, not `Result` in general.

**2. The spec helper's compiler options had drifted from the CLI's.**

`createProject()` in `lib/src/spec-helpers/helper.ts` duplicated `contractCompilerOptions` but was missing `skipLibCheck` and `resolveJsonModule` — so specs type-checked contracts under settings the real command does not use. That drift is what made `@types/node@22` fail every generator spec: **ts-morph 18 vendors TypeScript 5.0.2**, which cannot resolve `CompressionStream` in `@types/node/web-globals/streams.d.ts`, and only the CLI path suppressed it.

It now spreads `contractCompilerOptions` and overrides just the `@airtasker/spot` mapping, so the duplication cannot drift again. The underlying version skew disappears when ts-morph moves (PR 8).

**3. `process.exitCode` widened** to `string | number | null | undefined`, which the ts-lint command's spec was saving into a `number | undefined`. Now typed off the property itself so it cannot drift again.

## Guarding the invariant

`util.ts` now carries a comment saying those two return types are load-bearing, so `lib/src/util.spec.ts` pins each one with its own case. Mutation-tested both arms independently:

| Mutation | Suite |
|---|---|
| `Ok.isErr` → `boolean` | **fails** |
| `Err.isOk` → `boolean` | **fails** |
| neither | passes |

The first version of that test only caught the `Ok.isErr` mutation, which is why there are two helpers narrowing through different methods rather than one.

## How this was verified

| Check | Result |
|---|---|
| `pnpm build` | 0 — was 119 errors before the `util.ts` fix |
| `pnpm lint:check` | 0 |
| `pnpm build-docs` | 0 |
| `validate` / `lint` / `ts-lint` / `checksum` / `generate` | 0, with `ts-lint` 0 on `clean` and 1 on `dirty` |

**Generated output is unchanged.** All seven generator × language combinations from `test-fixtures/contract/api.ts` are byte-identical between `master` and this branch:

```
diff -r /tmp/spot-before /tmp/spot-after   # empty
```

Expected, since contracts are parsed by ts-morph's vendored compiler rather than this one — but the `target` moved from es2019 to es2022, so it was worth confirming rather than assuming.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,7 +70,7 @@ Spot is a TypeScript-based API contract definition tool that generates OpenAPI, 
 ### TypeScript Configuration
 - Strict TypeScript settings enabled
 - Experimental decorators enabled for syntax support
-- Builds to CommonJS modules targeting ES2019
+- Builds to CommonJS modules targeting ES2022, with `useDefineForClassFields` off so class-field semantics stay as they were
 - Path mapping: `@airtasker/spot` → `./lib/src/lib`
 
 ### Testing
--- a/cli/src/commands/ts-lint.spec.ts
+++ b/cli/src/commands/ts-lint.spec.ts
@@ -13,7 +13,9 @@ const FIXTURES = path.join(__dirname, "../../../test-fixtures/ts-lint");
 describe("ts-lint command", () => {
   jest.setTimeout(60000);
 
-  let restoreExitCode: number | undefined;
+  // Typed off the property rather than spelled out: node widened `exitCode` to
+  // allow a string and null, and this is only ever a saved copy of it.
+  let restoreExitCode: typeof process.exitCode;
   let out: string;
   let writeSpy: jest.SpyInstance;
 
--- a/lib/src/spec-helpers/helper.spec.ts
+++ b/lib/src/spec-helpers/helper.spec.ts
@@ -1,0 +1,28 @@
+import { createSourceFile } from "./helper";
+
+describe("createSourceFile", () => {
+  /**
+   * `validateProject` rejects a source file with any pre-emit diagnostic, and
+   * every parser and linting spec leans on it to reject a bad fixture. The
+   * options it checks under come from `contractCompilerOptions`, which includes
+   * `skipLibCheck`, so the gate is narrower than "no diagnostics at all" — these
+   * two cases pin that it still fails on an error in the contract itself.
+   */
+  it("rejects a contract with a type error", () => {
+    expect(() =>
+      createSourceFile({
+        path: "main",
+        content: 'export const answer: number = "forty-two";\n'
+      })
+    ).toThrow();
+  });
+
+  it("accepts a contract that type-checks", () => {
+    expect(() =>
+      createSourceFile({
+        path: "main",
+        content: "export const answer: number = 42;\n"
+      })
+    ).not.toThrow();
+  });
+});
--- a/lib/src/spec-helpers/helper.ts
+++ b/lib/src/spec-helpers/helper.ts
@@ -1,5 +1,5 @@
-import * as path from "path";
-import { Project, SourceFile, ts } from "ts-morph";
+import { Project, SourceFile } from "ts-morph";
+import { contractCompilerOptions } from "../ts-project";
 
 /**
  * Create an AST source file. Any files imported from the main file must also be provided.
@@ -40,23 +40,12 @@ interface FileDetail {
 export function createProject(): Project {
   return new Project({
     compilerOptions: {
-      target: ts.ScriptTarget.ESNext,
-      module: ts.ModuleKind.CommonJS,
-      strict: true,
-      noImplicitAny: true,
-      strictNullChecks: true,
-      strictFunctionTypes: true,
-      strictPropertyInitialization: true,
-      noImplicitThis: true,
-      alwaysStrict: true,
-      noImplicitReturns: true,
-      noFallthroughCasesInSwitch: true,
-      moduleResolution: ts.ModuleResolutionKind.NodeJs,
-      experimentalDecorators: true,
-      baseUrl: "./",
-      paths: {
-        "@airtasker/spot": [path.join(__dirname, "../lib")]
-      }
+      // Contracts are parsed under the same options `createContractProject`
+      // uses, so a spec and the command it stands in for agree on the compiler
+      // settings. The `@airtasker/spot` mapping needs no override: it is an
+      // absolute path resolved when that module loads, so it points at the same
+      // declarations from here.
+      ...contractCompilerOptions
     }
   });
 }
--- a/lib/src/util.spec.ts
+++ b/lib/src/util.spec.ts
@@ -1,0 +1,42 @@
+import { err, ok, Result } from "./util";
+
+describe("Result", () => {
+  /**
+   * `Ok.isErr` and `Err.isOk` both return `this is never` rather than `boolean`,
+   * and that is what lets a `Result` narrow through a method call — the shape
+   * every parser in `lib/src/parsers` is written in. Narrowing a union by a
+   * method call needs *every* member of the union to answer with a type
+   * predicate; one member answering `boolean` costs the call its type
+   * information, leaving the value un-narrowed in both branches.
+   *
+   * What is under test is the narrowing, so the failure is a compile error
+   * rather than a failed expectation: widen either return type back to
+   * `boolean` and this file stops type-checking, which ts-jest reports as a
+   * failing suite. The two helpers below narrow through different methods, so
+   * each pins one method and neither covers the other.
+   */
+  const narrowViaIsErr = (r: Result<number, Error>): Result<string, Error> => {
+    // Needs `r` to narrow to `Err`, or an `Err<Error>` is not assignable to a
+    // `Result<string, Error>`. Pins `Ok.isErr`.
+    if (r.isErr()) return r;
+    return ok(String(r.unwrap()));
+  };
+
+  const narrowViaIsOk = (r: Result<number, Error>): string => {
+    if (r.isOk()) return String(r.unwrap());
+    // Needs `r` to narrow to `Err`, or `unwrapErr` is not a property of the
+    // union. Pins `Err.isOk`.
+    return r.unwrapErr().message;
+  };
+
+  it("narrows to Ok when the value is present", () => {
+    expect(narrowViaIsErr(ok(7)).unwrapOrThrow()).toBe("7");
+    expect(narrowViaIsOk(ok(7))).toBe("7");
+  });
+
+  it("narrows to Err when the error is present", () => {
+    const failure: Result<number, Error> = err(new Error("no value"));
+    expect(narrowViaIsErr(failure).unwrapErrOrThrow().message).toBe("no value");
+    expect(narrowViaIsOk(failure)).toBe("no value");
+  });
+});
--- a/lib/src/util.ts
+++ b/lib/src/util.ts
@@ -75,7 +75,16 @@ class Ok<T> {
     return true;
   }
 
-  isErr(): boolean {
+  /**
+   * `this is never` rather than `boolean`, because the return type is what makes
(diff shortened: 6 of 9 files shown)


timdawborn:

Agents run: **code-reviewer**, **comment-analyzer**, **type-design-analyzer**, **pr-test-analyzer**. Scoped to this PR's own commit (`8100bce`).

Two agents disagreed on one point; I verified it myself and say which won below.

### 🔴 Critical (1)

**`lib/src/spec-helpers/helper.ts:42-49` — the `paths` override is dead code, and the comment exists to justify it.**

The comment says "Only the `@airtasker/spot` mapping differs, because this file sits a directory deeper". It doesn't differ. `path.join` runs at load, so both sides already hold the same absolute path:

```
ts-project.ts mapping : <root>/lib/src/lib
helper.ts     mapping : <root>/lib/src/lib
identical?: true
```

`ts-project.ts` is in `lib/src` (`"lib"`), `helper.ts` in `lib/src/spec-helpers` (`"../lib"`) — different expressions, same result. So the spread alone would carry the correct value and the override does nothing.

> **Agent disagreement:** code-reviewer read this as correct ("resolves one level deeper"), comment-analyzer called it dead code. The *source file* is a level deeper, but the resolved path is identical, so comment-analyzer is right. I ran the probe above to settle it.

Same comment, second issue: "so a spec cannot pass on a setting the real command does not use" overclaims — nothing enforces that; the literal can still override keys, and already does.

### 🟡 Important (3)

**1. `tsconfig.json:9` — the target bump silently changes the published API.** `es2019 → es2022` flips `useDefineForClassFields` to `true` (confirmed via `tsc --showConfig`). Verified against the built package:

| target | `Object.keys(new ParserError(...))` | `JSON.stringify` |
|---|---|---|
| es2019 | `["locations"]` | `{"locations":[…]}` |
| es2022 | `["message","locations"]` | `{"message":"boom","locations":[…]}` |

`ParserError` is exported, so **any consumer serialising one gets a different payload.** Nothing in the diff, the comment or a test pins this, and the commit message enumerates three things the bump surfaced without mentioning it. Either set `"useDefineForClassFields": false` to keep emit unchanged, or split the target bump into its own commit that states the consequence and covers the error shape with a test.

**2. `AGENTS.md:73` — "Builds to CommonJS modules targeting ES2019" is now false.** Doc/code drift inside this PR; the rule says fix it in the same commit.

**3. `helper.ts:44-48` — the loosened diagnostic gate has no test.** `validateProject` throws on `getPreEmitDiagnostics()`, and 16 parser/linting suites depend on it. The spread now silently adds `skipLibCheck: true`, so that gate no longer sees declaration-file errors. `parser.spec.ts:27` covers the equivalent CLI path but nothing covers `createProject()`. One case asserting it still throws on a contract type error would stop the helper going quiet.

### 🔵 Suggestions (5)

- **`helper.ts:45` overclaims "the drift cannot recur"** — `paths` is wholly replaced, not merged, so a second `paths` entry added upstream is dropped here with no signal.
- **The standalone `isOk`/`isErr` guards have zero call sites** repo-wide. This commit picked the method form and left two exported narrowing mechanisms with nothing marking which is canonical. Delete them or say which wins.
- **9 vacuous assertions in adjacent parser specs** (pre-existing, not from this commit): `expect(x!.isOk).toBeTruthy()` in `schemaprop-parser.spec.ts` and `example-parser.spec.ts` asserts the *method reference*, which is always truthy — those specs never verify success. Direct evidence the method encoding invites silent misuse.
- **`util.spec.ts:20-21` states the mechanism backwards** — it says an `Err<Error>` isn't assignable to `Result<string, Error>`; under correct code `r` narrows to `Err<Error>` there and that *is* assignable. The real mutation error is `Ok<number>` leaking in un-narrowed.
- **`util.spec.ts:6-7` "the shape every parser in `lib/src/parsers` is written in"** is an enumerated claim already false (13 of 16 files) — the repo's rule about lists that go stale.
- **The `@types/node` rationale in the commit message overclaims**: `@types/node@22.20.1` declares the full Node 22 surface including APIs added *after* the `>=22.13.0` floor, so the major pin bounds the line, not the floor.

### ✅ Verified sound

- **`util.spec.ts` genuinely pins both predicates** — independently reproduced: widening `Ok.isErr` fails only inside `narrowViaIsErr`, widening `Err.isOk` fails only inside `narrowViaIsOk`, and runtime polarity is pinned too (flipping any of the four bodies fails an assertion). Unusually well-motivated coverage.
- **`this is never` is correct and sound**, verified by probe on the union and on a generic receiver: the dead branch is *restrictive*, not permissive (`ok(1).isErr()` narrows `this` to `never`, so `unwrapErr()` there is a type error). Ratings: encapsulation 8/10, invariant expression 6/10 (correct but oblique, comment-dependent), usefulness 9/10.
- `typeof process.exitCode` is the right fix. `tsc --noEmit`, prettier and eslint clean.
- **A caveat on the test signal:** `tsconfig.json` excludes `**/*.spec.ts`, so `pnpm build` never type-checks `util.spec.ts` — ts-jest's diagnostics are the only gate there. Mitigated by the same mutation failing 16 parser suites via files that *are* in the `tsc` include, so the invariant has a second, stronger gate. `util.spec.ts` is best read as a named canary, not the last line of defence.

### Recommended action

1. **Decide on `useDefineForClassFields`** — this is the only finding that changes shipped behaviour.
2. Delete the dead `paths` override and the sentence justifying it.
3. Fix `AGENTS.md:73` in this commit.
4. Then the comment corrections and the `createProject` gate test.

timdawborn:

Agents: **code-reviewer**, **comment-analyzer**. Scoped to the revised commit `a0326e3`.

All four round-1 fixes verified sound. The problems are in the **comment I wrote to explain one of them**, and in round-1 findings I left unaddressed.

### 🔴 Critical (2)

**1. `tsconfig.json:10-14` — the mechanism I described is wrong, and a reader following it would conclude the pin is unnecessary.**

I wrote that define semantics turn "a declared-but-unassigned field into an own enumerable property". Both agents independently disproved it. `locations` *is* a declared field and is enumerable under **both** settings. The field that actually flips is **`message`, a constructor parameter property**: under define semantics TS emits a bare `message;` declaration that redefines `Error`'s non-enumerable own `message` as enumerable. A genuinely declared-but-unassigned field is `undefined`-valued, which `JSON.stringify` **omits** — so the rule as stated cannot produce the effect it names.

The real predicate is *a field declaration shadowing a base-class own property*. Suggested: "Under define semantics a field declaration redefines the property rather than assigning it, so `ParserError`'s `message` — inherited non-enumerable from `Error` — becomes enumerable and appears in `JSON.stringify` output."

**2. `tsconfig.json:15` — a stray comment now documents the wrong option.** The trailing `/* Specify ECMAScript target version: 'ES3' … 'ESNEXT' */` belonged to the deleted `"target": "es2019"` line and is now glued to `useDefineForClassFields`. Its enumeration doesn't even list ES2022. Delete it. My edit created this.

### 🟡 Important (2)

**3. Overclaim on blast radius.** The comment says `ParserError` "and its siblings are exported" and frames it as "what a consumer gets". `errors.ts` is **not reachable from the public surface** — `index.ts` → `lib/src/lib.ts` exports only `core` and `syntax`, and nothing in `lib/` or `cli/` stringifies one. Say the shape is observable on any instance and drop "consumer". My PR description overstates this too.

**4. No test pins the `ParserError` JSON shape.** The comment declares a published-API invariant and nothing fails if a future target bump drops the pin. Round 1 added a test for the *lesser* invariant (the `createProject` gate) and not this one. An `errors.spec.ts` asserting `Object.keys(new ParserError("m", …))` equals `["locations"]` closes it.

### 🔵 Round-1 findings I did NOT address (4)

These were Suggestions in round 1 so I skipped them, but they are exactly the "fix one site, leave the siblings" pattern:

- **`util.spec.ts:19-20`** — still states the mechanism backwards. Widening `Ok.isErr` gives `TS2322: Ok<number> is not assignable to Result<string, Error>`. `Err<Error>` *is* assignable; the un-narrowed `Ok<number>` is not. (Its twin at :27-28 is correct.)
- **`util.spec.ts:7`** — "every parser in `lib/src/parsers`" is still an enumerated claim, still false: 13 of 16.
- **`util.ts:117`** — "`Ok.isErr` above" is still a positional reference.
- **`util.ts:78-86` / `util.spec.ts:4-16`** — the rationale is still stated twice at full length and will drift.

Plus two new small ones: `helper.spec.ts:6` says "every parser and linting spec" (`find-lint-violations.spec.ts` does not), and `ts-lint.spec.ts:16-17` still narrates ("node widened `exitCode`").

### ✅ Verified sound

- **The pin restores the published shape exactly** — built and instantiated for real: `{"locations":[…]}`, `message` descriptor `enumerable: false`, byte-identical to an es2019 baseline. Recompiled with the flag true: `{"message":"boom",…}`.
- **The pin is safe across the codebase.** Emit diffed both ways over the whole tree: 15 files differ, none a regression — the rest are standalone classes or `static`-only. `ParserError` is the sole class shadowing a base own property. And `target: es2019` already implied `useDefineForClassFields: false`, so **the pin is a no-op against master by construction**.
- **`false` is the right call** — a consumer-visible serialisation change does not belong in a `build(...)` commit. The alternative may be the better end state (`JSON.stringify` on an error currently drops the message, a real logging footgun) but needs a major and a changelog.
- The dead `paths` override removal is clean, and `helper.spec.ts` is load-bearing — flipping `skipLibCheck` to `false` fails it.
- **No sibling claim missed** on the build target: `git grep` finds only `AGENTS.md:73` (fixed) and `tsconfig.json`. Two notes below the bar: `docs/tsconfig.json` still targets `es5`, and `contractCompilerOptions` targets `ESNext` with no pin — harmless, since contracts are parsed and never emitted, but more visible now `helper.ts` spreads those options.

### Recommended action

1. **Rewrite the mechanism comment** — as written it would actively mislead the next person into dropping the pin.
2. Delete the stray target enumeration at `:15`.
3. Add `errors.spec.ts` to pin the shape the comment claims.
4. Sweep the four unaddressed round-1 comment findings together.

